Issue #42, opened 2026-06-08
Queue: SPSC batched handoff scenario


vic-lsh:
## Inherits

- Queue parent: #37
- Reference / benchmark / checker harness: #43

## Scenario

A single producer hands off ordered items to a single consumer in finite batches.

## Motivation / Plausible Workloads

This scenario captures handoffs where the producer exposes work at batch boundaries, the consumer naturally processes finite groups of items, and partial batches must never be observed.

Concrete case: a tracing subsystem accumulates events during an interval and publishes a completed batch to one writer thread; the writer must see complete batches in publish order, not individual events as they are being collected.

## Semantic Requirements

- Exactly one producer may submit items.
- Exactly one consumer may receive batches.
- The producer may publish finite batches of submitted items.
- The consumer observes complete published batches, not partially published batches.
- Items within each batch must preserve producer submission order.
- Published batches must be observed in publish order.
- No item from a later batch may be observed before an item from an earlier batch.
- Every successfully submitted item must be delivered exactly once unless the scenario defines full or shutdown behavior that prevents delivery.
- When the producer outpaces the consumer, submit or publish behavior must follow the capacity and failure behavior defined by the scenario configuration.

## Correctness Checks

- The checker must verify exactly-once delivery for successfully submitted items.
- The checker must verify item order within each batch.
- The checker must verify batch publish order.
- The checker must verify that no partially published batch is observed.
- The checker must verify that failed submit or publish operations do not make items visible to the consumer.

## Non-Goals

- No multiple-producer support.
- No multiple-consumer support.
- No lossy overwrite behavior.
- No prescribed implementation strategy.

vic-lsh:
Not planned for now

Issue #40, opened 2026-06-08
Queue: MPMC bounded scenario


vic-lsh:
## Inherits

- Queue parent: #37
- Reference / benchmark / checker harness: #43

## Scenario

A bounded handoff with multiple producers and multiple consumers.

## Motivation / Plausible Workloads

This scenario captures shared work pools where many producers create independent units of work and many consumers process them, with bounded memory and exactly-once handoff as the core requirements.

Concrete case: a runtime thread pool accepts tasks from many application threads and has many worker threads execute them; each accepted task must run once, and the queue must define behavior when capacity is exhausted.

## Semantic Requirements

- Multiple producers may submit items concurrently.
- Multiple consumers may receive items concurrently.
- Capacity is finite and defined by the scenario configuration.
- Every successfully submitted item must be received exactly once by exactly one consumer unless the scenario defines shutdown behavior that prevents delivery.
- For each producer, that producer's successfully submitted items must be received in submission order.
- No total ordering is required between items submitted by different producers.
- A submit operation that fails because the queue is full must not make the item visible to any consumer.
- A receive operation that fails because the queue is empty must not consume an item.

## Correctness Checks

- The checker must verify exactly-once delivery for successfully submitted items.
- The checker must verify that no item is received by more than one consumer.
- The checker must verify per-producer order.
- The checker must not require a total order across different producers.
- The checker must verify that failed submit and receive operations do not corrupt queue state.

## Non-Goals

- No global FIFO guarantee across producers.
- No lossy overwrite behavior.
- No prescribed implementation strategy.

vic-lsh:
## Research anchor: moodycamel

[moodycamel::ConcurrentQueue](https://github.com/cameron314/concurrentqueue) is a useful baseline family for this scenario's per-producer FIFO semantics. Its design uses independent producer subqueues and explicitly does **not** provide one global linearizable order across coordinated producers, matching this issue's non-goal of total cross-producer ordering.

For the bounded scenario, an adapter would still need to enforce the configured capacity with preallocated storage and non-allocating try operations; the upstream queue's dynamically growing mode should not silently change the scenario contract. The project also documents that it is not NUMA-aware, which is worth retaining as benchmark context.

This is a research/baseline anchor, not a prescribed implementation strategy.

PR #54, opened 2026-06-17
Add default queue reference, accuracy checker, and benchmark harness

### LINKED ISSUE #43: Queue: create default reference, benchmark, and checker harness
## Inherits

- Queue parent: #37

## Purpose

Create the default starting point for queue scenarios: reference implementations, benchmark harnesses, and correctness checkers that the queue scenarios can reuse.

## Required Components

- Reference implementation or reference implementations for the initial queue scenarios.
- Scenario-specific API definitions where the scenarios require different operations.
- Benchmark driver with configurable item size, capacity, producer count, consumer count, duration, warmup, and scenario selection.
- Correctness checker for shared queue terms and scenario-specific semantic requirements.
- Scenario-specific checker hooks for SPSC bounded FIFO, MPSC bounded, MPMC bounded, lossy overwrite, and SPSC batched handoff.
- Scenario configuration files or flags for each initial scenario.

## Acceptance Criteria

- Each reference implementation passes its scenario's correctness checker.
- The benchmark can run each initial scenario without changing benchmark code.
- Scenario issues can reference this harness without restating shared terminology.

## Scope

This issue is only the reusable harness and reference/checker starting point. It should not prescribe optimization strategies or implementation data structures.

zatchbell1311-wq:
Closes #43

## Summary

Adds a reusable harness for the five initial VibeServe queue scenarios: `spsc`, `mpmc`, `mpsc`, `lossy`, and `batch`. This follows the same `reference/ + accuracy_checker/ + benchmark/` layout used by the existing LLM-serving examples, so scenario issues for queues (#40–#42, etc.) can build on a shared contract instead of redefining harness conventions each time.

## What's included

- `reference/reference.py` — baseline implementations for all five scenarios, built on `collections.deque` with `threading.Lock`/`Condition` for blocking variants. All variants expose the same minimal protocol: `enqueue(item)`, `dequeue()`, `size()`, `capacity`.
- `accuracy_checker/checker.py` — verifies a candidate implementation against the reference:
  - `spsc`: replays a randomized op log against both reference and candidate, asserts identical results at every step.
  - `mpmc` / `mpsc`: runs concurrent producer/consumer threads and checks for no lost or duplicated items.
  - `lossy`: checks the queue never exceeds capacity and never returns an item that wasn't enqueued.
  - `batch`: checks batched dequeues never return duplicates or fabricated items.
- `benchmark/benchmark.py` — measures throughput (ops/sec) for each scenario under configurable producer/consumer counts and duration.
- `OBJECTIVE.md` / `README.md` — scenario descriptions and usage instructions for implementers.

## Verification

Ran the checker against the reference implementation itself (sanity baseline — all 5 scenarios pass). I also deliberately broke the SPSC candidate (had it pop from the wrong end of the deque, violating FIFO order) to confirm the checker actually catches incorrect implementations rather than passing everything — it correctly failed and reported the exact operation index where the mismatch occurred.

## Notes

`mpmc` and `mpsc` currently share the same underlying bounded-queue class as `spsc` — the harness doesn't yet enforce scenario-specific concurrency guarantees beyond what the checker tests for. Happy to extend if reviewers want stricter differentiation between these reference variants.

### DIFF
--- a/examples/queue-default/OBJECTIVE.md
+++ b/examples/queue-default/OBJECTIVE.md
@@ -1,0 +1,20 @@
+# Objective - Queue default harness
+
+Maximize throughput (operations/second) of a concurrent queue under the
+workload defined by the active scenario, while satisfying correctness invariants.
+
+## Scenarios
+
+| Scenario | Description |
+|----------|-------------|
+| spsc | Single-producer single-consumer bounded FIFO |
+| mpmc | Multi-producer multi-consumer bounded FIFO |
+| mpsc | Multi-producer single-consumer bounded FIFO |
+| lossy | Single-writer lossy overwrite (newest item wins on full) |
+| batch | SPSC batched handoff (producer drains into consumer in bulk) |
+
+## Notes
+
+- Reference uses collections.deque with threading.Lock.
+- Protocol: enqueue(item), dequeue(), size(), capacity.
+- No hardware accelerator required; CPU-only.
--- a/examples/queue-default/README.md
+++ b/examples/queue-default/README.md
@@ -1,0 +1,23 @@
+# Queue Default Harness
+
+Resolves #43.
+
+Reusable reference implementations, correctness checkers, and benchmark
+drivers for the five initial VibeServe queue scenarios.
+
+## Running the correctness checker
+
+    python accuracy_checker/checker.py --scenario all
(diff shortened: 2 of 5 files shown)


vic-lsh:
LGTM for an initial version. Thanks!

Issue #35, opened 2026-06-08
Expand VibeSys beyond LLM serving systems


vic-lsh:
## Purpose

Track the effort to expand VibeServe-style system specialization beyond LLM serving into other widely used systems and infrastructure code.

## Current Areas

- KV store: #26
- Distributed queue: #27
- Specialized data structures: #36
- Online application specialization: #44

## Some Areas To Consider Next

- Operating systems
- Application runtimes
- Networking stack
- Storage systems
- Web/API serving
- Developer infrastructure

## Scope

This issue is the top-level planning container for the expansion effort. Area-specific issues should define shared API and semantic contracts. Scenario issues should describe semantic or workload requirements without prescribing optimization strategies.

Issue #26, opened 2026-06-08
KV store: shared API and semantics


vic-lsh:
## Purpose

Define the common key-value store contract that all KV scenarios inherit.

## Shared API Shape

The shared contract should define a minimal API for:

- `get(key)`
- `put(key, value)`
- `delete(key)`
- optional `list` or range operations only if the harness requires them
- health/stats reporting needed by benchmarks and correctness checks

## Shared Semantics

- Keys are byte strings or UTF-8 strings, as defined by the harness.
- Values are opaque byte strings.
- A successful `put` makes the value visible to later successful `get` calls for the same key.
- A successful `delete` removes the key for later successful `get` calls.
- Failed operations must not be reported as successful.
- The shared contract should explicitly define what consistency model the scenarios inherit.

## Scope

This issue is only the shared API and semantic contract. It should not prescribe optimization strategies.

vic-lsh:
Done

PR #57, opened 2026-06-18
feat(kv): add default reference, checker, and benchmark harness

### LINKED ISSUE #28: KV store: create default reference, benchmark, and checker harness
## Inherits

- Shared API and semantics: #26

## Purpose

Create the default starting point for KV store scenarios: a reference implementation, benchmark harness, and correctness checker that all KV workload scenarios can reuse.

## Required Components

- Reference implementation that satisfies the shared KV contract.
- Benchmark driver with configurable workload mix, key distribution, value size, client count, duration, and warmup.
- Correctness checker for the shared KV semantics.
- Scenario configuration files or flags for read-heavy, write-heavy, and high-concurrency write-heavy workloads.

## Acceptance Criteria

- The reference implementation passes the shared correctness checker.
- The benchmark can run each initial KV scenario without changing benchmark code.
- Scenario issues can reference this harness without restating the shared API.

## Scope

This issue is only the reusable harness and reference/checker starting point. It should not prescribe optimization strategies.

zatchbell1311-wq:
## Summary

Adds `examples/kv-default/` — a reusable harness for the four initial KV
store scenarios defined in issue #26. Structured identically to existing
examples (reference / accuracy_checker / benchmark) so the VibeServe agent
loop can run against it without any framework changes.

## What's included

**`reference/reference.py`**
Thread-safe dict-backed reference implementations for all four scenarios:
- `point` — balanced mixed put/get/delete on individual keys
- `scan` — point ops plus prefix-range scan over ordered keys
- `heavy-write` — write-dominated workload (high put/delete rate)
- `read-heavy` — read-dominated workload (high get rate, infrequent puts)

All classes share a common `_BaseKVStore` with the contract from #26:
`put(key, value) → bool`, `get(key) → bytes | None`, `delete(key) → bool`,
`size() → int`, `stats() → dict`. The `scan` scenario adds
`scan(prefix) → list[tuple[str, bytes]]`.

**`accuracy_checker/checker.py`**
Validates a candidate `VibeServeKV` from `main.py` against four invariants:
- Sequential put/get/delete correctness against the reference (point scenario)
- Delete semantics: `delete` returns `True` only for existing keys, `get`
  after `delete` returns `None`, double-delete returns `False`
- Prefix scan result matches reference output exactly (scan scenario)
- Concurrent correctness under multi-writer/multi-reader load with no lost
  writes and no wrong-type returns (heavy-write and read-heavy scenarios)

**`benchmark/benchmark.py`**
Throughput driver with configurable writers, readers, key count, value size,
warmup period, and optional JSON output. Reports per-operation rates (put/s,
get/s) plus hit rate and total ops/s. Supports `--use-reference` to baseline
against the reference implementation.

## Consistency model implemented

Follows the contract in #26:
- A successful `put` makes the value visible to all later `get` calls for
  that key (read-your-writes within a single process)
- A successful `delete` removes the key; subsequent `get` must return `None`
- Failed operations do not modify visible state

## Testing

All four scenarios pass the correctness checker against the reference:

    $ python accuracy_checker/checker.py --scenario all
    [POINT] Checking ...
      PASS - Delete semantics OK (200 cycles)
    [SCAN] Checking ...
      PASS - Scan OK (2000 puts, 4 prefixes checked)
    [HEAVY-WRITE] Checking ...
      PASS - Concurrent heavy-write OK (2000 ops, 4W/1R)
    [READ-HEAVY] Checking ...
      PASS - Concurrent read-heavy OK (2000 ops, 1W/4R)
    Results: 4/4 passed

## Related issues

- Closes #28
- Implements shared API contract from #26
- Part of the expansion tracked in #35

### DIFF
--- a/examples/kv-default/OBJECTIVE.md
+++ b/examples/kv-default/OBJECTIVE.md
@@ -1,0 +1,36 @@
+# Objective - KV store default harness
+
+Maximize throughput (operations/second) of a key-value store under the
+workload defined by the active scenario, while satisfying the correctness
+invariants defined in issue #26.
+
+## Scenarios
+
+| Scenario    | Description                                               |
+|-------------|-----------------------------------------------------------|
+| point       | Mixed put/get/delete on individual keys (balanced load)   |
+| scan        | Point ops plus prefix-range scans over ordered keys       |
+| heavy-write | Write-dominated: high put/delete rate, minimal reads      |
+| read-heavy  | Read-dominated: high get rate, infrequent puts            |
+
+## API Contract (issue #26)
+
+    put(key: str, value: bytes) -> bool      # True on success
+    get(key: str) -> bytes | None            # None if key absent
+    delete(key: str) -> bool                 # True if key existed
+    size() -> int                            # current key count
+    stats() -> dict                          # implementation-defined metrics
+
+## Consistency Model
+
+- A successful put makes the value visible to later get calls for the same key.
+- A successful delete removes the key; subsequent get must return None.
+- delete on a non-existent key returns False and is a no-op.
+- Failed operations must not modify visible state.
+
+## Notes
+
+- Reference uses a threading.Lock-protected dict.
+- Keys are UTF-8 strings. Values are opaque bytes.
+- No hardware accelerator required; CPU-only.
+- scan scenario requires an additional scan(prefix) method.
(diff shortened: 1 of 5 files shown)


vic-lsh:
Thanks for opening this PR! We have an internal KV implementation working in parallel, so we'll proceed with merging that one.

PR #60, opened 2026-06-21
feat(agent): abstract domain knowledge out of the prompt skeleton via --domain

RockingMat:
## What

Splits the agent loop's prompts into a **neutral skeleton** and a swappable **domain pack**. All the LLM-serving-specific prose — the serving-systems skill, the accuracy/benchmark/reward-hack judge gates, the continuous-batching → attention-kernel → CUDA-graph optimization floor — moves out of `implementer/judge/single_agent/orchestrator` templates and into one Markdown file. The skeleton keeps only domain-agnostic structure.

A domain is **one `.md` file** with `## <role>` sections injected via `{{ domain_<role> }}`. `--domain` takes a built-in name (`llm-serving`, `generic`) or a path. Default `llm-serving`.

Every `## <role>` section is rendered with the **same uniform variable set** (`_domain_render_context`: `modality`, `reference_path`, `bench_path`, `accuracy_checker_path`, `runtime_notes`), so a pack author can branch (`{% if bench_path %}`) on any run variable from any section — no per-role context to memorize and no `is defined` guards.

## Why

This is the abstraction seam that lets the same loop target non-LLM systems (a KV store, etc.) without forking the templates.

## Verification

- **Byte-identical default** — prompts under `--domain llm-serving` render exactly as before (verified by diffing renders); `generic` injects nothing.
- `tests/loops/agent/` **165 passing**, incl. 17 new in `test_domain_packs.py` (resolver, renderer, role parsing, single-agent derivation, orchestrator injection).

## Known / intentional scope

`orchestrator_plan_prompt.j2` still uses LLM *examples* in its prose (FlashInfer, CUDA graph, `/v1/completions`). These are illustrative skeleton text, **not** injected content — a non-LLM planner overrides them via its `## orchestrator` section + the objective. Generalizing them splits principle-from-example across four sections, so it's deferred until a second domain needs it.

## Reviewing

Start at `_domain/README.md` (format), then `domain.py` (resolver/renderer), then the template diffs. History has two self-corrections — **squash-merge**.


vic-lsh:
Superceded by #62 .

@RockingMat Thanks for opening this PR! I added a few additional tests to your PR and merged it. Feel free to check out #62 to see how we use snapshot testing to test and see fully rendered prompts.

PR #118, opened 2026-07-08
Add domain runtime environment hooks

### LINKED ISSUE #117: Add domain-specific environment setup hooks
## Problem

The generic run context currently assumes every reference directory may require model weights. During workspace setup, _RunContext calls _ensure_model_weights(ref_dir) for reference directories without meta.json, which makes non-model-serving inputs such as examples/data-structures/queue-default fail unless they provide a dummy reference/model directory.

That behavior only makes sense for the llm-serving domain. Generic domains should be able to prepare and run without model-serving artifacts.

## Proposal

Introduce a small first-party domain runtime hook API for environment setup/teardown. The base runtime should stay generic and let built-in domains request domain-specific setup as a declarative patch.

A minimal initial patch should support:

- names to exclude from normal workspace copy, such as model artifact directories
- read-only bind mounts requested by the domain, such as model weights exposed at /model

Then move model-weight materialization and /model setup into the llm-serving runtime policy. The generic runtime should be a no-op.

## Acceptance Criteria

- --domain generic with examples/data-structures/queue-default does not require reference/model or reference/meta.json.
- llm-serving still supports existing reference/meta.json model download/link behavior.
- Docker/local model mount behavior still works through the new domain runtime patch.
- The domain runtime API starts minimal and avoids hardcoding model-serving semantics into its names.

vic-lsh:
## Summary

Closes #117.

- Add a first-party domain runtime hook API with `DomainSetupContext`, `DomainEnvironmentPatch`, and no-op/LLM-serving runtime implementations.
- Move model-weight materialization out of generic `_RunContext` and into the LLM-serving runtime policy.
- Pass domain-declared bind mounts into sandbox environment setup so `/model` and `/draft_model` are no longer inferred globally from any `reference/model` directory.
- Preserve legacy model-serving behavior for plain/evolve/openevolve by explicitly using the LLM-serving runtime there.

## Validation

- `./scripts/check_lint.sh`
- `uv run pytest`
- Smoke: `uv run vibe-serve --outer-loop agent --config /tmp/vibeserve-agent-gpt55.toml --agent-backend cli --cli-provider codex --domain generic --backend cpu --interface inprocess --input examples/data-structures/queue-default --no-skills --max-rounds 0 --exp-name generic-setup-smoke`

### DIFF
--- a/src/vibe_serve/context.py
+++ b/src/vibe_serve/context.py
@@ -1,4 +1,4 @@
-"""Shared run context: _RunContext, setup_exp_dir, _TeeWriter, _ensure_model_weights."""
+"""Shared run context: _RunContext, setup_exp_dir, and _TeeWriter."""
 
 import json
 import os
@@ -21,6 +21,11 @@ from vibe_serve.constants import (
     DEFAULT_COMPUTE_BACKEND,
     PROJECT_ROOT,
     ComputeBackend,
+)
+from vibe_serve.domain_runtime import (
+    DomainRuntime,
+    DomainSetupContext,
+    NoopDomainRuntime,
 )
 from vibe_serve.llm_client import _build_model
 from vibe_serve.sandbox.run_environment import (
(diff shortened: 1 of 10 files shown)


vic-lsh:
Superseded by #119, which combines registered domains, typed domain roles, and environment hooks in one PR.

Issue #76, opened 2026-07-02
Support generic example directories with standardized run scripts


vic-lsh:
## Problem

VibeServe currently assumes examples follow a Python/model-serving layout:

- `--ref` is a Python file or a directory containing `reference.py`
- accuracy checkers are run as `checker.py`
- benchmarks are run as `benchmark.py`
- judge/profiler prompts assume a `/health` endpoint and Python benchmark/checker commands

That blocks examples where the target is a service stack, the checker/benchmark are compiled binaries, setup requires Docker Compose, or the service exposes readiness somewhere other than `/health`.

PR #74 (`social-network-read-timeline`) is a concrete example: it has C++ checker/benchmark tools, a DeathStarBench Docker Compose stack, nginx patches, and a service on port `8080`.

## Proposal

Add support for loading an example directory as the unit of configuration:

```bash
vibe-serve run --example examples/social-network-read-timeline
```

Each example can implement standardized lifecycle scripts at the example root:

```text
setup.sh
build.sh
check.sh
bench.sh
teardown.sh
```

The scripts define how to prepare, build, check, benchmark, and clean up the example. They may call Python, CMake, Docker Compose, shell tools, or anything else the example needs.

The framework should pass standard environment variables to these scripts, such as:

```bash
VIBESERVE_BASE_URL=http://localhost:8080
VIBESERVE_OUTPUT_DIR=/tmp/vibeserve-run-...
VIBESERVE_LOAD_LEVEL=medium
```

## Minimal manifest

Add a small `vibeserve.example.toml` at the example root for metadata the framework must understand:

```toml
[service]
base_url = "http://localhost:8080"
health_check = "GET /wrk2-api/user-timeline/read?user_id=1&start=0&stop=1"

[benchmark]
primary_metric = "p50_ms"
direction = "minimize"
metrics_format = "jsonl"
```

Field meanings:

- `service.base_url`: where the target service is exposed; examples should not have to use the current `localhost:8000` convention.
- `service.health_check`: example-specific readiness probe; not every target has `/health`.
- `benchmark.primary_metric`: the metric VibeServe should optimize.
- `benchmark.direction`: whether lower or higher is better.
- `benchmark.metrics_format`: how to parse benchmark output. Start with JSONL metrics like:

```json
{"metric":"p50_ms","value":5.64}
```

## Initial contract

- `setup.sh`: start or initialize dependencies.
- `build.sh`: build checker/benchmark/helper tools if needed.
[...]

PR #132, opened 2026-07-10
Complete issue #76: generic example directory support

AyanBinRafaih:
# Generic Example Directory Support

Initially, VibeServe assumed every target was a Python model-serving example (a single reference.py, a checker.py, a benchmark.py, a `/health` endpoint, and a `VibeServeModel` class). This change lets a target be described by a manifest file plus a small set of shell scripts, so a more examples, including microservices, data-structure targets, and non-Python targets, can be optimized too. All existing examples keep working unchanged.

## Implementation Detaills

The follow five tasks have been implemented, keeping in mind the requirements outlined in the issue:

1. `--example <dir>` discovers `vibeserve.example.toml` and the root-level lifecycle scripts.
  - `_expand_example_flag` in `src/vibe_serve/cli.py` expands `--example <dir>` into `--ref <dir>/reference`, `--acc-checker <dir>/accuracy_checker`, `--bench <dir>/benchmark`, only for flags the user did not pass explicitly.
  - `ExampleManifest.detect` in `src/vibe_serve/example_manifest.py` finds the manifest and `_RunContext` switches to the generic contract.
2. Existing `--ref`, `--acc-checker`, `--bench` flows continue to work.
  - When no manifest is present, `_RunContext` uses the original legacy path. Verified against all nine `examples/model-serving/*` targets and `examples/data-structures/queue-default`.
3. Judge and profiler prompts use the manifest and scripts instead of assuming `/health`, `checker.py`, `benchmark.py`.
  - `render_check_instructions` and `render_bench_instructions` on the manifest produce the text injected into the Judge, Implementer, and Mutator prompts.
  - Prompt templates in all four loops (agent, plain, evolve, openevolve) branch on `is_generic_example` and drop the FastAPI / uv / pytest / TTFT-TPOT specific sections for generic targets.
4. Non-Python checker and benchmark implementations are supported.
  - check.sh and bench.sh may call anything (compiled binaries, other languages, docker compose). The framework only reads their exit code and, for bench.sh, their JSONL metric output.
5. Benchmark JSONL output is parsed and the declared primary metric is used for comparison.
  - `ExampleManifest.parse_metrics` parses lines of the form `{"metric": "<name>", "value": <number>}`.
  - `run_example_benchmark` in `src/vibe_serve/loops/profiler.py` runs bench.sh in the sandbox, parses its output, and returns a `ProfilerSummary` whose `perf_metric` and `metrics` fields the loops already consume.
  - `to_objective` (wired in `_resolve_objectives` in `src/vibe_serve/cli.py`) turns the declared `primary_metric` and `direction` into the evolve loop's objective, so ranking uses the declared metric with the declared direction.

## The manifest format

The manifest file is named `vibeserve.example.toml` and sits at the example root, next to the lifecycle scripts and `OBJECTIVE.md`. Only the `[benchmark]` section is required.

```
[benchmark]
primary_metric = "p50_ms"
direction = "minimize"
metrics_format = "jsonl"

[service]
base_url = "http://localhost:8080"
health_check = "GET /health"
readiness_timeout_sec = 60

[workspace]
candidate_path = "src"
requires_model_weights = false
externally_managed = false

[setup]
needs_docker_socket = false
timeout_sec = 300

[profiler]
kind = "none"
```

Field notes:

- `benchmark.primary_metric` and `benchmark.direction` are required. `direction` is `minimize` or `maximize`. `metrics_format` defaults to `jsonl`.
- `service.health_check` must be `METHOD path`, for example `GET /health`.
- `workspace.requires_model_weights` defaults to false for generic examples.
- `workspace.externally_managed` set to true supports targets that have no `reference/` directory at all.
- `setup.needs_docker_socket` set to true bind-mounts the host docker socket into the sandbox. This is rejected when running with `--modal`.
- `profiler.kind` is `none` by default for generic examples.

## Lifecycle scripts

All scripts live at the example root and receive the environment variables described below.

- setup.sh runs once when the container starts. Use it to start dependencies, for example `docker compose up`. It must exit 0 and should block until the service is ready.
- build.sh runs once right after setup.sh. Use it to compile checker or benchmark tools.
- check.sh runs correctness checks. Exit 0 means pass, nonzero means fail. It is surfaced to the agent through the judge prompt.
- bench.sh runs the benchmark and prints JSONL metrics. The framework parses its output for the primary metric.
- teardown.sh runs on session close for cleanup. It is best effort and never fails the run.

check.sh and bench.sh are required. setup.sh, build.sh, and teardown.sh are optional.

## Environment variables

Every lifecycle script receives these, and they are also present in the container environment so agent-run scripts see them:

- VIBESERVE_BASE_URL comes from `service.base_url`.
- VIBESERVE_OUTPUT_DIR is `/workspace/example_output`. bench.sh should write its JSONL output here or to stdout.
- VIBESERVE_LOAD_LEVEL is a load hint, default `medium`.

## Benchmark metrics

bench.sh emits one JSON object per line:

```
{"metric": "p50_ms", "value": 5.64}
{"metric": "throughput_rps", "value": 1180.0}
```

Lines that are blank, not JSON, or missing a string metric or numeric value are ignored, so bench.sh may print human-readable logs alongside metrics. The value of `benchmark.primary_metric` is the number used for comparison. For the evolve loop, ranking uses that metric with the declared direction.

## To Note

- Fixed an ordering bug in `_RunContext.__init__` where `example_manifest` was read during profiler resolution before it was assigned. This broke every `_RunContext` construction and surfaced once the full suite could run.
- The evolve profiler path no longer falls through to the nsys template when the profiler kind is none.
- `--modal` combined with `setup.needs_docker_socket = true` now fails fast with a clear message instead of silently running without a docker socket.

## Known limitations

- `service.health_check` and `service.readiness_timeout_sec` are parsed and available, but the framework does not itself poll the health endpoint. Readiness is expected to be handled inside setup.sh, which must block until the service is ready and then exit 0. 
- No example in the repository ships a `vibeserve.example.toml` or lifecycle scripts yet.  The generic path is therefore covered by unit and integration tests using a fake sandbox, not by a live end-to-end run inside this repository for any of the examples, which could be tested after integration.

## Setup

The project uses uv.

```
cd vibe-serve
uv sync --dev
```

If uv is not installed:

```
python -m pip install uv
```

On some machines the librosa dependency chain builds llvmlite from source and needs a system LLVM. If uv sync fails on llvmlite, install LLVM first, for example with `brew install llvm` on macOS, then run uv sync again. This is a local toolchain issue only and does not affect CI.

## Running the tests

A quick summary is given below:
- Full suite passes (819 tests), overall coverage ~79%
- ruff lint + format clean

Full suite with coverage, the same command CI runs:

```
uv run python -m pytest -v --cov=vibe_serve --cov=vs_issue_board --cov-report=term-missing --cov-fail-under=75
```

Only the tests added or touched by this issue:

```
uv run python -m pytest \
  tests/test_example_manifest.py \
  tests/test_issue_76_fixes.py \
  tests/test_issue_76_generic_examples.py \
  tests/test_run_environment_manifest.py \
  tests/test_agent_loop_prompt_wiring.py \
  tests/test_evolve_loop_prompt_wiring.py \
  tests/test_plain_loop_prompt_wiring.py \
  tests/test_openevolve_reuses_evolve_prompts.py \
  -v
```

Lint and format, the same commands CI runs:

```
./scripts/check_lint.sh
./scripts/check_format.sh
```

## Trying an example end to end

No generic example ships in the repo yet, but you can confirm the existing legacy examples still run. This runs the queue-default correctness checker and throughput benchmark on a plain Python candidate:

```
EX=examples/data-structures/queue-default
mkdir -p /tmp/qd
printf 'import sys\nsys.path.insert(0, "%s/reference")\nfrom reference import QueueFactory\n\n\nclass VibeServeQueue:\n    def __init__(self, scenario, capacity):\n        self._q = QueueFactory(scenario, capacity)\n\n    def enqueue(self, item, *a, **k):\n        return self._q.enqueue(item, *a, **k)\n\n    def dequeue(self, *a, **k):\n        return self._q.dequeue(*a, **k)\n\n    def size(self):\n        return self._q.size()\n\n    @property\n    def capacity(self):\n        return self._q.capacity\n' "$PWD/$EX" > /tmp/qd/main.py
PYTHONPATH=/tmp/qd uv run python "$EX/accuracy_checker/checker.py" --scenario all --ops 2000
uv run python "$EX/benchmark/benchmark.py" --use-reference --scenario all --duration 1 --warmup 0.2
```

To try a real generic target, create an example directory that contains `vibeserve.example.toml`, the required check.sh and bench.sh, and any of the optional setup.sh, build.sh, teardown.sh, then run:

```
uv run vibe-serve --outer-loop evolve --example examples/<your-example>
```

### DIFF
--- a/src/vibe_serve/cli.py
+++ b/src/vibe_serve/cli.py
@@ -160,7 +160,7 @@ def _add_common_args(parser: argparse.ArgumentParser) -> None:
     )
     parser.add_argument(
         "--profiler",
-        choices=["nsys", "torch", "neuron", "auto"],
+        choices=["nsys", "torch", "neuron", "none", "auto"],
         default="auto",
         help=(
             "Which profiler to use between rounds. "
@@ -577,7 +577,16 @@ def _load_objectives_toml(reference_path: str) -> list:
 def _resolve_objectives(args: argparse.Namespace) -> list:
     if args.objective:
         return list(args.objective)
-    return _load_objectives_toml(args.ref)
+    from_toml = _load_objectives_toml(args.ref)
+    if from_toml:
+        return from_toml
+    from vibe_serve.example_manifest import ExampleManifest
+
+    ref = Path(args.ref).expanduser().resolve()
+    manifest = ExampleManifest.detect(ref)
+    if manifest is not None:
+        return [manifest.to_objective()]
+    return []
 
 
 def _build_evolve_parser() -> argparse.ArgumentParser:
@@ -881,8 +890,33 @@ _RUNNERS = {
 }
 
 
+def _expand_example_flag(argv: list[str]) -> list[str]:
+    """Expand ``--example <dir>`` into --ref/--acc-checker/--bench (issue #76).
+
+    Pure sugar: fills in only the flags the user didn't already pass
+    explicitly, following the ``examples/<name>/{reference,accuracy_checker,
+    benchmark}`` convention. Detection of generic-vs-legacy example
+    behavior itself happens later, in _RunContext, based on whether
+    vibeserve.example.toml is present next to --ref -- not based on
+    whether --example was used to construct these paths.
+    """
+    example_dir, rest = _extract_flag(argv, "--example")
+    if example_dir is None:
+        return argv
+    example_path = Path(example_dir)
+    additions: list[str] = []
+    if "--ref" not in rest and not any(t.startswith("--ref=") for t in rest):
+        additions += ["--ref", str(example_path / "reference")]
+    if "--acc-checker" not in rest and not any(t.startswith("--acc-checker=") for t in rest):
+        additions += ["--acc-checker", str(example_path / "accuracy_checker")]
+    if "--bench" not in rest and not any(t.startswith("--bench=") for t in rest):
+        additions += ["--bench", str(example_path / "benchmark")]
+    return rest + additions
+
+
 def main() -> None:
-    loop_kind, remaining = _extract_loop_selection(sys.argv[1:])
+    argv = _expand_example_flag(sys.argv[1:])
+    loop_kind, remaining = _extract_loop_selection(argv)
     args = _PARSER_BUILDERS[loop_kind]().parse_args(remaining)
     # Resolve validator + runner via globals() so unittest.mock.patch on
     # ``vibe_serve.cli._{validate,run}_<kind>`` takes effect.
--- a/src/vibe_serve/context.py
+++ b/src/vibe_serve/context.py
@@ -22,6 +22,7 @@ from vibe_serve.constants import (
     PROJECT_ROOT,
     ComputeBackend,
 )
+from vibe_serve.example_manifest import ExampleManifest
 from vibe_serve.llm_client import _build_model
 from vibe_serve.sandbox.run_environment import (
     RunEnvironmentRequest,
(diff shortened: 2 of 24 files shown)


vic-lsh:
Closing as it is partially completed by other PRs already.

PR #137, opened 2026-07-11
Add interactive run server and Ink client

vic-lsh:
## Problem

VibeServe sessions can run for hours or days, but today the human interaction model is effectively limited to watching terminal output until the process finishes. Long-running optimization sessions need a durable way for operators to understand the run without stopping it.

This PR lays the foundation for richer human participation. Near-term interactions include viewing a synopsis of prior optimization rounds and asking a separate agent what the run is doing or why a result changed. These interactions need to remain auditable and must not couple the lifetime of the VibeServe session to the lifetime of one terminal UI.

The same foundation should support more than one presentation surface. The initial client is a TUI, but future clients could include a website, an Electron application, or other operator tools without reimplementing run-control semantics in each frontend.

### High-level design: one server, many possible frontends

Introduce a client-neutral Python run-control server with a versioned message protocol, plus an initial TypeScript Ink client. The server is the authoritative owner of run state, pause/resume semantics, audit persistence, snapshots, and sequenced events. Presentation clients own rendering and communicate with the server through JSONL messages over a private Unix socket.

This separation establishes the architecture for multiple clients to attach to the same session, for clients to replay events from a cursor after reconnecting, and ultimately for a persistent VibeServe session to survive an accidental client exit and accept a dynamically attached replacement client. This PR provides the server/client boundary, multi-client transport, and replay primitives; a user-facing attach workflow and non-terminal frontends remain future work.

Interactive mode starts automatically for TTY sessions, while `--headless` preserves backend and CI execution. All backend and subprocess output is converted into typed, monotonically sequenced messages rather than rendered by Python. The Ink client owns terminal presentation, chat input, live output, history, and the deliberately small slash-command surface (`/pause`, `/resume`, `/live`, `/history`, `/help`).

Pydantic generates the committed JSON Schema and TypeScript protocol types so future clients can share the same validated contract. CI regenerates these bindings and fails on contract drift, then applies strict TypeScript checking, frontend tests, and a production build. Human chat and status queries are recorded in the run history without altering agent prompts.

## Verification

Validated the rebased branch against the current `origin/main` with Python formatting/lint, the full Python suite, generated-schema consistency, TypeScript strict checking, TypeScript tests, and a production Ink build.

### Correctness properties

- Human chat is audited but never injected into an agent prompt.
- Pause takes effect only after the current agent call and resume releases the safe-point wait.
- Interactive Python and descendant-process stdout/stderr become sequenced `output` events rather than terminal writes.
- Event sequence numbers are unique and replayable by cursor across multiple clients.
- Interactive mode is selected only when stdin and stdout are TTYs; `--headless` and non-TTY runs do not launch the client.
- Concurrent frontend requests are correlated by request ID even when responses arrive out of order, and pending requests fail cleanly on disconnect.
- The live terminal viewport retains a persistent input prompt while rendering the newest output.

### Testing

- `./scripts/check_format.sh`
- `./scripts/check_lint.sh`
- `uv run pytest -q` — 1082 passed
- `uv run python -m pytest -q tests/test_tui.py` — 12 passed
- `pnpm --dir clients/tui generate:protocol`
- `pnpm --dir clients/tui check`
- `pnpm --dir clients/tui test` — 9 passed
- `pnpm --dir clients/tui build`
- `pnpm --dir clients/tui install --frozen-lockfile`
- `git diff --check`


vic-lsh:
Superseded by #138, which contains the current OpenTUI launcher/event-stream implementation and updated verification.

PR #138, opened 2026-07-11
Add OpenTUI supervision client and event-stream server

vic-lsh:
## Problem

VibeServe sessions can run for hours or days, but today the human interaction model is effectively limited to watching terminal output until the process finishes. Long-running optimization sessions need a durable way for operators to understand the run without stopping it.

This PR lays the foundation for richer human participation. Near-term interactions include viewing a synopsis of prior optimization rounds, inserting steering messages, and asking another agent what the run is doing or why a result changed. These interactions need to remain auditable and should not require presentation clients to own run-control semantics.

The same foundation should support more than one presentation surface. The initial client is a TUI, but future clients could include a website, an Electron application, or other operator tools without reimplementing session state and event processing in each frontend.

### High-level design: one server, many possible frontends

Introduce a client-neutral Python supervision server with a versioned message protocol, plus an initial TypeScript OpenTUI client. The Python backend is the authoritative owner of run state, audit persistence, snapshots, and sequenced events. Presentation clients own rendering and communicate with the backend through typed JSONL messages over a private Unix socket.

This boundary establishes the architecture for multiple clients to attach to the same session, clients to replay events from a cursor after reconnecting, and ultimately a persistent VibeServe session to survive an accidental client exit and accept a dynamically attached replacement client. This PR deliberately ties one backend to one frontend through a small Python launcher; multi-client attachment and independently persistent sessions remain future work.

`./vs` prepares the TypeScript client when needed and starts the launcher. The launcher owns a headless backend process and an OpenTUI process, waits for socket readiness, reports backend startup failures, and terminates the backend process group when the frontend exits. Direct `uv run vibe-serve` remains compatible and delegates interactive TTY runs to the same launcher, while `--headless` and non-TTY execution continue without a TUI.

Agent output, prompts, analysis, tool interactions, framework subprocess output, judge results, benchmark results, phase transitions, and round results are emitted as typed events rather than captured from global stdout/stderr. Python retains plain-text headless logging; OpenTUI owns interactive presentation. Pydantic generates the committed JSON Schema and strict TypeScript types so future clients can share the same validated contract.

The TypeScript client separates event processing from presentation. A pure session model/controller handles replay, ordering, conversation turns, round-history summaries, and the intentionally small command surface (`/help` and `/history`). OpenTUI renders Markdown prompts and assistant turns, paired tool call/result cards, bounded previews with expansion controls, keyboard scrolling, a persistent input, and live-view navigation.

## Verification

Validated the complete branch with repository formatting and lint checks, the full Python suite, generated-schema consistency, strict TypeScript checking, all frontend tests, a production OpenTUI build, and a real launcher/backend/frontend smoke run using the deterministic stub agent.

### Correctness properties

- The backend and frontend communicate through typed, monotonically sequenced events over a Unix socket; interactive rendering does not depend on global stdout/stderr capture.
- Event sequence numbers are unique and replayable by cursor, and replayed events are idempotently reduced by the client.
- Frontend state/event processing has no OpenTUI dependency; OpenTUI owns presentation and keyboard/mouse interaction only.
- A tool call and its corresponding result form one conversation card, while the next tool call starts a new card.
- Prompt and tool previews are presentation-only: full content remains in session state and persisted run history.
- `/history` derives completed or running round durations from audited event timestamps; `/help` is local and does not contact the backend.
- Enter submits the OpenTUI input through its concrete `ENTER` event, and Escape returns non-live views to current progress.
- The launcher terminates the backend process group when the TUI exits; backend failures are surfaced from a private log without corrupting terminal rendering.
- Interactive mode is selected only for TTY sessions; `--headless` and non-TTY runs preserve backend-only behavior.
- Generated TypeScript protocol types and the committed JSON Schema must match the Python Pydantic contract.

### Testing

- `./scripts/check_format.sh` — passed
- `./scripts/check_lint.sh` — passed
- `uv run pytest -q` — 1,083 passed
- `pnpm --dir clients/tui install --frozen-lockfile` — passed
- `pnpm --dir clients/tui generate:protocol` plus generated-file diff check — no drift
- `pnpm --dir clients/tui check` — strict TypeScript check passed
- `pnpm --dir clients/tui test` — 23 passed
- `pnpm --dir clients/tui build` — passed
- `git diff --check` — passed
- `./vs --outer-loop agent --input examples/model-serving/moonshine-streaming --max-rounds 1 --exp-name pr-open-smoke --stub-agent` — passed; TUI rendered the streamed run and both processes exited cleanly


vic-lsh:
Launcher follow-up: `./vs` now explicitly sources nvm and runs `nvm use node` for interactive launches. Backend failures are reported even when the TUI exits first, while normal frontend-first teardown gives Python a grace period and no longer reports SIGTERM as a crash. Verified with `bash -n vs`, Ruff, `uv run pytest tests/test_launcher.py tests/test_cli.py -q` (36 passed), strict TypeScript checks, 23 TUI tests, a production build, and a repeated no-preloaded-Node stub-agent smoke run.

vic-lsh:
Correction to the launcher follow-up: `./vs` does not assume nvm or proactively activate Node. nvm is only an optional fallback when a stale/missing frontend build actually requires rebuilding. Before launching OpenTUI, Python now resolves the configured agent backend/provider and exits with a direct error if the required host agent CLI (for example `codex`) is absent from PATH. Stub, Docker, Modal, and deepagents runs bypass the host-CLI check. Verified with shell syntax, Ruff, and 39 launcher/CLI tests including a simulated missing-agent executable.

PR #146, opened 2026-07-13
Add microservices domain

Ryclic:
## Problem

VibeServe doesn't yet include a domain for microservices. There are microservice examples in the repo, but they currently use the `generic` domain. Adding the microservices domain base framework allows for reusable templates and broader support of optimization for microservices.

## Solution

Add a registered `microservices` domain with role templates for implementer, judge, orchestrator, and profiler behavior.

This new domain is intentionally lightweight for now, it makes general changes. This means:
- no new manifest schema
- no service lifecycle automation
- no benchmark or telemetry framework changes

These changes will be added in a later PR. Please note that the current template MD files are subject to change as microservice functionality is extended (eg. standardized load balancing + telemetry).

The current PR provides reusable prompt guidance for microservice optimization while keeping task-specific API contracts, ports, startup steps, and workloads in each input bundle. The existing Train Ticket and Social Network microservice examples now select `[agent].domain = "microservices"` in order to move them off the "generic" domain.

Domain tests were extended to cover registration, resolution, prompt rendering, environment hook behavior, profiler compatibility, and the updated microservice example manifests.

## Verification

All the validation tests run + pytest execution of the new tests.

### Correctness properties

- `microservices` is a valid `DomainName`.
- The `microservices` domain is registered and resolves to its template directory.
- The domain uses no-op environment hooks, so it does not change workspace materialization, mounts, or runtime setup.
- The domain does not claim Torch profiler compatibility.
- Existing microservice input bundles load with `domain = "microservices"`.
- Microservice guidance remains prompt-only and does not alter evaluator
  commands, benchmark commands, or checker behavior.

### Testing

`uv run pytest tests/loops/agent/test_domain_packs.py` -> 23 passed

### DIFF
--- a/examples/microservices/social-network-read-timeline/vibeserve.input.toml
+++ b/examples/microservices/social-network-read-timeline/vibeserve.input.toml
@@ -1,7 +1,7 @@
 version = 1
 
 [agent]
-domain = "generic"
+domain = "microservices"
 
 [accuracy]
 command = ["./accuracy_checker/checker"]
--- a/examples/microservices/train-ticket/vibeserve.input.toml
+++ b/examples/microservices/train-ticket/vibeserve.input.toml
@@ -1,7 +1,7 @@
 version = 1
 
 [agent]
-domain = "generic"
+domain = "microservices"
 
 [accuracy]
 command = ["uv", "run", "python", "accuracy_checker/checker.py"]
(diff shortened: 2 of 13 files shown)


vic-lsh:
LGTM

PR #242, opened 2026-07-25
Skip GPU reselection for Docker environments (closes #24)

### LINKED ISSUE #24: Skip GPU reselection when environment is stable
## Summary

`reselect_gpu()` is called before every agent phase (implementer, judge, perf_eval) in both loops. On stable GPU environments (single-GPU, no contention), this triggers unnecessary container restarts and device checks.

## Current behavior

```python
# plain loop — called 3+ times per iteration
ctx.reselect_gpu()  # before implementer
ctx.reselect_gpu()  # before judge  
ctx.reselect_gpu()  # before perf_eval
```

## Proposed improvement

Only call `reselect_gpu()` when:
- The run environment signals `host_device_reselect=True` (already partially gated in `context.py:741-743`)
- A configurable interval has passed (e.g., reselect every N rounds instead of every phase)
- A GPU error or contention event was detected

For containerized environments (Docker, Modal) where the GPU is pinned at container start, reselection within a round is unnecessary.

## Impact

Low — saves some container restart overhead. Low effort.

zatchbell1311-wq:
## Problem
`reselect_gpu()` is called before every agent phase (implementer, judge, perf_eval) across all three loops. In Docker environments the GPU is pinned at container start, so these calls trigger unnecessary container restarts and device checks on every iteration.

## Solution
Set `host_device_reselect=False` in `DockerEnvironment.open()`, mirroring the existing Modal behaviour. The gate in `DeviceLease.reselect()` (line 63) already handles the early-return when this flag is `False` — no loop-level changes needed.

**One line changed** in `src/vibesys/sandbox/run_environment.py`:
```python
# Before
cli_sandboxed=True,
env_kind="docker",

# After
cli_sandboxed=True,
host_device_reselect=False,
env_kind="docker",
```

## Verification
- [x] Pattern matches existing Modal implementation at line 410
- [x] `DeviceLease.reselect()` gate confirmed at line 63 of `run/device.py`
- [ ] Manual Docker run confirming no reselection overhead

Closes #24

### DIFF
--- a/src/vibesys/agents/deepagents_runner.py
+++ b/src/vibesys/agents/deepagents_runner.py
@@ -55,6 +55,55 @@ class DeepAgentsRunner:
         self._skills = skills
         self._model_name = model_name
         self._run_log_file = run_log_file
+        # Cache the built agent graph + its checkpointer per kind, so the
+        # underlying agent graph (tools, model binding, skills wiring) is
+        # only rebuilt when the inputs that actually define it change.
+        # Keyed by (kind, system_prompt, tuple of tool names) so a change
+        # in either triggers a rebuild rather than silently reusing a stale
+        # agent. Each invocation still gets a fresh thread_id (below) so
+        # conversation context does not leak across rounds even though the
+        # graph itself is reused.
+        self._agents: dict[str, Any] = {}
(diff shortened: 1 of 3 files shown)


vic-lsh:
Thanks for the PR, but we've closed the original github issue (see that issue)

PR #243, opened 2026-07-25
Materialize skills once per workspace instead of per invoke (closes #20)

### LINKED ISSUE #20: Materialize skills once at loop start instead of per-invoke
## Summary

`_materialize_skills()` in `cli_runner.py:212` runs on **every** `invoke()` call, copying skill directories into 5 CLI skill-discovery paths (`.claude/skills`, `.agents/skills`, `.gemini/skills`, `.cursor/skills`, `.opencode/skills`). Skills don't change during a run, so this is redundant I/O.

## Current behavior

```python
# cli_runner.py:212 — called on EVERY invoke()
_materialize_skills(workspace, self._skills, log_file=self._run_log_file)
```

## Proposed improvement

Call `_materialize_skills()` once during `CliAgentRunner.__init__()` or at the start of the outer loop, then skip on subsequent invocations. Alternatively, check file mtimes before copying.

## Impact

Medium — reduces filesystem I/O per invocation. Low effort.

zatchbell1311-wq:
## Problem
`_materialize_skills()` was called on every `invoke()`, copying skill directories into 5 CLI skill-discovery paths (`.claude/skills`, `.agents/skills`, `.gemini/skills`, `.cursor/skills`, `.opencode/skills`) on every agent phase. Skills don't change during a run, so this was redundant I/O multiplied across every round.

## Solution
Add `_materialized_workspaces: set[Path]` to `CliAgentRunner.__init__()` and guard the call site — skills are materialized on the first `invoke()` per workspace and skipped on all subsequent calls.

```python
# Before — called on every invoke()
_materialize_skills(workspace, self._skills, log_file=self._run_log_file)

# After — called once per workspace
if workspace not in self._materialized_workspaces:
    _materialize_skills(workspace, self._skills, log_file=self._run_log_file)
    self._materialized_workspaces.add(workspace)
```

## Notes
- Per-workspace keying is correct: different workspaces still get their own materialization
- Skills changing mid-run is not a supported use case per the existing docstring

## Verification
- [x] Cache initialized in `__init__` at line 224
- [x] Guard applied at call site line 331-333
- [ ] Manual run confirming skills present on first invoke and skipped on subsequent

Closes #20

### DIFF
--- a/src/vibesys/agents/cli_runner.py
+++ b/src/vibesys/agents/cli_runner.py
@@ -221,6 +221,7 @@ class CliAgentRunner:
         # the exception: its history is carried explicitly in each prompt, so
         # it must not depend on provider-side session state.
         self._agents: dict[str, CodingAgent] = {}
+        self._materialized_workspaces: set[Path] = set()
 
     def invoke(
         self,
@@ -327,7 +328,9 @@ class CliAgentRunner:
     ) -> str:
         """Run one CLI generation with shared setup, logging, and cleanup."""
         label = _agent_label(kind)
-        _materialize_skills(workspace, self._skills, log_file=self._run_log_file)
+        if workspace not in self._materialized_workspaces:
+            _materialize_skills(workspace, self._skills, log_file=self._run_log_file)
+            self._materialized_workspaces.add(workspace)
 
         logger = AgentLogger(
             log_file=self._run_log_file,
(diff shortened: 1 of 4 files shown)


vic-lsh:
Thanks for the PR! This issue is no longer relevant under the new codebase and will be closed

PR #245, opened 2026-07-25
Add --plateau-stop flag for early termination on detected plateau (closes #21)

### LINKED ISSUE #21: Add early termination when plateau is detected and round passed
## Summary

The agent loop (`loops/agent/loop.py:835`) always runs the full `max_rounds` budget, even when the plateau detector identifies that performance has stagnated and the round passed. The comment at line 878-882 explains the `done` field was removed because the orchestrator can't reliably self-assess — but the plateau detector provides an objective signal that could be used.

## Current behavior

```python
# loop.py:878-882
# No early stop: the loop always consumes the full max_rounds
# budget. Previously OrchestratorPlan had a ``done`` field that
# could halt the loop; it was removed because the orchestrator
# can't reliably tell when the objective is "fully met" and
# early-stopping masks further optimization opportunities.
```

## Proposed improvement

Add a `--plateau-stop` flag (default off for backwards compat). When enabled:
- If the plateau detector triggers (3+ consecutive rounds within 5% spread) **and** the current round passed the judge, emit a warning and stop early.
- Log the plateau metrics so users can decide whether to resume with `--start-round` for another attempt.

This doesn't require the orchestrator to self-assess — it uses the same objective metric comparison that the plateau detector already performs.

## Impact

Medium — saves wasted rounds when the harness has converged. Low effort.

zatchbell1311-wq:
## Problem
The agent loop always consumed the full `max_rounds` budget even when the plateau detector identified that performance had stagnated and the current round passed the judge. The round budget was being wasted on iterations that the objective metric confirmed were not making progress.

## Solution
Add an opt-in `plateau_stop: bool = False` parameter to `run_agent_loop()`. When enabled, if the plateau detector triggers **and** the current round passed the judge, the loop logs the plateau metrics with a `--start-round` resume hint and stops early.

```python
if plateau_stop and passed and plateau_warning is not None:
    ctx.lprint(
        f"\n[plateau-stop] Stopping early — plateau detected and round passed.\n"
        f"{plateau_warning}\n"
        f"Resume with --start-round {round_number + 1} to continue."
    )
    return False
```

## Notes
- Default is `False` for full backwards compatibility — existing callers are unaffected
- Uses the existing `_detect_plateau()` signal — no new heuristics introduced
- Only stops when both conditions hold: plateau detected **and** round passed — avoids stopping on a bad round where the agent just happened to stagnate

## Verification
- [x] `plateau_stop: bool = False` added to signature at line 876
- [x] Early-exit inserted after `_save_rounds_state()` at line 1197
- [x] Default False preserves existing behaviour

Closes #21

### DIFF
--- a/src/vibesys/agents/cli_runner.py
+++ b/src/vibesys/agents/cli_runner.py
@@ -221,6 +221,7 @@ class CliAgentRunner:
         # the exception: its history is carried explicitly in each prompt, so
         # it must not depend on provider-side session state.
         self._agents: dict[str, CodingAgent] = {}
+        self._materialized_workspaces: set[Path] = set()
 
     def invoke(
         self,
@@ -327,7 +328,9 @@ class CliAgentRunner:
     ) -> str:
         """Run one CLI generation with shared setup, logging, and cleanup."""
         label = _agent_label(kind)
-        _materialize_skills(workspace, self._skills, log_file=self._run_log_file)
+        if workspace not in self._materialized_workspaces:
+            _materialize_skills(workspace, self._skills, log_file=self._run_log_file)
+            self._materialized_workspaces.add(workspace)
 
         logger = AgentLogger(
             log_file=self._run_log_file,
--- a/src/vibesys/agents/deepagents_runner.py
+++ b/src/vibesys/agents/deepagents_runner.py
@@ -55,6 +55,55 @@ class DeepAgentsRunner:
         self._skills = skills
         self._model_name = model_name
         self._run_log_file = run_log_file
+        # Cache the built agent graph + its checkpointer per kind, so the
+        # underlying agent graph (tools, model binding, skills wiring) is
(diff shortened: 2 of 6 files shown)


zatchbell1311-wq:
The Tests/tui failure is a Bun v1.3.9 runtime segfault unrelated to this PR — all TUI tests passed before the crash. This appears to be a known flaky CI issue on the runner.

zatchbell1311-wq:
Same Bun v1.3.9 segfault as #244 — unrelated to the plateau-stop changes. Could you re-run the TUI check?

vic-lsh:
Thanks for resolving this. I think plateau in performance is actually a VibeSys performance issue; ideally some agent would realize that it is in a plateau and jump out of the local maxima. We've made some design changes recently to address this, so we're closing this PR.

PR #265, opened 2026-08-05
Add support for semantic themes for VibeSys Terminal UI

AyanBinRafaih:
## Problem

The TUI's colors are hex literals written at each call site — the app shell, conversation cards, overlays, input panel, setup screen, agent map, round strip, and todo strip. `ui/styles.ts` covers only conversation entries and Markdown; everything else carries its own.

There is no way to select a theme, so a light terminal or a projector leaves the client unreadable. There are also many differences between the cards, rounds strip, todos, agent phases and the chat panel's inner border not matching its outer one.

## Solution

Add a semantic theme layer that owns every color in the client, and include eight themes as four light/dark pairs: `dark` (default) / `light`, `solarized-dark` / `solarized-light`, `catppuccin-mocha` / `catppuccin-latte`, and `high-contrast-dark` / `high-contrast-light`. The user selects one of these given labels, as above, with the literal colour being included in `ui/theme.ts`.

**Selection:** in increasing precedence: `[tui].theme` in `agent.toml` (persistent), `--theme NAME` (per launch), and `/theme <name>` (current session only, no restart). `/theme` with no argument lists the themes and marks the active one. Precedence is computed in one place — the launcher forwards `--theme` to the Python `tui-defaults` subcommand, which resolves `args.theme or config.tui.theme`; Node only pre-validates the name so a typo exits 2 before any process is spawned.

**Handling:** Each theme declares a `ThemeSpec`, which is a semantic core plus one accent per conversation role and `buildTheme()` derives card fills, labels, body text, tool bands and the Markdown palette from it. Every derived foreground passes through `ensureContrast()` against the surface it actually lands on, walking toward white or black until it clears the theme's floor (4.5:1, or 7:1 for the high-contrast pair). Where the foreground has already saturated, `readableBandBackground()` pulls the background back toward the canvas instead.

**Live repaint:** Each view takes a `Theme` and exposes `applyTheme()`, which re-applies container props and invalidates its memo cache. Nothing is rebuilt, so scroll positioning in the agent traces and expanded prompts are left unchanged and static. The hazard is `SyntaxStyle`, an OpenTUI native handle: destroying the old one before the new is attached would free memory the renderer still references. `applyTheme` therefore returns a deferred release closure that `render()` invokes only after the rebuild pass.

The demonstration is as follows:

https://github.com/user-attachments/assets/89cf7f93-00e6-4fa3-b05d-f18b18af7517

### Architecture

```mermaid
flowchart TD
    C[agent.toml · tui.theme] --> P[Python tui-defaults<br/>args.theme or config.tui.theme]
    F["--theme NAME"] --> N[Node launcher<br/>validates name only]
    N --> P
    P --> D[InteractiveSetupDefaults.theme] --> E[VIBESYS_THEME]
    E --> S[setup.ts]
    E --> A[app.ts]
    T["/theme NAME"] --> M[session-model.setTheme] --> A
    A --> R[resolveTheme] --> V[applyTheme on every view]
```

Note: The setup screen runs as a separate process, so the launcher passes `VIBESYS_THEME` to both children.

`TuiTheme` (Python) and `THEME_NAMES` (TypeScript) are necessarily duplicated. `tests/test_tui_theme.py` parses the TypeScript source and asserts the name list and default match the enum, so they cannot drift.

## Verification

The main thing to check are if the themes render properly and contrast is present, with respect to the chosen theme. For this, every role in every theme is checked against the surface it's actually drawn on. All test-cases pass. 

Specific roles are listed below, not covered by the overall theme description:

| Site | Before | After | Role |
| --- | --- | --- | --- |
| Todo completed | `#4ade80` | `#22c55e` | `success` |
| Completed phase, prompt-disclosure hint | `#60a5fa` | `#38bdf8` | `info` |
| Round labels, agent-map heading, collapsed todo | `#cbd5e1` | `#e2e8f0` | `textPrimary` |
| Chat inner Message border | `#8b5cf6` | `#a78bfa` | `conversation.analysis.label` |

### Correctness properties

- One selected theme styles every surface: app, setup screen, overlays, input, conversation cards, Markdown, agent map, rounds, todos.
- No module outside `ui/theme.ts` contains a color literal.
- Selecting `dark` reproduces the pre-theme appearance apart from the four merged shades above.
- Every role clears its theme's contrast floor against its own surface, enforced at build time so new themes inherit the guarantee.
- Status never depends on color alone: agent phases carry a glyph and the spelled-out status; todos carry per-status glyphs with a fallback for unknown wire values; judge and round verdicts appear in the text label; overlays are titled.
- Light themes place canvas luminance above text luminance, dark themes below; borders stay ≥1.7:1 against the canvas.
- Precedence is `--theme` > `[tui].theme` > `dark`, resolved in one place; an unknown flag value fails before any process starts.
- An unknown theme name from any source resolves to the default rather than crashing the client.
- `/theme` does not mutate configuration, and a live switch preserves scroll position and expanded-prompt state without freeing an attached `SyntaxStyle`.
- Themes are compile-time constants: no file read, no `JSON.parse`, no dynamic import, no theme path in the config schema.

### Testing

- `bun --bun vitest run` (in `clients/tui`) — 153 passed, 13 files. Adds `theme.test.ts` (17 blocks × 8 themes via `it.each` = 52 cases), 4 live-repaint tests in `app.test.ts`, 8 `/theme` assertions in `session-controller.test.ts`, and a `launcher.test.ts` case asserting `VIBESYS_THEME` reaches both child processes.
- `uv run pytest -q` — 2669 passed, 11 skipped (pre-existing environment gates: bwrap, `codex` CLI, Go/Rust/C++ toolchains).
- `uv run pytest -q tests/test_tui_theme.py tests/test_config.py tests/test_main.py` — 150 passed. Covers config default and rejection of unknown themes/keys, `--theme` overriding `[tui].theme`, and Python/TypeScript name-list agreement.
- `npx tsc -p tsconfig.check.json --noEmit` and `npx tsc -p tsconfig.json --noEmit` — passed.
- `npx biome check .` — passed, 39 files.
- `./scripts/check_format.sh` — passed, 296 files. `./scripts/check_lint.sh` — passed.
- `grep -rlE '#[0-9a-fA-F]{6}' --include='*.ts'` over non-test sources — returns `ui/theme.ts` only.


vic-lsh:
@AyanBinRafaih Thanks for this PR! Could you fix the CI errors? Thank you!

AyanBinRafaih:
@vic-lsh I've made a few changes to code to prevent bun spawning the entire test suite with all the files and the CI checks now pass.

vic-lsh (approved):

Issue #293, opened 2026-08-08
Mechanical enforcement of documented code standards


vic-lsh:
## Purpose

Group the work that turns VibeSys's documented code standards into mechanically enforced ones, and closes the scope and granularity gaps in the checks that already exist.

This is the **mechanical enforcement** work area of roadmap #288. It exists because the repository's static-analysis and coverage setup is strong but uneven: the standards in `docs/coding-best-practices.md` are largely enforced by reviewer memory rather than by a failing check, and the checks that do run leave whole trees and modules outside their scope.

## Problem

Two distinct shapes of gap, both general:

**Standards live in prose rather than in checks.** `docs/coding-best-practices.md` mandates small functions, no blind `except`, typed cross-boundary values, and no reaching through interfaces. None of these is machine-enforced today. Pyright cannot close the gap — it reports zero errors on the current tree and structurally cannot express most of these properties, since explicit `Any` is a deliberate escape hatch and complexity is not a type question.

**Existing checks have scope holes and averaging blind spots.** Type checking, formatting, and coverage each omit a first-party tree or let a module hide behind an aggregate. The result is that a check can be green while the code it nominally governs is unexamined.

## Sub-issues

- **#295** — Enable the ruff rule families the best-practices doc mandates, with a self-tightening per-site suppression baseline and a pinned toolchain.
- **#296** — Extend format checking to `libs/`.
- **#297** — Bring `tests/` under pyright.
- **#298** — Close coverage scope and granularity gaps.

These are the currently identified increments. They do not bound this area; further enforcement work attaches here as it is found.

## Scope

This issue is a grouping container and does not itself change behavior. Each sub-issue should be independently closable and independently reviewable.

Enforcement changes should adopt via ratchets rather than one-time baselines wherever possible, so that a suppression cannot silently become permanent. Where a check must be relaxed, the relaxation should be explicit, scoped, and documented with its reasoning — following the precedent of the existing pyright relaxation comment in `pyproject.toml`.

## Non-goals

- Substantive refactors that newly enabled checks will flag. Those are separate issues under #288; this area baselines them.
- Weakening any existing standard. The strict bar for first-party source must not regress in service of expanding scope.
- Biome and TypeScript rule changes for `clients/`.
- Test authoring to raise coverage numbers.

## Acceptance criteria

- [ ] Documented standards that are mechanically checkable are enforced by a check rather than by review alone.
- [ ] Static analysis, formatting, and coverage apply to every first-party source tree.
[...]

Issue #292, opened 2026-08-08
Cover the accuracy gate's failure and degradation paths


vic-lsh:
## Workstream

Core/runtime

## Problem

> **Correction (2026-08-08):** this issue originally claimed the accuracy gate was "never executed by the test suite." That was wrong. Measurement shows `src/vibesys/loops/gates.py` at **89% branch coverage** — `tests/loops/agent/test_orchestrate.py` calls `_run_framework_accuracy_gate` directly in seven tests (`:1208`, `:1346`, `:1365`, `:1384`, `:1406`, `:1430`, `:1448`), which exercises the real `run_accuracy_gate`. The original claim generalized from the evolve-loop tests alone. The real gap is narrower and is described below.

`src/vibesys/loops/gates.py` implements the accuracy gate — the correctness check the framework runs against a candidate system before accepting it. It is well covered overall, but the uncovered remainder is concentrated in exactly the paths that matter when something goes wrong.

Measured with `uv run pytest tests/loops/agent/test_orchestrate.py --cov=vibesys.loops.gates --cov-branch`:

```
Name                         Stmts   Miss Branch BrPart  Cover   Missing
src/vibesys/loops/gates.py      49      5     12      2    89%   46, 68-70, 103
```

The three uncovered regions:

- **`gates.py:68-70` — the broad-except degradation path.** `except Exception as exc:` converts any execution failure into `output = f"accuracy command could not be executed: {exc}"` and `passed = False`. This is the highest-value gap: it silently reclassifies "the gate could not run" as "the gate failed," and `docs/coding-best-practices.md` explicitly warns against "implicit fallback behavior for agent-visible contracts." Whether this fallback is correct is currently unverified.
- **`gates.py:46` — the no-command early return**, which reports `passed=True, executed=False`. A gate that passes because no command was configured is a meaningfully different outcome from a gate that passed on its merits, and nothing asserts that distinction.
- **`gates.py:103` — the `supervisor is None or not content` guard** in `_publish_subprocess_output`.

Separately, the two call sites are covered asymmetrically. The agent loop's wiring is exercised directly, but the evolve loop's is mocked out at `tests/loops/evolve/test_evolutionary_loop.py:182` and `:1144`, so the evolve loop's use of the gate is untested even though the gate itself is not.

## Desired outcome

[...]

Issue #288, opened 2026-08-08
Improve VibeSys codebase quality and robustness


vic-lsh:
## Purpose

Track the ongoing effort to keep the VibeSys codebase easy to change, hard to break, and honestly documented as the project grows.

VibeSys has expanded quickly along several axes at once: a Python framework, reusable libraries, a TUI client, a large example and evaluator corpus, and a skills library. Growth has been feature-led, which is appropriate for a research framework. This roadmap is the standing home for the internal-quality work that keeps that growth sustainable, so it stops arriving only as incidental cleanup discovered during unrelated feature work.

This is not a response to a neglected codebase. The repository already documents its standards in [`docs/coding-best-practices.md`](docs/coding-best-practices.md) and already enforces formatting, linting, strict type checking, coverage floors, client checks, and generated-contract drift checks in CI. The durable gap is narrower and more general: **the distance between the standards the repository documents and the standards it mechanically guarantees**, and the way that distance widens as new surfaces are added.

## Problem

Internal-quality work has no standing home, so it is discovered late and prioritized ad hoc. The recurring patterns are general rather than tied to any one module:

- **Standards live in prose rather than in checks**, so adherence depends on reviewer attention instead of a failing build.
- **Existing checks have scope and granularity gaps**, letting whole trees or modules escape enforcement by exclusion or by averaging.
- **Hand-synced artifacts drift**, while generated and verified ones stay correct.
- **Complexity and responsibility concentrate** in a few modules, with no signal identifying the next one to become a change bottleneck.
- **Weakly typed boundaries let contradictions become live defects** instead of being unrepresentable.
- **Documentation and onboarding decay silently**, because correctness of prose is not verified against code.

## Work areas

Durable categories, not a closed checklist. Each child issue should name one independently closable outcome within one area.

- **Structure and ownership** — module and function decomposition, layering direction, real interfaces in place of duck-typing and private reach-through.
- **Typed contracts and validation** — enums for closed sets, validated persisted state, rejection of unknown keys, errors that name the offending field.
- **Failure modes and resource lifecycle** — deterministic cleanup across success, failure, timeout, and cancellation; atomic persistence; idempotent teardown.
- **Duplication and external-CLI boundaries** — one authoritative implementation per rule, injectable runners for subprocess integrations.
- **Test effectiveness** — critical paths actually executed, contract-level assertions, shared contract tests for interchangeable implementations.
[...]

PR #299, opened 2026-08-08
Add libs to the ruff format target set

vic-lsh:
## Problem

`scripts/check_format.sh` and `scripts/format.sh` both set
`targets=(src tests examples resources)`, which omits `libs/` entirely.
That leaves ~6.3k lines of source and test code across the four
workspace libraries (`vs-feature-flags`, `vs-github`, `vs-issue-board`,
`vs-sandbox`) never checked by `ruff format --check`. `ruff check .` in
`scripts/check_lint.sh` already lints these paths (it runs `.` rather
than an explicit target list), so this gap is specific to the
formatter. Tracked by #296, under area #293 and roadmap #288.

## Solution

Added `libs` to the `targets` array in both `scripts/check_format.sh`
and `scripts/format.sh`, keeping the two scripts consistent (they
already shared an identical target list, so no other reconciliation
was needed).

`scripts/` and `clients/` were evaluated for inclusion but contain no
Python files (`find scripts -name '*.py'` and `find clients -name
'*.py'` both return zero results), so they are intentionally left out
of the target set.

### Architecture

Mechanical, single-boundary change: only the formatter's target list
in the two wrapper scripts. No change to `[tool.ruff.lint]`, pyright
config, or CI job structure, per the coordination note that a separate
PR is concurrently expanding the lint rule set and may also touch
`libs/` formatting.

## Verification

Ran `uv run ruff format --check --diff libs` before making any change
to measure scope: 4 files would be reformatted (24 already
compliant). Applied `./scripts/format.sh` with `libs` added, which
reformatted exactly those 4 files. All diffs are line-wrapping only
(long lines collapsed to fit under the 100-column limit); no logic
changed.

### Correctness properties

- Formatting-only change: no behavior, API, or contract changes.
- `scripts/format.sh` and `scripts/check_format.sh` remain in sync
  (identical `targets` arrays).
- Does not touch `[tool.ruff.lint]`, pyright config, or CI job
  structure.

### Testing

- `uv run pytest` (baseline, before change): 2948 passed, 1 skipped.
- `./scripts/format.sh`: reformatted 4 files (all under `libs/`).
- `./scripts/check_format.sh`: all checks passed (338 files).
- `./scripts/check_lint.sh`: all checks passed.
- `./scripts/check_types.sh`: 0 errors, 0 warnings, 0 informations.
- `uv run pytest` (after change): 2948 passed, 1 skipped — no
  regression.

Files reformatted, by path:
[...]

PR #301, opened 2026-08-08
Close coverage scope and granularity gaps (#293)

### LINKED ISSUE #298: Close coverage scope and granularity gaps
## Workstream

CLI/DX

## Problem

Coverage is measured and enforced — a 75% global floor and an 80% `diff-cover` patch floor in `.github/workflows/test.yml` — but three gaps let untested code through.

**1. A library is excluded from measurement.** CI runs `pytest --cov=vibesys --cov=vs_github --cov=vs_issue_board --cov=vs_sandbox`. `vs_feature_flags` is missing, so its source contributes nothing to the number even though `pyproject.toml` includes its tests in `testpaths`.

**2. Coverage is CI-only and line-only.** There is no `[tool.coverage.run]` or `[tool.coverage.report]` in `pyproject.toml`, so `uv run pytest` locally reports nothing, and branch coverage is off. Contributors cannot see their own coverage without reconstructing the CI flag string.

**3. The floor is a global average, which hides zero-coverage modules.** This is the substantive gap. Two modules pass today while being effectively untested:

- `src/vibesys/_model_config/` — roughly 360 lines with zero tests and zero callers anywhere in the repo.
- `src/vibesys/loops/gates.py` — the framework accuracy gate. **Correction:** this module is in fact ~89% covered, so it is *not* an example of the averaging problem; see #292 for the narrower, real gap. It is named here only because the original framing of this issue cited it in error.

`_model_config/` does not move a repository-wide average enough to breach 75%, so the strongest gate in the project is blind to a package that is entirely untested and entirely unreachable.

## Desired outcome

Coverage measurement covers every first-party package, is visible locally, and cannot be satisfied by an average that conceals an untested module. Modules intentionally below the floor are explicit and justified rather than invisible.

## Acceptance criteria

- [ ] `vs_feature_flags` is included in coverage measurement.
- [ ] `[tool.coverage.run]` exists with `branch = true` and a correct `source`, so `uv run pytest` reports coverage locally.
- [ ] A per-module floor exists, so a module at or near zero coverage fails CI regardless of the global average.
- [ ] Modules intentionally below the floor are listed in an explicit, commented allowlist; the failure message names the offending module and its actual coverage.
- [ ] The existing 80% patch floor remains intact.
- [ ] If enabling branch coverage changes the global percentage, the new number is reported and any threshold adjustment is deliberate and documented.

## Scope and non-goals

In scope: coverage configuration, the per-module gate and its allowlist, and the corresponding CI wiring.

Not in scope:

- Writing tests to raise coverage. `_model_config/` is a separate delete-or-adopt decision and belongs on the allowlist referencing that decision, not "fixed" here. `loops/gates.py` should NOT be allowlisted: it passes comfortably, and listing it would falsely imply it is untested.
- Deleting code to improve numbers.
- Ruff, pyright, and formatter scope — sibling issues.

- `coverage` has no native per-module `fail-under`, so this likely needs a small check over `coverage json` output wired into CI. Keep the floor configurable in one obvious place.
- Choose the floor from measured per-module figures rather than picking a round number, so it binds meaningfully without becoming a test-writing project.
- Enabling `branch = true` will probably lower the reported global percentage. Report the real figure rather than quietly relaxing the threshold.
- An allowlist entry is a deferral, not an exemption; each should reference the issue that will retire it.

## Verification approach

- Record global coverage before and after, noting the branch-coverage effect, plus per-module figures for the modules driving the floor.
- Demonstrate the gate works in both directions: it fails on a module below the floor, and passes once that module is allowlisted.
- `./scripts/check_format.sh`, `./scripts/check_lint.sh`, `./scripts/check_types.sh`, and `uv run pytest` all pass.

Parent: #293. Roadmap: #288. Related: #292 (accuracy gate coverage), which should retire one allowlist entry.

vic-lsh:
## Problem

Issue #298 (under the #293 mechanical-enforcement area, part of the #288 codebase-quality roadmap) named three coverage gaps that let the repo's stated standards drift from what CI actually enforces:

1. `vs_feature_flags` is absent from CI's `--cov` flags (`.github/workflows/test.yml:207`), so its ~116 source lines contribute nothing to the measured number even though `pyproject.toml` already runs its tests.
2. There is no `[tool.coverage.run]`/`[tool.coverage.report]` in `pyproject.toml`, so `uv run pytest` reports no coverage locally, and there is no branch coverage anywhere.
3. The 75% global floor is an average, so a module at or near zero coverage can hide inside a healthy total. `src/vibesys/_model_config/` demonstrates this concretely: ~360 lines at **0%** coverage with zero callers anywhere in the repo (verified by grepping `src/`, `tests/`, and `libs/`), and it passes CI today purely because the repo-wide average absorbs it.

`src/vibesys/loops/gates.py` was initially suspected of the same problem, on the grounds that the framework accuracy gate is mocked at its call sites. Measuring it disproved that: it sits at **89%** branch coverage (49 statements, 5 missed), because `tests/loops/agent/test_orchestrate.py` calls `_run_framework_accuracy_gate` directly in seven tests. Only the evolve-loop tests patch it out. It is therefore **not** allowlisted here — allowlisting a module at 89% would wrongly imply it is untested.

That distinction is worth flagging for reviewers, because it shows what a per-module floor does and does not buy: `gates.py` clears any sane floor, yet the specific uncovered lines are the broad `except Exception` degradation path at `gates.py:68-70` — exactly the untested agent-visible fallback behavior `docs/coding-best-practices.md` warns against. A percentage gate cannot see that; only looking at *which* lines are uncovered can. That narrower, real gap is tracked in #292.

## Solution

- Add `[tool.coverage.run]` (`branch = true`, `source` covering all five first-party packages including `vs_feature_flags`, `relative_files = true`, and an `omit` for the `python -m vibesys` launcher stub) and `[tool.coverage.report]` to `pyproject.toml`.
- Add `--cov --cov-report=term-missing` to `[tool.pytest.ini_options] addopts` so a bare `uv run pytest` reports coverage against the same source list, without the CI flag string.
- Simplify CI's `Run tests` step to rely on that source config instead of repeating `--cov=<package>` per package, and add `--cov-report=json` for the new per-module check.
- Add `scripts/check_coverage_floor.py`: reads a `coverage json` report and `[tool.vibesys.per_module_coverage]` from `pyproject.toml` (a `floor` percentage plus an explicit, commented `allowlist` of file paths or directory prefixes), and fails naming any non-allowlisted file below the floor or any stale (no-longer-matching) allowlist entry. Wired in as a new CI step, `Enforce per-module coverage floor`, right after the test step.
- Set the floor to 40% (see Correctness properties for how it was chosen), with `src/vibesys/_model_config/` as the sole allowlist entry, commented with its reason and issue reference.
- Update the `vs-issue-board` README's coverage example to the simplified command now that package coverage is centrally configured.

### Architecture

```mermaid
flowchart LR
  A[uv run pytest] -->|addopts: --cov, source config| B[coverage.py: branch=true, 5 packages]
  B --> C[term-missing report - local]
  B -->|CI only| D[coverage.xml]
  B -->|CI only| E[coverage.json]
  D --> F[diff-cover: 80% patch floor, unchanged]
  E --> G[scripts/check_coverage_floor.py]
  G -->|reads| H["[tool.vibesys.per_module_coverage] in pyproject.toml"]
  H --> I[floor: 40%]
  H --> J["allowlist: _model_config/"]
```

Coverage scope/threshold configuration lives entirely in `pyproject.toml` (one obvious place, per the issue's design ask); the CI workflow and the new script only read it.

## Verification

Ran the exact CI sequence locally (`uv run python -m pytest --cov-report=xml --cov-report=json --cov-fail-under=75 && uv run python scripts/check_coverage_floor.py coverage.json`): 2948 passed, 1 skipped, 75% floor reached at 85.31%, per-module check passed with `_model_config/` as the sole allowlist entry. Also ran `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80` locally (passes; this PR's own diff has no coverable lines since it only touches config/CI/a new unmeasured script).

Demonstrated the new gate actually gates: temporarily removed the `_model_config` allowlist entry and reran the check script against the real coverage report — it failed, naming both files at `0.0% < 40%`. Restored the entry and reran — passed again.

### Correctness properties

- **Global floor is unaffected, not silently lowered.** Baseline (current `main`, line coverage only, `vs_feature_flags` excluded, matching the exact CI command before this PR): 2948 passed, 1 skipped, **87.88%** total. After this PR (branch coverage, `vs_feature_flags` included): **85.31%** total. Branch coverage costs ~2.6 points as expected, but the result still clears the existing 75% floor comfortably, so 75% is left unchanged.
- **Per-module floor is set from measured data, not guessed.** The full per-file distribution was captured before choosing a number. Excluding the allowlisted `_model_config/` (0%), every module actually exercised by the suite measures at 53.75% or higher (`src/vibesys/agents/docker_executor.py`, the next-lowest). 40% sits with ~14 points of headroom below that module and well above 0%, so it catches genuinely untested modules without demanding new tests for anything currently passing.
- **The allowlist contains only modules that actually fail the floor.** `src/vibesys/_model_config/` is the sole entry, with an inline comment naming the reason and issue, matching the existing `[tool.ruff.lint] ignore` comment style. `loops/gates.py` measures 89% and is deliberately not listed.
- **Stale allowlist entries fail the build.** The script flags any exact-path entry that no longer matches a file in the coverage report, so an entry cannot silently outlive the module it was written for.
- **80% diff-cover patch floor is untouched** (`.github/workflows/test.yml` patch-coverage job).
- **No new tests were added and no code was deleted** to move these numbers; `_model_config/` and `gates.py` remain exactly as they were, tracked by their own issues (#298 and #292).

### Testing

- `./scripts/check_format.sh` — passed
- `./scripts/check_lint.sh` — passed
- `./scripts/check_types.sh` — passed
- `uv run python -m pytest --cov-report=xml --cov-report=json --cov-fail-under=75` — 2948 passed, 1 skipped, 85.31% total
- `uv run python scripts/check_coverage_floor.py coverage.json` — passed
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80` — passed (no coverable lines in this diff)
- Manual negative test: removed `src/vibesys/_model_config/` from the allowlist, reran the floor check against the real coverage report — failed, correctly naming `src/vibesys/_model_config/__init__.py: 0.0% < 40%` and `src/vibesys/_model_config/_core.py: 0.0% < 40%`; restored the entry, reran — passed.
- Script logic additionally exercised against synthetic reports for the pass, below-floor, directory-prefix-allowlist, zero-statement-file, and stale-entry cases.

Closes #298. Part of #293 and #288.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -212,7 +212,7 @@ jobs:
 
       - name: Enforce per-module coverage floor
         # Repo-wide --cov-fail-under is an average and can't see a single
-        # near-zero module hiding inside it (issue #293). Reads the floor and
+        # near-zero module hiding inside it (issue #298). Reads the floor and
         # allowlist from [tool.vibesys.per_module_coverage] in pyproject.toml.
         run: uv run python scripts/check_coverage_floor.py coverage.json
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -204,7 +204,17 @@ jobs:
         run: uv sync --dev
 
       - name: Run tests
-        run: uv run python -m pytest -v --cov=vibesys --cov=vs_github --cov=vs_issue_board --cov=vs_sandbox --cov-report=term-missing --cov-report=xml --cov-fail-under=75
+        # `addopts` in pyproject.toml already sets `--cov --cov-report=term-missing`
+        # against the source list in `[tool.coverage.run]` (vibesys, vs_feature_flags,
+        # vs_github, vs_issue_board, vs_sandbox). Add the report formats CI needs on
+        # top of that instead of repeating `--cov=...` for each package.
+        run: uv run python -m pytest -v --cov-report=xml --cov-report=json --cov-fail-under=75
+
+      - name: Enforce per-module coverage floor
+        # Repo-wide --cov-fail-under is an average and can't see a single
+        # near-zero module hiding inside it (issue #293). Reads the floor and
+        # allowlist from [tool.vibesys.per_module_coverage] in pyproject.toml.
+        run: uv run python scripts/check_coverage_floor.py coverage.json
 
       - name: Upload coverage report
         if: matrix.python-version == '3.12'
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,7 @@ htmlcov/
 .cache
 nosetests.xml
 coverage.xml
+coverage.json
 *.cover
 *.py.cover
 .hypothesis/
--- a/libs/vs-issue-board/README.md
+++ b/libs/vs-issue-board/README.md
@@ -137,8 +137,11 @@ When moving code into this package, move or add the tests that define that
 generic behavior here too. App-level compatibility tests can stay in the app
 test tree when they verify old import paths or VibeSys integration wiring.
 
-For the full repository gate, include package coverage with the app coverage:
+For the full repository gate, run the whole suite. `[tool.coverage.run]` in
+the root `pyproject.toml` already measures `vs_issue_board` alongside
+`vibesys` and the other workspace libraries, so a bare `uv run pytest`
+reports combined coverage locally; CI additionally enforces the floor:
 
 ```bash
-uv run python -m pytest -v --cov=vibesys --cov=vs_issue_board --cov-report=term-missing --cov-report=xml --cov-fail-under=75
+uv run pytest --cov-fail-under=75
 ```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,8 +67,66 @@ test = [
 
 [tool.pytest.ini_options]
 testpaths = ["tests", "libs/vs-feature-flags/tests", "libs/vs-github/tests", "libs/vs-issue-board/tests", "libs/vs-sandbox/tests"]
-addopts = "-ra"
+# Bare `--cov` measures the packages listed in `[tool.coverage.run] source`
+# below, so `uv run pytest` reports coverage locally without needing the full
+# flag string CI uses. CI layers on `--cov-report=xml`, `--cov-report=json`,
+# and `--cov-fail-under` on top of this; it does not need its own `--cov=...`
+# flags.
+addopts = "-ra --cov --cov-report=term-missing"
 pythonpath = ["."]
+
+[tool.coverage.run]
+branch = true
+source = ["vibesys", "vs_feature_flags", "vs_github", "vs_issue_board", "vs_sandbox"]
+relative_files = true
+omit = [
+    # `python -m vibesys` launcher stub: one import and a guarded call into
+    # `vibesys.main.main`, which is itself measured. Not exercised by unit
+    # tests because doing so would mean invoking the process as a script;
+    # there is no branching logic here for a test to miss.
+    "*/__main__.py",
+]
+
+[tool.coverage.report]
+show_missing = true
+skip_covered = false
+
+[tool.vibesys.per_module_coverage]
+# Per-module coverage floor (percent, statement + branch combined by
+# `coverage`). `--cov-fail-under` below is a repo-wide average and cannot
+# see a single near-zero module hiding inside a healthy total (issue #293).
+# `scripts/check_coverage_floor.py` enforces this floor per file using the
+# `coverage json` report.
+#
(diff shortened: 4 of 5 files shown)


vic-lsh:

Position 2 — independent of the other three (coverage config + CI step, 5 files).

PR #302, opened 2026-08-08
Adopt a stricter ruff lint configuration with a self-tightening suppression baseline

### LINKED ISSUE #295: Enable the ruff rule families the best-practices doc mandates
## Workstream

CLI/DX

## Problem

`docs/coding-best-practices.md` mandates several properties in prose that no check enforces, because `pyproject.toml` enables only `select = ["E", "F", "I", "UP", "B", "W"]`:

| Documented rule | Rule that would enforce it | Enabled? |
| --- | --- | --- |
| "Keep functions small enough that ownership is obvious" | `C901`, `PLR0912`, `PLR0915` | No |
| "Avoid implicit fallback behavior for agent-visible contracts" | `BLE001` | No |
| "Keep cross-boundary values typed and explicit" | `ANN401` | No |
| "Do not reach through an interface to depend on implementation details" | `SLF001` | No |
| Put repeated subprocess calls behind a focused interface | `S603`, `S607` | No |

Enabling just the first three families surfaces **198 violations**:

```
$ uv run ruff check --select ANN,BLE,C901,PLR0912,PLR0915 src libs/*/src
58 ANN401   49 BLE001   26 C901   19 ANN204
18 PLR0912  11 ANN201   10 PLR0915  7 ANN202
Found 198 errors.
```

Pyright cannot cover these. It reports `0 errors` on the same tree, and structurally cannot flag most of them — explicit `Any` is a deliberate escape hatch it permits, and complexity and blind-except are not type questions.

There is direct evidence the codebase already expects some of these rules: it carries `# noqa: ARG002`, `# noqa: PLC0415`, and `# noqa: BLE001` for rules that are not enabled. Those are dead suppressions, so adoption cost is partly pre-paid.

Separately, `ruff>=0.6` is a floor rather than a pin (currently resolving to 0.15.12), and CI runs `uv sync --dev` without `--frozen`. Ruff adds stable rules to existing families in minor releases, so a routine lock refresh can silently change what CI enforces.

## Desired outcome

The properties the best-practices document mandates are enforced by CI rather than by reviewer memory, and the enforced rule set is reproducible across time and machines.

## Acceptance criteria

- [ ] The rule families the best-practices doc mandates are enabled, with thresholds chosen from the measured distribution and recorded.
- [ ] Remaining violations are suppressed **per site** with a specific rule code and a tracking reference, not via global `ignore` entries or broad `per-file-ignores`.
- [ ] `RUF100` (unused-noqa) and `PGH003`/`PGH004` (blanket suppressions) are enabled, so suppressions must be specific and become CI failures once obsolete — the baseline can only shrink.
- [ ] Dead suppressions for unenabled rules are resolved.
- [ ] Rule families deliberately not enabled are documented with a one-line reason each.
- [ ] Ruff is pinned to an exact version, bumped deliberately in its own change.
- [ ] Vendored and example code the repo does not own is scoped out explicitly rather than accumulating suppressions.

## Scope and non-goals

In scope: ruff `select`, `ignore`, thresholds, per-file scoping, the suppression baseline, and the version pin.

Not in scope:

- Substantive refactors the new complexity rules will flag. `run_agent_loop` is #290; this issue baselines it.
- Pyright scope, formatter scope, and coverage — sibling issues under the same parent.
- Biome and TypeScript rules for `clients/`.

- Adopt via a per-site ratchet, not a `per-file-ignores` baseline. The `RUF100` + `PGH` combination is what makes the baseline self-tightening; a per-file block gives none of that.
- Suppression markers must not read `TODO`/`FIXME` if `FIX`/`TD` are enabled, or the baseline flags itself.
- Do not enable `COM` or `Q` — ruff documents them as conflicting with `ruff format`.
- Do not enable preview rules or set `preview = true`; preview semantics shift between releases, which moves the CI bar without a commit.
- `FA` should stay off despite firing zero times: it pushes toward `from __future__ import annotations`, which stringizes annotations and changes Pydantic's runtime resolution.
- `TC` is safe only with `runtime-evaluated-base-classes = ["pydantic.BaseModel"]`.
- `scripts/check_lint.sh` runs `ruff check .` across the whole repo, so scoping for `examples/` and `resources/` must be settled before adopting.

## Verification approach

- `./scripts/check_format.sh`, `./scripts/check_lint.sh`, `./scripts/check_types.sh`, `uv run pytest`, with a pre-change pytest baseline to prove no regression.
- Confirm the ratchet works: remove a suppression whose underlying issue is fixed and verify `RUF100` fails; add a bare `# noqa` and verify `PGH004` rejects it.
- Confirm autofixes are safe-only; `--unsafe-fixes` should not be applied mechanically.

Parent: #293. Roadmap: #288. Related: #290 (complexity violations this will baseline).

vic-lsh:
Closes #295
Part of #293
Part of #288

## Problem

`[tool.ruff.lint]` only selects `E, F, I, UP, B, W`. That misses whole classes
of bugs and hazards ruff can catch cheaply: bandit-style security checks (`S`),
blanket-except and error-handling patterns (`BLE`, `TRY`), missing type
annotations (`ANN`), complexity (`C901`, `PL`), and pathlib/datetime/logging
footguns (`PTH`, `DTZ`, `LOG`, `G`). `docs/coding-best-practices.md` already
asks for typed contracts, explicit failure modes, and small well-scoped
functions; ruff can enforce a meaningful slice of that automatically instead of
relying on review alone.

`ruff>=0.6` is also an open-ended floating bound: a routine `uv lock` refresh
could pick up a newer ruff with new default-on rules or changed autofix
behavior and silently change what CI enforces.

## Solution

Three commits: config, safe autofixes, then a per-site suppression baseline.

### Config (commit 1)

Expanded `select` to the families below, added `mccabe.max-complexity = 10`,
`pydocstyle.convention = "google"`, and a `flake8-type-checking`
`runtime-evaluated-base-classes = ["pydantic.BaseModel"]` entry so `TC` does
not try to move Pydantic field types into `TYPE_CHECKING` blocks, which would
break them at runtime. Pinned `ruff==0.15.12` (the version already resolved in
`uv.lock`) and refreshed the lock.

**Newly enabled:** `SIM, RUF, C4, PIE, ISC, RSE, INT, FLY, PERF, FURB, RET, A,
DTZ, LOG, G, N, PTH, YTT, EXE, ASYNC, ERA, T10, T20, TID, ICN, INP, FBT, FIX,
TD, PGH, ANN, BLE, SLF, C901, PL, S, PT, ARG, TRY, TC, D`

**Deliberately not enabled:**

| Family | Reason |
|---|---|
| `COM`, `Q` | Formatter conflict: both fight `ruff format`'s own comma and quote handling. |
| `preview = true` | Preview instability: preview rules churn across releases, which would defeat the version pin. |
| `FA` | Pydantic runtime-annotation hazard, and `from __future__ import annotations` is moot at py312. |
| `EM`, `CPY`, `DOC` | High noise-to-signal here (raw exception message strings, copyright headers, docstring content duplicating `D`); worth revisiting deliberately, not adopting sight-unseen. |
| `DJ`, `AIR`, `PD`, `FAST`, `NPY`, `PYI` | Inapplicable frameworks: no Django, Airflow, pandas, FastAPI, NumPy, or `.pyi` stubs in this repo. |

### Scoping (commit 1)

`scripts/check_lint.sh` runs `ruff check .`, i.e. the whole repo, so the new
rules had to be scoped to code this repo actually owns:

- **`resources/skills/` → `extend-exclude`.** Vendored reference material:
  submodule checkouts of vllm/sglang/TensorRT-LLM plus copied reference docs.
  The existing `UP040`/`UP046`/`UP047` comment already documents this subtree
  as not owned by VibeSys.
- **`examples/` → `per-file-ignores` for the new families only.** It ships
  candidate reference implementations, benchmark harnesses, and vendored
  submodules (transformers, DeathStarBench, TensorRT-LLM, Show-o, train-ticket,
  TraceLab) rather than first-party framework code. The pre-existing
  `E, F, I, UP, B, W` baseline, which it already passes, stays enforced; only
  the newly added families are skipped.
- **`resources/profilers/` gets the full rule set, same as `src/`.** An earlier
  revision of this PR carved it out; that did not hold up on review.
  `docs/extending-profilers.md` documents
  `resources/profilers/<kind>/server.py` as the extension point contributors
  are expected to write, so it is first-party maintained code, not vendored or
  reference material. It also launches subprocesses inside sandboxes, which
  makes `BLE`, `S`, and `ANN` *more* valuable there than average, not less. It
  is only 15 Python files.

`tests/**` and `libs/*/tests/**` get the four standard pytest per-file-ignores:
`S101` (assert), `PLR2004` (magic values), `D` (docstrings), `INP001` (implicit
namespace package).

### Safe autofixes (commit 2)

`ruff check --fix .` (safe fixes only, no `--unsafe-fixes`) plus
`ruff format .`. 108 fixes, all mechanical: `SIM117` nested-`with` collapsing,
`RUF022` `__all__` sorting, `PERF102` redundant dict iterator, import sorting,
and formatter whitespace.

### Suppression baseline (commit 3)

6,412 violations remained. Every one got a per-site
`# noqa: <CODE>  # tracked: #288` comment. No rule went into the global
`ignore` list, and no `per-file-ignores` were added beyond commit 1's.
Generated with `ruff check --add-noqa`, iterated against
`ruff check --fix` / `ruff format` to a fixed point.

**Ratchet mechanism:** `RUF100` (unused-noqa) and `PGH003`/`PGH004` (blanket
`type: ignore` / bare `noqa`) are enabled. When someone fixes an underlying
violation, its now-unused suppression becomes a `RUF100` CI failure and must be
deleted. The baseline can only shrink, never silently persist past the code that
justified it. The `# tracked: #288` marker is used instead of `TODO`/`FIXME`
precisely because `FIX002`/`TD002`/`TD003` are now enabled and would flag the
baseline itself.

**One site shape differs, deliberately.** In `analyze_nsys.py`, nine section
dividers of the form `# Subcommand: export` are false-positive `ERA001` hits
(`Subcommand: export` parses as an annotated assignment). Appending
`# noqa: ERA001  # tracked: #288` to such a line changes the comment text
enough that `ERA001` stops matching, which makes the directive unused and trips
`RUF100` instead, so inline suppression is self-defeating. Those nine carry a
bare `# noqa: ERA001` with the `# tracked: #288` marker on the preceding line.
This deliberately avoids a per-file-ignore, which would *not* ratchet; `RUF100`
still guards these nine sites.

**Commit 3 is mechanical suppression with no behavior change.** No types were
weakened (no `Any` added to silence `ANN`), no code deleted, no logic
restructured.

## Verification

Baseline recorded against `origin/main` before any change, in a separate
worktree: **2948 passed, 1 skipped**.

### Correctness properties

- **AST equivalence.** Commit 3 was verified by comparing
  `ast.dump(ast.parse(...))` for all **338** tracked Python files against the
  pre-suppression commit: **0 mismatches**. This is what proves the added
  comments (and the line reflow they trigger, plus the hand-cleaned stray
  fragments) changed no logic.
- `uv run pytest` matches `origin/main` exactly: no regressions, no new skips.
- Suppressions live only in per-site `# noqa` comments; the global `ignore` list
  still contains exactly the four pre-existing documented entries.
- `RUF100`/`PGH003`/`PGH004` enabled, so stale suppressions fail CI.

### Testing

```
./scripts/check_format.sh   # All checks passed! (280 files already formatted)
./scripts/check_lint.sh     # All checks passed!
./scripts/check_types.sh    # 0 errors, 0 warnings, 0 informations
uv run pytest               # 2948 passed, 1 skipped  (matches origin/main)
```

Violations after safe autofixes, before suppression:

| Tree | Violations |
|---|---|
| `src` | 1,251 |
| `tests` | 3,998 |
| `libs` (4 packages, src + tests) | 835 |
| `resources/profilers` | 328 |
| `examples`, `resources/skills` | 0 (scoped out of new rules; prior baseline still enforced) |
| **Total** | **6,412** |

`ruff check .` reports zero violations after suppression.

### DIFF
--- a/libs/vs-feature-flags/src/vs_feature_flags/__init__.py
+++ b/libs/vs-feature-flags/src/vs_feature_flags/__init__.py
@@ -1,4 +1,4 @@
-from vs_feature_flags.config import parse_feature_flag_overrides
+from vs_feature_flags.config import parse_feature_flag_overrides  # noqa: D104  # tracked: #288
 from vs_feature_flags.core import FeatureDefinition, FeatureRegistry
 
 __all__ = [
--- a/libs/vs-feature-flags/src/vs_feature_flags/config.py
+++ b/libs/vs-feature-flags/src/vs_feature_flags/config.py
@@ -1,4 +1,4 @@
-from __future__ import annotations
+from __future__ import annotations  # noqa: D100  # tracked: #288
 
 from collections.abc import Mapping
 from enum import StrEnum
@@ -22,7 +22,7 @@ def parse_feature_flag_overrides(
         return {}
 
     if not isinstance(raw, Mapping):
-        raise ValueError(f"{section_name} must be a TOML table")
+        raise ValueError(f"{section_name} must be a TOML table")  # noqa: TRY003, TRY004  # tracked: #288
 
     parsed: dict[FlagT, bool] = {}
     for key, value in raw.items():
@@ -30,10 +30,10 @@ def parse_feature_flag_overrides(
             flag = flag_type(str(key))
         except (TypeError, ValueError):
             valid = _format_valid_flags(flag_type)
-            raise ValueError(f"Unknown feature flag {key!r}. Supported flags: {valid}") from None
+            raise ValueError(f"Unknown feature flag {key!r}. Supported flags: {valid}") from None  # noqa: TRY003  # tracked: #288
 
         if not isinstance(value, bool):
-            raise ValueError(f"{section_name}.{key} must be true or false")
+            raise ValueError(f"{section_name}.{key} must be true or false")  # noqa: TRY003, TRY004  # tracked: #288
 
         parsed[flag] = value
 
--- a/libs/vs-feature-flags/src/vs_feature_flags/config.py
+++ b/libs/vs-feature-flags/src/vs_feature_flags/config.py
@@ -30,9 +30,7 @@ def parse_feature_flag_overrides(
             flag = flag_type(str(key))
         except (TypeError, ValueError):
             valid = _format_valid_flags(flag_type)
-            raise ValueError(
-                f"Unknown feature flag {key!r}. Supported flags: {valid}"
-            ) from None
+            raise ValueError(f"Unknown feature flag {key!r}. Supported flags: {valid}") from None
 
         if not isinstance(value, bool):
             raise ValueError(f"{section_name}.{key} must be true or false")
--- a/libs/vs-feature-flags/src/vs_feature_flags/core.py
+++ b/libs/vs-feature-flags/src/vs_feature_flags/core.py
@@ -1,6 +1,6 @@
-from __future__ import annotations
+from __future__ import annotations  # noqa: D100  # tracked: #288
 
-from collections.abc import Mapping
+from collections.abc import Mapping  # noqa: TC003  # tracked: #288
 from dataclasses import dataclass
 from enum import StrEnum
 from types import MappingProxyType
@@ -10,7 +10,7 @@ FlagT = TypeVar("FlagT", bound=StrEnum)
 
 
 @dataclass(frozen=True)
-class FeatureDefinition:
+class FeatureDefinition:  # noqa: D101  # tracked: #288
     description: str
     default: bool = False
 
@@ -18,7 +18,7 @@ class FeatureDefinition:
 class FeatureRegistry(Generic[FlagT]):
     """Registry for a project's typed feature flag manifest."""
 
-    def __init__(
+    def __init__(  # noqa: D107  # tracked: #288
         self,
         flag_type: type[FlagT],
         definitions: Mapping[FlagT, FeatureDefinition],
@@ -28,17 +28,17 @@ class FeatureRegistry(Generic[FlagT]):
         self._validate_definitions()
 
     @property
-    def flag_type(self) -> type[FlagT]:
+    def flag_type(self) -> type[FlagT]:  # noqa: D102  # tracked: #288
         return self._flag_type
 
     @property
-    def definitions(self) -> Mapping[FlagT, FeatureDefinition]:
+    def definitions(self) -> Mapping[FlagT, FeatureDefinition]:  # noqa: D102  # tracked: #288
         return MappingProxyType(self._definitions)
 
-    def default_for(self, flag: FlagT) -> bool:
+    def default_for(self, flag: FlagT) -> bool:  # noqa: D102  # tracked: #288
         return self._definitions[flag].default
 
-    def is_enabled(
+    def is_enabled(  # noqa: D102  # tracked: #288
         self,
         flag: FlagT,
         overrides: Mapping[FlagT, bool] | None = None,
@@ -50,10 +50,10 @@ class FeatureRegistry(Generic[FlagT]):
     def _validate_definitions(self) -> None:
         for flag in self._definitions:
             if not isinstance(flag, self._flag_type):
-                raise TypeError(
+                raise TypeError(  # noqa: TRY003  # tracked: #288
                     f"Feature definition key {flag!r} is not a {self._flag_type.__name__}"
                 )
 
         missing = [flag.value for flag in self._flag_type if flag not in self._definitions]
         if missing:
-            raise ValueError(f"Missing feature definitions for: {', '.join(missing)}")
+            raise ValueError(f"Missing feature definitions for: {', '.join(missing)}")  # noqa: TRY003  # tracked: #288
--- a/libs/vs-feature-flags/tests/test_core.py
+++ b/libs/vs-feature-flags/tests/test_core.py
@@ -30,21 +30,21 @@ def _registry() -> FeatureRegistry[ExampleFlag]:
     )
 
 
-def test_registry_uses_defaults():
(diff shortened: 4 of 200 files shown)


vic-lsh:

Position 3 — large mechanical baseline (200 files). Land before #303.

vic-lsh:
Superseded by #307, which combines this PR with the other of the pair.

Rationale: the two had **45/45 file overlap** (every file the pyright PR touched, the ruff PR also touched) and produced **18 conflicted files / 33 hunks in either merge order** — reordering was symmetric and did not help. Every conflict was a single line's comment tail, where one PR appended a `# noqa` while the other deleted a now-stale `# pyright: ignore`. Because `RUF100` and `reportUnnecessaryTypeIgnoreComment` each validate the other tool's comments, and `ruff format` reflows appended comments, the correct end state only converges when both baselines are generated together on one tree.

#307 does that. Verified independently: both suppression commits are AST-identical to their parents (195/195 and 40/40 files), all five ratchet directions fire, test collection matches `main` exactly at 2950, and the trial merge `#299 → #301 → #307` now conflicts on **3 hunks instead of 36** — only the known `libs/` formatting overlap with #299, whose resolution is documented in #307.

No work is lost: #307 contains this branch's commits.

PR #303, opened 2026-08-08
Bring tests/ under pyright with a ratcheted baseline

### LINKED ISSUE #297: Bring tests/ under pyright
## Workstream

CLI/DX

## Problem

Pyright runs in `strict` mode, but `include = ["src", "libs/vs-*/src"]` in `pyproject.toml` omits `tests/` and `libs/*/tests`. That leaves roughly 28.6k lines — about the same volume as `src/` — with no type checking at all, in a repository that otherwise holds first-party code to a strict bar.

The gap matters more than a test-only gap normally would, because test code here builds typed fixtures (`MagicMock(spec=AgentRunner)`, real `MutatorResponse`/`JudgeResponse` objects, manifest and config models) that encode assumptions about production contracts. When a contract changes, untyped test code silently keeps compiling against the old shape.

The in-repo precedent for closing this is on the TypeScript side: `clients/tui/tsconfig.check.json` overrides `exclude: []` specifically so test files are typechecked. Python has no equivalent.

## Desired outcome

Test code is type checked, so contract drift between production models and the fixtures that exercise them is caught at check time. The strict bar for `src/` and `libs/*/src` is unchanged.

## Acceptance criteria

- [ ] `tests/` and `libs/*/tests` are included in pyright's checked set.
- [ ] The strict configuration applying to `src/` and `libs/*/src` is unchanged, including the three existing documented relaxations.
- [ ] Any relaxation for test paths is scoped to those paths, documented with reasoning in `pyproject.toml`, and does not use blanket suppressions.
- [ ] `./scripts/check_types.sh` passes, and CI enforces the expanded scope.

## Scope and non-goals

In scope: pyright `include`, `executionEnvironments`, and any test-scoped diagnostic settings; fixing or scoping the errors the expansion surfaces.

Not in scope:

- Weakening any setting that applies to first-party source.
- Ruff rules, formatter scope, and coverage — sibling issues.
- Rewriting tests to improve their design. Assertions on private members are a real problem but belong to the test-effectiveness work, not here.

- Measure the error count before choosing an approach; the right strategy differs by an order of magnitude between "a few dozen" and "several thousand."
- Prefer a scoped `executionEnvironments` entry with a relaxed `typeCheckingMode` for test paths over disabling individual rules globally. Confirm the pinned pyright version actually honors per-environment `typeCheckingMode` rather than assuming it.
- Do not silence errors by adding `Any` annotations or blanket `# type: ignore`. If ruff's `PGH003` lands first, blanket ignores will be rejected outright.
- A relaxed-but-enabled mode for tests is more valuable than perfect strictness, because the goal is catching contract drift rather than achieving a grade.

## Verification approach

- Record the pyright error count before and after, with a breakdown by rule.
- `./scripts/check_types.sh`, `./scripts/check_format.sh`, `./scripts/check_lint.sh`, and `uv run pytest` all pass.
- Demonstrate the new scope catches something real: change a production model field locally and confirm pyright now reports the stale test fixture.

Parent: #293. Roadmap: #288.

vic-lsh:
## Problem

Pyright runs in `strict` mode, but `[tool.pyright].include` only covered
`src` and `libs/*/src`. `tests/` and `libs/*/tests` (~28.6k lines, roughly
the same size as `src/`) were never type-checked at all — half the Python
in the repo was invisible to the strict checker.

That gap matters more than a test-only gap normally would, because test
code here builds typed fixtures (`MagicMock(spec=...)`, real
`MutatorResponse`/`JudgeResponse` objects, manifest and config models)
that encode assumptions about production contracts. When a contract
changes, untyped test code silently keeps compiling against the old
shape. The in-repo precedent for closing this is on the TypeScript side:
`clients/tui/tsconfig.check.json` sets `exclude: []` specifically so test
files are type-checked.

Closes #297. Part of #293 and #288.

## Solution

Added `tests` and `libs/*/tests` to `[tool.pyright].include`. Measuring
first: running the full tree at strict surfaced **4,445 errors**. ~77%
(3.4k) came from two rules alone (`reportMissingParameterType` /
`reportUnknownParameterType`), fired almost entirely on untyped pytest
fixture parameters (`tmp_path`, `monkeypatch`, custom fixtures).

#297 prefers a scoped `executionEnvironments` entry with a relaxed
`typeCheckingMode` over disabling rules. I verified whether pyright
1.1.411 honors `typeCheckingMode` set *inside* an `executionEnvironments`
entry: **it does not**. Setting it to `"standard"` and separately to
`"off"` on a `tests`-scoped environment left the error count unchanged
(4446 both times), versus a single rule-level override alone taking it to
1043. It is a top-level-only setting in this version, silently accepted
as a per-environment key. So the relaxation is spelled out explicitly
instead, in two tiers — both scoped to test paths only. `src` and
`libs/*/src` keep their existing environments and the unmodified strict
bar; the three pre-existing global relaxations
(`reportUnknownMemberType`/`Variable`/`ArgumentType`) and their
justifying comment are preserved verbatim.

**Tier 1 — rule-level, for rules categorically inapplicable to test
code.** Untyped fixture params and lambdas
(`reportMissingParameterType`, `reportUnknownParameterType`,
`reportUnknownLambdaType`), bare generics in ad hoc test data
(`reportMissingTypeArgument`), deliberate private-state/import access for
asserting on internals and monkeypatching (`reportPrivateUsage`,
`reportPrivateImportUsage`), and `unittest.mock`'s dynamic attribute
surface (`reportAttributeAccessIssue`, `reportFunctionMemberAccess`).
These have no annotation to check or are the documented way to write the
test, so a rule-level relaxation is the honest expression of that.

**Tier 2 — per-site suppressions with a ratchet, for everything else.**
The remaining 358 findings across 8 rules (`reportArgumentType`,
`reportOptionalMemberAccess`, `reportOperatorIssue`, `reportReturnType`,
`reportCallIssue`, `reportOptionalOperand`, `reportIndexIssue`,
`reportAbstractUsage`) can be real bugs rather than missing
declarations, so they must not become invisible. Setting them to `"none"`
would also silently accept every *future* violation of those 8 rules in
test code — no ratchet at all. Instead they stay at the strict default
and are suppressed at **233 individual sites** with
`# pyright: ignore[<rule>]  # tracked: #297`, backed by a new top-level:

```toml
reportUnnecessaryTypeIgnoreComment = "error"
```

That makes the baseline monotonically shrinking: once an underlying error
is fixed, its now-dead suppression becomes an error itself and must be
removed. This mirrors the `# noqa: CODE` + `RUF100` approach used for the
ruff rule expansion under the same parent. `grep 'tracked: #297'` yields
the burn-down list. Set at the top level, so it covers `src` and
`libs/*/src` too, not just the test paths whose baseline depends on it.

Every suppression names its specific rule; no bare `# type: ignore` was
added anywhere (relevant because a concurrent PR enables ruff's
`PGH003`).

**Dead suppressions removed (10).** Enabling
`reportUnnecessaryTypeIgnoreComment` flagged 10 pre-existing suppressions
as no longer doing anything, all removed:

- **2 in first-party source**, dead independently of this change (verified
  by running the ratchet against the original `main` include scope):
  `libs/vs-sandbox/src/vs_sandbox/modal_sandbox.py` (dropped a stale
  `reportOptionalMemberAccess` from a two-rule ignore, kept the live
  `reportAssignmentType`) and `src/vibesys/loops/plain/loop.py` (a
  `# type: ignore[no-redef]` that suppressed nothing).
- **8 in `tests/`**, dead because those paths are newly checked and tier 1
  covers their rules: 7 × `# type: ignore[method-assign]` in
  `tests/agents/test_omnigent_backend.py`, 1 ×
  `# pyright: ignore[reportAttributeAccessIssue]` in
  `tests/backends/test_cuda_backend.py`.

**Genuine fixes rather than suppressions.** Small, single-instance issues
the new coverage surfaced were fixed properly:

- `tests/test_tui.py` — a genuinely possibly-unbound `events` variable.
- `tests/agents/test_agent_runners.py` — an always-true comparison from a
  too-narrow inferred literal type; annotated `str | None`.
- `tests/loops/plain/test_plain_loop_state.py` — two unused unpacked
  variables, renamed `_i`.
- `tests/_agent_cli/fixtures/agents.py` — a test double's `generate()`
  override was incompatible with `CodingAgent.generate`; widened the
  parameter types to match.
- `tests/loops/test_profiler.py` — 5 imports of a `resources/`
  runtime-only module that pytest's `pythonpath = ["."]` resolves at
  runtime but pyright cannot once `tests` has its own execution
  environment. I tried `extraPaths = ["."]` first; that made pyright
  treat the first-party `vibesys` package as an external stub-only
  dependency (`reportMissingTypeStubs`), so I reverted it and suppressed
  the 5 lines by name with an explanatory module docstring note.
- `tests/test_context.py` — one `autouse=True` fixture pyright cannot see
  is used, suppressed by name.

### Architecture

Config plus mechanical test annotations. No production code, runtime
behavior, or contract changed; the only non-test source edits are the two
dead-suppression removals. `[tool.pyright]` is the only `pyproject.toml`
section touched, per the coordination note that concurrent PRs are
modifying `[tool.ruff.lint]` and the coverage config in the same file.

### Correctness properties

- `src` and `libs/*/src` remain at unmodified strict — no rule relaxed or
  added for those trees; the existing three-rule relaxation is
  byte-identical. Verified by appending a deliberate
  undeclared-attribute access to `src/vibesys/errors.py` and confirming
  pyright still flagged it, then reverting.
- `tests` and `libs/*/tests` are now type-checked at all (previously: not
  checked), at 0 errors.
- **The tier-2 baseline can only shrink.** Verified end-to-end in three
  directions, each producing exactly one error and then reverted:
  1. Removed a live suppression → the real error resurfaced
     (`reportOptionalMemberAccess`).
  2. Added a suppression on a clean line → rejected
     (`Unnecessary "# pyright: ignore" rule`).
  3. Properly fixed a suppressed site (narrowed the `Optional`) while
     leaving its ignore in place → the now-dead ignore was flagged.
  Direction 3 is the one that matters: fixing a finding forces removal of
  its suppression, so the count cannot silently stay flat.
- No blanket suppressions: every ignore added names its rule(s). No
  `# type: ignore` was added by this PR.

Residual, noted rather than changed: 12 pre-existing `# type: ignore[code]`
comments remain in `src` and `tests`. Pyright ignores the bracketed codes
and treats each as an all-rules suppression for its line, so they are
coarser than the new named ignores — but they are validated as still
necessary by the new setting, and ruff's `PGH003` accepts them since they
carry codes. Converting them to named `# pyright: ignore[...]` is a
mechanical follow-up, deliberately left out to keep this diff scoped.

### Testing

```
./scripts/check_format.sh   # All checks passed! (310 files already formatted)
./scripts/check_lint.sh     # All checks passed!
./scripts/check_types.sh    # 0 errors, 0 warnings, 0 informations
uv run pytest               # 2948 passed, 1 skipped
```

Error count: **4,445 → 0**, with 233 tracked per-site suppressions and 10
dead suppressions removed.

One process note for reviewers: `ruff format` reflowed 4 of the
mechanically-inserted comments off their error lines. The new
`reportUnnecessaryTypeIgnoreComment` setting caught all 4 immediately
(each showed up as a paired real-error + dead-ignore), and they were
repositioned by hand. That is the ratchet doing its job on its own
rollout.

### DIFF
--- a/libs/vs-feature-flags/tests/test_core.py
+++ b/libs/vs-feature-flags/tests/test_core.py
@@ -70,4 +70,4 @@ def test_definitions_mapping_is_read_only():
     registry = _registry()
 
     with pytest.raises(TypeError):
-        registry.definitions[ExampleFlag.NEW_LOOP] = FeatureDefinition(description="Changed.")
+        registry.definitions[ExampleFlag.NEW_LOOP] = FeatureDefinition(description="Changed.")  # pyright: ignore[reportIndexIssue]  # tracked: #297
--- a/libs/vs-issue-board/tests/test_issue_board_core.py
+++ b/libs/vs-issue-board/tests/test_issue_board_core.py
@@ -322,10 +322,10 @@ class TestStateTransitions:
             issue.id, IssueStatus.IN_PROGRESS, actor="loop", iteration=1, note="claimed"
         )
         reloaded = store.get(issue.id)
-        assert reloaded.status == IssueStatus.IN_PROGRESS
-        assert len(reloaded.history) == 2
-        assert reloaded.history[-1].action == "open->in_progress"
-        assert reloaded.history[-1].note == "claimed"
+        assert reloaded.status == IssueStatus.IN_PROGRESS  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert len(reloaded.history) == 2  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert reloaded.history[-1].action == "open->in_progress"  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert reloaded.history[-1].note == "claimed"  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
 
     def test_update_status_to_closed_records_closed_iter(self, tmp_path):
         store = _make_store(tmp_path)
@@ -333,8 +333,8 @@ class TestStateTransitions:
         store.update_status(issue.id, IssueStatus.IN_PROGRESS, actor="loop", iteration=2)
         store.update_status(issue.id, IssueStatus.CLOSED, actor="judge", iteration=2, note="passed")
         reloaded = store.get(issue.id)
-        assert reloaded.status == IssueStatus.CLOSED
-        assert reloaded.closed_iter == 2
+        assert reloaded.status == IssueStatus.CLOSED  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert reloaded.closed_iter == 2  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
 
     def test_update_status_blocked_records_closed_iter(self, tmp_path):
         store = _make_store(tmp_path)
@@ -343,8 +343,8 @@ class TestStateTransitions:
             issue.id, IssueStatus.BLOCKED, actor="loop", iteration=3, note="exhausted retries"
         )
         reloaded = store.get(issue.id)
-        assert reloaded.status == IssueStatus.BLOCKED
-        assert reloaded.closed_iter == 3
+        assert reloaded.status == IssueStatus.BLOCKED  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert reloaded.closed_iter == 3  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
 
     def test_increment_attempts_appends_event(self, tmp_path):
         store = _make_store(tmp_path)
@@ -353,8 +353,8 @@ class TestStateTransitions:
         store.increment_attempts(issue.id, actor="implementer", iteration=1, note="first try")
         store.increment_attempts(issue.id, actor="implementer", iteration=1, note="retry")
         reloaded = store.get(issue.id)
-        assert reloaded.attempts == 2
-        assert sum(1 for e in reloaded.history if e.action == "attempt") == 2
+        assert reloaded.attempts == 2  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
+        assert sum(1 for e in reloaded.history if e.action == "attempt") == 2  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
 
     def test_update_unknown_id_raises(self, tmp_path):
         store = _make_store(tmp_path)
@@ -412,13 +412,13 @@ class TestNextOpen:
         feature = _create(store, type=IssueType.FEATURE, title="feature")
         bug = _create(store, type=IssueType.BUG, title="bug")
         first = store.next_open()
-        assert first.id == bug.id
+        assert first.id == bug.id  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
         store.update_status(bug.id, IssueStatus.CLOSED, actor="x", iteration=1)
         second = store.next_open()
-        assert second.id == feature.id
+        assert second.id == feature.id  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
         store.update_status(feature.id, IssueStatus.CLOSED, actor="x", iteration=1)
         third = store.next_open()
-        assert third.id == perf.id
+        assert third.id == perf.id  # pyright: ignore[reportOptionalMemberAccess]  # tracked: #297
 
     def test_next_open_fifo_within_type(self, tmp_path):
         store = _make_store(tmp_path)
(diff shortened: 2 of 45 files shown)


vic-lsh:

Position 4 — land last. Its suppression comments are written to satisfy #302's `PGH003`, and `ruff format` from #302 can reflow inserted comments off their target lines.

vic-lsh:
Superseded by #307, which combines this PR with the other of the pair.

Rationale: the two had **45/45 file overlap** (every file the pyright PR touched, the ruff PR also touched) and produced **18 conflicted files / 33 hunks in either merge order** — reordering was symmetric and did not help. Every conflict was a single line's comment tail, where one PR appended a `# noqa` while the other deleted a now-stale `# pyright: ignore`. Because `RUF100` and `reportUnnecessaryTypeIgnoreComment` each validate the other tool's comments, and `ruff format` reflows appended comments, the correct end state only converges when both baselines are generated together on one tree.

#307 does that. Verified independently: both suppression commits are AST-identical to their parents (195/195 and 40/40 files), all five ratchet directions fire, test collection matches `main` exactly at 2950, and the trial merge `#299 → #301 → #307` now conflicts on **3 hunks instead of 36** — only the known `libs/` formatting overlap with #299, whose resolution is documented in #307.

No work is lost: #307 contains this branch's commits.

PR #311, opened 2026-08-08
Extract _resolve_round_outcome from run_agent_loop (#290)

### LINKED ISSUE #290: Decompose run_agent_loop into round-scoped units
## Workstream

Core/runtime

## Problem

`run_agent_loop` (`src/vibesys/loops/agent/loop.py:2242`) has grown into the single largest change risk in the framework. Measured against the working tree:

- **1018 lines** in one function
- **cyclomatic complexity 63**, **77 branches**, **316 statements**
- **maximum nesting depth 11** (44 spaces of indentation)
- 589 of its lines sit at indent level 24 or deeper

It owns, in one body: round sequencing, profiling invocation, judge dispatch, gate execution, rollback and parent-commit resolution, hypothesis-outcome labelling, state persistence, and supervision event emission.

Two consequences are already visible in the repository rather than hypothetical:

1. The rollback-parent logic buried in this function depends on a hand-maintained string set (`loop.py:359`) that is out of sync with its own enum, producing a wrong-commit rollback. Logic at depth 11 does not get reviewed as carefully as logic at depth 2.
2. `docs/coding-best-practices.md` states "Keep functions small enough that ownership is obvious." Nothing enforces this, and `C901`/`PLR0912`/`PLR0915` are not enabled in `pyproject.toml:79`, so the function grew without any check objecting.

`src/vibesys/loops/agent/loop.py` is 3279 lines overall, and `run_agent_loop` is roughly a third of it.

## Desired outcome

Round execution is expressed as a small number of named, independently readable units with obvious ownership, so that a reader can locate the sequencing, gating, rollback, and persistence responsibilities without following an 11-level-deep control flow. Observable behavior of the agent loop is unchanged.

## Acceptance criteria

- [ ] `run_agent_loop` no longer exceeds the `C901`, `PLR0912`, and `PLR0915` thresholds that this repository adopts.
- [ ] Round sequencing, gate execution, rollback/parent-commit resolution, and state persistence are separately addressable units with explicit typed inputs and outputs.
- [ ] Maximum nesting depth within the extracted units is substantially reduced from 11.
- [ ] No change to emitted supervision events, persisted `rounds.json` contents, or loop-visible behavior.
- [ ] Existing loop tests pass unmodified, or any modification is justified as a contract change rather than an implementation-detail fix.

## Scope and non-goals

In scope:

- Decomposing `run_agent_loop` and directly supporting helpers within `src/vibesys/loops/agent/`.
- Introducing typed intermediate values where they clarify the boundaries between extracted units.

Not in scope:

- Changing loop semantics, round ordering, gate behavior, or event payloads.
- The evolve and plain loops (`loops/evolve/loop.py:1225` `run_evolve_loop` at 254 lines and `loops/plain/loop.py:344` `run_plain_loop` at 446 lines have the same shape but should be separate work).
- Fixing the rollback outcome-vocabulary defect, which is tracked separately and should land first so this refactor does not carry the bug forward.

- This should be staged across multiple reviewable pull requests, following the precedent of #170 decomposed into #171, #172, and #173. A single large rewrite is explicitly not wanted.
- The loop is the highest-churn area of the framework, so long-lived branches will conflict. Prefer a sequence of small extractions that each land independently.
- Per `docs/coding-best-practices.md`, do not introduce abstraction without a shared interface to protect; extraction for readability should not become a speculative plugin layer.

## Verification approach

- `uv run pytest tests/loops/agent/` first, then the full suite.
- `uv run ruff check --select C901,PLR0912,PLR0915 src/vibesys/loops/agent/loop.py` to confirm the complexity reduction is real and measurable.
- Prompt snapshot review (`tests/loops/agent/test_prompt_snapshots.py`) to confirm no rendered agent-visible change.
- Since `rounds.json` shape and supervision events are contracts, confirm they are byte-identical for an equivalent run rather than relying on unit tests alone.

Parent: #288. Precedent: #170, #171, #172, #173. Should land after the rollback outcome-vocabulary fix.

zatchbell1311-wq:
## Problem

`run_agent_loop`'s post-retry-loop bookkeeping resolved per-round metrics,
hypothesis outcome labels, and candidate evidence inline as a single ~160-line
block, mixing single-agent and multi-agent branches, gate-revalidation
fallbacks, and candidate-disposition inheritance all in one place inside an
already very large function. This made the round-result logic hard to read,
test in isolation, or reason about independently from the rest of the loop.

Closes #290.

## Solution

Extract the block into a dedicated `_RoundOutcome` dataclass and a pure
`_resolve_round_outcome()` function:

- `_RoundOutcome` holds the resolved values previously assigned to a dozen
  loose locals (`perf_metric`, `perf_unit`, `profile_skipped`,
  `accepted_metrics`, `hypothesis_outcome`, `candidate_disposition`, etc.).
- `_resolve_round_outcome()` takes the retry-loop's terminal state
  (`inner_loop`, `passed`, `implementation`, `single_agent_response`,
  `active_hypothesis`, etc.) and returns a single `_RoundOutcome`. It is a
  read-only function — it does not mutate `active_hypothesis`; the caller
  still owns any subsequent state transitions.
- `run_agent_loop` now calls `_resolve_round_outcome(...)` once after the
  retry loop exits and unpacks the result into the same local variable names
  used by the downstream `_RoundRecord(...)` construction, so the persisted
  round-record shape and behavior are unchanged.

No behavioral change is intended — this is a pure extraction. Existing
single-agent vs multi-agent branching, gate-revalidation candidate/perf-metric
fallback, and Pareto-frontier candidate inheritance are preserved exactly as
before.

## Testing

- `python -m py_compile src/vibesys/loops/agent/loop.py`
- `uv run ruff check src/vibesys/loops/agent/loop.py` — all checks passed
- `uv run pytest tests/loops/agent/ -x` — 258 passed (run under WSL/Linux,
  since `tests/conftest.py` imports `socketserver.ThreadingUnixStreamServer`,
  which isn't available on native Windows Python)

### DIFF
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -755,6 +755,29 @@ class _CarryOver:
     regression_info: str | None = None
     exhaustion_info: str | None = None
 
+@dataclass
+class _RoundOutcome:
+    """Resolved per-round metrics, outcome labels, and candidate evidence.
+
+    Computed once after the retry loop exits so that ``run_agent_loop``
+    can hand a single typed value to round-record persistence rather than
+    unpacking a dozen local variables inline.
+    """
+
+    perf_metric: float | None
+    perf_unit: str | None
+    profile_skipped: bool
+    accepted_metrics: dict[str, float]
+    accepted_evaluation_artifact: str | None
+    hypothesis_outcome: str | None
+    reviewed: bool
+    candidate_disposition: str
+    candidate_metrics: dict[str, float]
+    candidate_evaluation_artifact: str | None
+    candidate_operating_point: str
+    candidate_retention_reason: str
+    last_single_agent_response: SingleAgentRoundResponse | None
+
 
 def _review_due(
     *,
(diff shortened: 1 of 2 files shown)


vic-lsh:
Closing this one: #577 (merged) restructured the same retry-loop bookkeeping in `run_agent_loop` that this PR extracts, so the branch would need a full rework against the new shape, and it has been conflicting with main for several weeks. The extraction of a round-outcome helper is still a good direction; if you want to pick it up again, please rebase onto current main and open a fresh PR referencing #290. Thanks for the original work.

Issue #316, opened 2026-08-10
Priority Queue


rairai77:
### Work item type

Reference, benchmark, and checker harness

### Parent issue

#36

### Purpose and motivation

This represents any workload in which you need to prioritize tasks by an order other than FIFO. i.e. deadline, customer status, urgency, etc.

The goal is for the implementor to design a priority queue following a given C ABI that (hopefully) demonstrates a speedup when compared to a baseline implementation

### Required behavior or semantic contract

Implementations must follow a specific C ABI found at examples/evaluators/priority-queue/testdata/abi_test_candidate.c

### Correctness checks

Validator checks for linearizability right now but it may be worthwhile to consider relaxed linearizability for non-strict priority queues in future implementations

### Benchmark dimensions

Parameters:
		"--scenario" (MPMC, SPMC, etc.),
		"--capacity",
		"--value-size",
		"--producers" (number of producers, each is given its own pinned thread in the default implementation),
		"--consumers" (number of consumers, each is given its own pinned thread in the default implementation),
		"--warmup-ns",
		"--duration-ns",
		"--output"

Results:
	Scenario 
	Enqueued  
	Dropped
	Dequeued 
	Empty
	Attempts
	Duration
	TotalOpsPerSec
	Producers
	Consumers
	Repetitions
	TotalOpsPerSecSamples

### Non-goals

Anything not stated as a goal is not a goal.

### Research or implementation anchors

Starter implementation included in starters

### Contract check

- [x] The issue defines required behavior without prescribing the implementation.

PR #317, opened 2026-08-10
Add a concurrent bounded priority queue evaluator, ABI, and four scenario inputs

rairai77:
## Problem

Related to #316

VibeSys has queue optimization inputs, but nothing that forces a candidate to
maintain a *priority* order under concurrency. Concurrent bounded queues and
concurrent bounded priority queues are different optimization problems: the
priority variant makes the fast paths (ordering, tie-breaking, capacity
accounting) genuinely contended, so it is a stronger test of whether the
optimizer can trade structure for throughput without silently breaking the
data-structure contract.

There was also no shared harness able to score such a candidate safely. Loading
untrusted native code into the process that owns the linearizability verdict
would let a candidate influence its own scoring.

This adds the priority queue evaluator, four scenario inputs (SPSC, MPSC, SPMC,
MPMC), and a naive Rust starter, following the layout already established by
the existing `queue` evaluator.

## Solution

Three pieces, mirroring the existing queue example:

- `examples/evaluators/priority-queue` — trusted Go evaluator: workload
  generation, history recording, Porcupine model checking, and the
  `check` / `benchmark` CLI.
- `examples/evaluators/priority-queue/native_runner` — trusted Rust process
  that `dlopen`s the candidate and makes every ABI call. Candidate code never
  runs inside the Go checker process.
- `examples/starters/priority-queue-rs` — untrusted seed: a mutex plus
  `BinaryHeap` implementation that is correct and intentionally slow.
- `examples/data-structures/prioqueue-{spsc,mpsc,spmc,mpmc}` — input bundles
  pointing at the shared evaluator and shared starter, differing only in
  scenario and producer/consumer counts.
- `examples/baselines/prioqueue-{locked-heap,sharded-heap,tbb}` — three
  reference implementations, outside the input bundles and never copied into
  optimization workspaces.

The baselines mirror the existing queue baselines. `locked-heap` is the
conservative control: one mutex over an array binary min-heap, with heap
entries holding `(priority, slot)` so sifting never moves payload bytes, and
copies inside the critical section. `sharded-heap` gives producers their own
locked heap and uses a shared slot pool as the global capacity token, so both
`memcpy` calls fall outside every lock.

`tbb` is the third-party anchor. The first two baselines were written against
this benchmark, so neither is independent of it; this one puts an unmodified
`tbb::concurrent_priority_queue` behind the ABI, making the scored number
Intel's structure rather than one written for this evaluator. oneTBB is
unbounded and stores no payload bytes, so the adapter supplies both: a tagged
Treiber stack of payload slots as the exact capacity token (the ABA tag is what
makes `VSPQ_FULL` exact rather than conservative), and a fixed-stride arena
addressed by `(priority, slot)` entries so payloads never move while TBB sifts.
Both `memcpy` calls fall outside every TBB operation. oneTBB is a system
dependency (`libtbb-dev`, Apache-2.0), not vendored, and validated against
2021.5. CI does not build baselines, so this adds no CI dependency.

TBB passes the strict `spsc`/`mpsc`/`spmc` gates and not only the relaxed
`mpmc` one because `concurrent_priority_queue` is flat-combining: operations
publish onto a lock-free list and one combiner thread executes the whole
pending batch alone, so each operation takes effect at a single point inside a
serialized batch. The combiner does reorder within a batch (pushes first, pops
deferred), but every operation in a batch overlaps every other in real time, so
that reordering stays inside the linearization freedom the batch already has.

Two places `tbb` deviates from a strict reading of the contract, documented in
its README rather than papered over. TBB cannot peek, so an undersized output
is only detectable after the item has been removed; the adapter requeues the
untouched entry and returns `VSPQ_INVALID`, leaving a window where the item is
invisible to other consumers. The contract asks for the item to stay queued.
This passes today only because the evaluator issues undersized dequeues from
the sequential ABI probe, never from the concurrent history, so the window is
unobservable. Second, flat combining blocks: a descheduled combiner stalls its
whole batch, so this is a blocking structure like the locked-heap control, not
a lock-free one. The try-style requirement still holds, since neither call ever
waits on queue state.

`sharded-heap` carries one non-obvious correctness detail. A consumer scanning
shards does not observe a globally atomic minimum, so the scan records the best
priority seen in any other shard and the pop commits only while the winning
shard still beats that runner-up, restarting otherwise. Without it, a second
consumer draining the winning shard between scan and pop makes the first
consumer return a non-minimum item. The restart is not a wait on another
operation: it re-reads current state and only triggers after another consumer
made progress, so the try-style contract holds.

The candidate interface is a copying C ABI (`include/vibesys_priority_queue_abi.h`),
normatively documented in `CANDIDATE_CONTRACT.md`. Copying is deliberate:
allocation and memcpy are inside measured time, so the candidate cannot win by
handing back borrowed pointers.

Two correctness models, and this is the part worth reviewing closely:

- SPSC / MPSC / SPMC use an exact bounded priority queue model, checked with
  Porcupine for linearizability.
- MPMC uses a **reservation-aware** model. A successful enqueue is split into a
  capacity reservation and a publication, both constrained to the original call
  interval. `FULL` observes reserved + published items; `EMPTY` and dequeue
  observe only published items.

The MPMC relaxation exists because the natural fast MPMC design (claim a slot,
then fill it) is not linearizable as a single atomic insert, and rejecting it
would force every candidate into a global lock. The relaxation is bounded: it
does not permit losing, duplicating, or fabricating items, and it does not
relax priority order among published items. Reviewers should focus on whether
that boundary is drawn where we want it, since it is the loosest contract in
the PR.

Both models treat published items as a multiset of `(priority, value)`, so
equal-priority items may be returned in any order.

### Architecture

```mermaid
flowchart LR
    subgraph go["Go evaluator process (trusted)"]
        gen[workload generation]
        hist[history + timestamps]
        porc[Porcupine model check]
    end

    subgraph rust["Rust native runner (trusted)"]
        probe[ABI probe]
        bench[benchmark threads]
    end

    cand[["priority-queue-candidate.so<br/>(untrusted)"]]

    gen -->|ops over stdio protocol| rust
    rust -->|call/return timestamps| hist
    hist --> porc
    probe --> cand
    bench -->|direct ABI calls| cand
```

Trust boundary: expected payloads, timestamps, histories, and the verdict live
in Go. The candidate shares an address space only with the Rust runner, which
avoids per-operation IPC in the benchmark. Scoring therefore assumes
cooperative native code; a crash, hang, malformed protocol frame, invalid ABI
status, or failed model check rejects the run.

Scoring order: `benchmark` runs the correctness gate first, then reports the
median `total_ops_per_sec` (successful enqueues + dequeues) across repetitions.

## Verification

Ran the Go and Rust test suites, both `cargo fmt`/`clippy` gates, bundle
validation for all four inputs, and an end-to-end
build → check → benchmark against the starter in a materialized workspace.

One rustfmt violation in `native_runner/src/abi.rs` was found and fixed. All
fmt and clippy gates were re-run under the toolchain CI actually pins (1.78.0),
not just the locally installed 1.97.0, since the new `cargo fmt -- --check`
steps gate these files for the first time. Both crates declare
`edition = "2021"` / `rust-version = "1.78"`, matching the queue crates, so no
toolchain bump was needed.

### One bug fixed

`runner.go` read `VIBESYS_QUEUE_NATIVE_RUNNER`, the same override the queue
evaluator uses. The temp build path and binary name had already been renamed,
so this was a missed rename rather than intent. Setting the variable for either
evaluator pointed the other at the wrong runner binary, and
`tests/test_queue_evaluator.py` sets it session-scoped. Now
`VIBESYS_PRIORITY_QUEUE_NATIVE_RUNNER`, verified by re-running the checker with
`VIBESYS_QUEUE_NATIVE_RUNNER=/bin/false` still set.

A file-by-file diff of both new trees against their `queue` originals, plus
greps for every stale symbol pattern, found no other missed renames.

### Known limitation, worth a reviewer's attention

While mutation-testing the `sharded-heap` regression test, the runner-up guard
was deliberately weakened to reintroduce the priority-inversion bug. The
correctness gate did **not** catch it: `mpmc` passed 3x100 trials at default
capacity, 3x300 trials at `--capacity 4`, and `spmc` passed 2x300 trials at
`--capacity 4`. The more severe variant found during development (committing
from an already-emptied shard) was caught at trial 64/100, so the gate is not
blind here, but narrow priority-inversion windows can survive it at these trial
counts. That matters because this evaluator is the scoring gate for an
optimizer. Not addressed in this PR; flagging it as a real sensitivity limit of
the current workload generator.

### Correctness properties

- ABI version is checked against `VSPQ_ABI_VERSION` before any operation.
- Handles are created per producer/consumer, confined to one native thread, and
  destroyed before the queue.
- Enqueue input pointers are borrowed for the call only; the candidate must
  copy the complete value.
- Dequeue returns a minimum-priority published item; ties break arbitrarily.
- `VSPQ_EMPTY` leaves output storage and `output_length` untouched;
  `VSPQ_INVALID` on insufficient output leaves the item queued.
- The candidate never writes beyond `output_capacity` and never retains the
  output pointer.
- `VSPQ_FULL` is legal only at capacity (reserved + published under MPMC).
- Operations are try-style and must not block on another operation's progress.
- Every successfully enqueued item is returned exactly once, in every scenario:
  no loss, no duplication, no fabrication.
- Probe coverage includes value lengths 0 through the configured maximum,
  including non-word-aligned sizes.
- The final benchmark drain checks item-count conservation and a commutative
  fingerprint over enqueued and dequeued payloads.

External contracts introduced (all new): `vibesys_priority_queue_abi.h`,
`CANDIDATE_CONTRACT.md`, the four input manifests, and the `check` /
`benchmark` CLI surface. One existing-shaped contract changed: the runner
override env var is now `VIBESYS_PRIORITY_QUEUE_NATIVE_RUNNER`. It was
introduced on this branch and never released under the old name, so nothing
downstream depends on it.

### Testing

| Command | Result |
| --- | --- |
| `uv run pytest tests/test_prioqueue_evaluator.py` | 18 passed, 12.7s |
| `./scripts/check_format.sh` / `check_lint.sh` / `check_types.sh` | all clean |
| `go test -race ./...` (evaluator) | ok |
| `go vet ./...` (evaluator) | clean |
| `cargo test --locked` (native_runner, starter) | 2 + 1 passed |
| `cargo fmt -- --check` under **1.78.0**, all 4 crates | clean |
| `cargo clippy --locked --all-targets -- -D warnings` under **1.78.0** | clean |
| `./vs validate examples/data-structures/prioqueue-{spsc,mpsc,spmc,mpmc}` | all 4 valid |
| `make clean all` all three baselines, `-Werror` | clean |

End-to-end against the starter in a materialized workspace:

```
make
go -C _evaluator/priority-queue run . check --workspace "$PWD" \
  --scenario <s> --operations 16 --trials 20
```

| Scenario | Result |
| --- | --- |
| spsc | PASS, linearizable, 1P/1C |
| mpsc | PASS, linearizable, 4P/1C |
| spmc | PASS, linearizable, 1P/4C |
| mpmc | PASS, reservation-aware, 4P/4C |

The C baselines were separately checked at `--operations 24 --trials 100`
across all four scenarios (8/8 pass), plus 3x`--operations 32 --trials 400` on
`sharded-heap` mpmc. Both were clean under ThreadSanitizer and under
ASan+UBSan with a 4P/4C, capacity-16, 80k-item driver exercising value lengths
0-64 and a short output buffer to hit `VSPQ_INVALID`.

`tbb` was checked the same way: `--operations 24 --trials 100` across all four
scenarios (4/4 pass, including the strict `spsc`/`mpsc`/`spmc` gates), plus
`mpmc` at `--capacity 4 --value-size 64` to squeeze the reservation window.

The two source-assertion tests were mutation-checked rather than assumed
meaningful: deleting the sharded guard, weakening it to `shard->count == 0`,
and hoisting the locked-heap `memcpy` past the unlock each make them fail,
while reformatting the guard does not.

Throughput, median of 5x3s reps, capacity 1024, 8-byte values:

| scenario | locked-heap | sharded-heap | rust starter |
| --- | --- | --- | --- |
| spsc | 2,166,728 | 1,340,728 | 2,475,162 |
| mpsc | 681,990 | 485,319 | 874,598 |
| spmc | 794,414 | 635,473 | 1,010,815 |
| mpmc | 833,431 | 895,339 | 877,438 |

Measured on a 256-core EPYC 9554 under load average ~148 with roughly ±40%
within-config spread, so these are same-order comparisons only.

Two honest results: `sharded-heap` only wins on mpmc and only narrowly, because
spsc/spmc have a single producer and degenerate to the control plus a second
lock, while mpsc bottlenecks on its lone consumer now paying a shard scan per
dequeue. And neither hand-written baseline decisively beats the naive Rust
starter at 8-byte values, whose `Mutex<BinaryHeap<Entry>>` is simply
competitive at this size; an adaptive-mutex variant was tested and made no
difference, so lock type is not the explanation.

That is the gap `tbb` is for. Separate benchmarking session (3 reps, same
machine and settings), so compare within this table, not against the one above:

| scenario | locked-heap | sharded-heap | oneTBB | oneTBB vs. best |
| --- | --- | --- | --- | --- |
| spsc | 2,147,349 | 1,666,962 | 1,762,788 | 0.8x |
| mpsc | 674,932 | 566,260 | 1,675,563 | 2.5x |
| spmc | 767,021 | 714,362 | 1,099,051 | 1.4x |
| mpmc | 808,357 | 872,443 | 2,229,637 | 2.6x |

Treat these as one significant figure. `mpsc` is noisy across repetitions for
every implementation: five 3-repetition runs of oneTBB produced medians from
1.3M to 2.2M, so that row is the median of those medians rather than a single
run.

TBB wins where the contention is real, and it is the only one of the three that
does not collapse in `mpsc`. It loses to the locked-heap control on `spsc`,
where flat combining pays for machinery no single-threaded pair needs. `spmc`
is its weakest contended case: every consumer serializes through the combiner,
so adding consumers buys less than adding producers does.

So the two hand-written baselines remain useful as correctness references and
weak as performance anchors until `--value-size` grows, and `tbb` supplies the
performance anchor they do not: a 2.5-2.6x headroom marker on the contended
scenarios that a candidate has to actually beat.


rairai77:
Superseded by #345, which has the same branch and contents but is opened from `uw-syfi:algos/prioqueue` rather than the fork. No review history to carry over.

PR #325, opened 2026-08-10
Add a headroom profiler kind for kernel-level optimization loops

kamahori:
## Problem

Optimization loops that target GPU kernels need per-round evidence about where recoverable time actually is. Raw profiler output (nsys, torch) ranks kernels by observed time, but a large kernel already at its hardware roofline is not an opportunity, while a mid-sized copy that a fused pipeline would never materialize is. External harnesses (e.g. a serving repo's perf tooling) can produce such "headroom reports" per kernel: observed time vs roofline speed-of-light, with the gap bucketed into kernel quality, eliminable data movement, and missing fusion. There is currently no profiler kind that puts such a report in front of the profiler agent as tools, so this evidence cannot drive multi-round hypothesis selection inside the loop.

## Solution

A new `headroom` profiler kind, following the documented `docs/extending-profilers.md` path end to end. Capture is deliberately target-owned: the bundle ships a capture entry point (conventionally `benchmark/headroom_capture.sh <output.json>`, or whatever `OBJECTIVE.md` documents) that measures the candidate and writes a report; the profiler analyzes that JSON. This keeps the framework free of any assumption about how a given target measures itself, and keeps the capture script inside a trusted input path.

- `ProfilerKind.HEADROOM` + `ProfilerDefinition` allowing `llm-serving` and `generic` domains. Analysis needs no host tooling and no domain torch support, so preflight always passes.
- `resources/profilers/headroom/analyze_headroom.py`: subcommand CLI over the report (`waterfall`, `top`, `kernel`, `subgraphs`, `compare`, `summary`), with the report schema ("headroom report v1") documented in the module docstring. `compare` diffs two rounds' reports (bucket deltas plus per-kernel observed deltas), which is the key multi-round signal: did the last hypothesis shrink its targeted bucket.
- `resources/profilers/headroom/server.py`: FastMCP server wrapping the same commands, mirroring the torch/nsys servers. Malformed reports surface as structured `error:` strings rather than exceptions.
- `src/vibesys/prompts/loops/agent/profilers/headroom.j2`: instructs the agent to rank by `opportunity_ms_step` rather than raw time, to persist reports under `progress/profiles/` for cross-round `compare`, to respect report `caveats`/`notes`, and to report a missing capture entry point as a capability gap instead of fabricating data. Includes the shared observer-effect discipline and the canonical perf-metric rules.

### Architecture

Same ownership model as the existing kinds: the registry declares behavior policy, packaging names derive from `kind.value` by convention, and `mcp_spec`/workspace materialization need no changes. The one boundary choice reviewers should inspect: the profiler consumes a JSON contract produced by target-owned capture code, so the schema is documented on the analysis side and treated tolerantly (unknown fields ignored, missing optional sections degrade to a message).

### Correctness properties

- A conventional profiler addition: no new context fields, sandbox mounts, CLI flags, or MCP/prompt dispatch mappings.
- Packaging names derive uniformly from `kind.value` (covered by the existing uniform-packaging test).
- The generic domain now legitimately allows `headroom`; every other non-CPU active kind is still rejected for `domain='generic'` (test updated with the rationale in a comment).
- Malformed or missing reports produce structured errors, never fabricated analysis.

### Testing

- `uv run pytest tests/test_profilers.py tests/loops/test_profiler_mcp.py tests/test_context.py -q` — 1808 passed (new: headroom domain/preflight test; six headroom MCP server tests covering tool registration, opportunity ranking, class filter, substring detail, cross-round compare, malformed-report error path).
- `uv run pytest tests/ -q` — full suite green except the two pre-existing `tests/_agent_cli/test_hostsandbox.py` failures, which fail identically on a clean `main` checkout on this host (bubblewrap environment), verified via `git stash`.
- `./scripts/format.sh`, `./scripts/check_format.sh`, `./scripts/check_lint.sh` — clean.
- `uv run pyright src/vibesys/profilers.py` — 0 errors (`resources/` is outside the pyright project roots, matching the existing profiler support packages).

### DIFF
--- a/src/vibesys/profilers.py
+++ b/src/vibesys/profilers.py
@@ -20,6 +20,7 @@ class ProfilerKind(StrEnum):
     NEURON = "neuron"
     MACOS_CPU = "macos_cpu"
     LINUX_CPU = "linux_cpu"
+    HEADROOM = "headroom"
 
 
 @dataclass(frozen=True)
@@ -83,6 +84,13 @@ PROFILER_DEFINITIONS: dict[ProfilerKind, ProfilerDefinition] = {
         ProfilerDefinition(ProfilerKind.NEURON, frozenset({DomainName.LLM_SERVING})),
         ProfilerDefinition(ProfilerKind.MACOS_CPU, frozenset({DomainName.GENERIC})),
         ProfilerDefinition(ProfilerKind.LINUX_CPU, frozenset({DomainName.GENERIC})),
+        # Analyzes a target-captured kernel headroom report (observed vs
+        # roofline speed-of-light per kernel). Scoped to LLM serving, where
+        # kernel-level roofline analysis drives the optimization loop.
+        ProfilerDefinition(
+            ProfilerKind.HEADROOM,
+            frozenset({DomainName.LLM_SERVING}),
+        ),
     )
 }
 
--- a/resources/profilers/headroom/__init__.py
+++ b/resources/profilers/headroom/__init__.py
@@ -1,0 +1,1 @@
+"""Kernel headroom-report profiling helpers exposed for tests."""
--- a/resources/profilers/headroom/analyze_headroom.py
+++ b/resources/profilers/headroom/analyze_headroom.py
@@ -1,0 +1,267 @@
+#!/usr/bin/env python3
+"""Kernel headroom report analysis toolkit — subcommand-based.
+
+Unlike the nsys/torch profilers, this toolkit does not capture anything
+itself. The *target* owns capture: its bundle ships a capture entry point
+(conventionally ``benchmark/headroom_capture.sh <output.json>``, or whatever
+``OBJECTIVE.md`` documents) that measures the candidate and writes a headroom
+report. This toolkit analyzes that report: which kernels are farthest from the
+hardware's speed-of-light, and whether the gap is kernel quality, eliminable
+data movement, or missing fusion.
+
+Usage:
+    python analyze_headroom.py waterfall report.json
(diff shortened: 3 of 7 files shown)


src/vibesys/profilers.py
```diff
@@ -83,6 +84,13 @@ def error_message(self) -> str:  # noqa: D102  # tracked: #288
         ProfilerDefinition(ProfilerKind.NEURON, frozenset({DomainName.LLM_SERVING})),
         ProfilerDefinition(ProfilerKind.MACOS_CPU, frozenset({DomainName.GENERIC})),
         ProfilerDefinition(ProfilerKind.LINUX_CPU, frozenset({DomainName.GENERIC})),
+        # Analyzes a target-captured kernel headroom report (observed vs
+        # roofline speed-of-light per kernel); capture is target-owned, so no
+        # host tooling or torch support is required of the domain.
+        ProfilerDefinition(
+            ProfilerKind.HEADROOM,
+            frozenset({DomainName.LLM_SERVING, DomainName.GENERIC}),
```
vic-lsh: This should probably only be in the LLM serving domain.

vic-lsh (changes requested):
LGTM except for my comment.

src/vibesys/profilers.py
kamahori: Done in f06ba0a: scoped HEADROOM to `{DomainName.LLM_SERVING}` only (dropped GENERIC). Updated `test_headroom_profiler_domains_and_preflight` to assert it's rejected for the generic domain. Also merged latest main to clear the conflict.

PR #326, opened 2026-08-10
Ship resources/ in the wheel so installed runs get profilers and skills

kamahori:
## Problem

`pip install vibesys` delivers a working engine and (when built with a JavaScript toolchain) the TUI, but silently loses every bundled profiler support package and the preset skills library: `context.py` resolved profiler support from `PROJECT_ROOT / "resources" / "profilers"` and the default skill root was the checkout-relative `resources/skills`. Neither exists in an installed layout, so no-clone runs get no profiler MCP servers and no preset skills, with no diagnostic. Skills have an escape hatch (`--skills-dir` accepts absolute paths); profilers have none.

## Solution

Mirror the existing `_tui` staging pattern for `resources/`:

- `resources_packaging.py` (repo root, stdlib-only, like `tui_packaging.py`): copies `resources/profilers` and `resources/skills` into `vibesys/_resources`, excluding `.git`, `__pycache__`, `*.pyc`, and the vendored skill `repos/` submodule checkouts (already excluded from workspace materialization; they would bloat the wheel by hundreds of MB). A missing `resources/` tree is a clean no-op, matching the best-effort TUI stage.
- `setup.py` stages resources from the same custom `build_py` step; `pyproject.toml` adds the `_resources/**/*` package-data glob; `MANIFEST.in` grafts the resource trees into the sdist (minus `repos/`).
- New `src/vibesys/resource_paths.py`: `resources_root()` prefers the checkout (`PROJECT_ROOT/resources`) and falls back to the staged `vibesys/_resources` (via `importlib.resources`, same lookup the TUI launcher uses), so a developer editing `resources/` always sees their edits. `profiler_support_dir(kind)` and `default_skill_roots()` are the two consumers' entry points.
- `context.py` resolves profiler support through `profiler_support_dir()`; `main.py` applies skill defaults through `default_skill_roots()`. The checkout-relative `DEFAULT_SKILL_ROOTS` constant in `skills.py` had exactly one consumer (that default-application site) and is removed.

### Architecture

Shared interface (`resource_paths`) owns location policy; the wiring sites (`context.py`, `main.py`) consume it; the build-time stager is a repo-root module imported only by `setup.py`. No behavior change in a checkout: `resources_root()` returns the same paths as before.

### Correctness properties

- Checkout behavior is unchanged: with `PROJECT_ROOT/resources` present, all resolved paths are identical to before.
- Installed behavior: profiler support and preset skills resolve from `vibesys/_resources`; when neither location exists, resolution returns `None`/`()` exactly where the old code produced a missing path (no fabricated defaults).
- The wheel never ships vendored skill `repos/` checkouts or `__pycache__`.

### Testing

- `uv run pytest tests/ -q` — 2792 passed; the only exclusions are the two pre-existing `tests/_agent_cli/test_hostsandbox.py` failures that fail identically on a clean `main` on this host (bubblewrap environment).
- New `tests/test_resources_packaging.py`: staging copies both trees, drops `repos/`/`__pycache__`, is idempotent, no-ops without `resources/`; `resource_paths` prefers the checkout, falls back to a staged copy, and returns `None`/empty when neither exists.
- `tests/test_context.py`: the 12 fixtures that relocate `PROJECT_ROOT` now also patch `vibesys.resource_paths.PROJECT_ROOT`, keeping the existing support-dir assertions meaningful.
- End-to-end: `uv build --wheel` stages 292 resource files (verified via zip listing: `profilers/nsys/server.py` and `skills/serving-systems/SKILL.md` present, no `repos/`, no `__pycache__`); unzipping the wheel and importing from that layout resolves `resources_root()`, `profiler_support_dir("nsys")`, and `default_skill_roots()` from `vibesys/_resources`.
- `./scripts/format.sh`, `./scripts/check_format.sh`, `./scripts/check_lint.sh` — clean; `uv run pyright src/vibesys/resource_paths.py src/vibesys/context.py src/vibesys/main.py src/vibesys/skills.py` — 0 errors.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -8,3 +8,9 @@ include clients/tui/tsconfig.check.json
 include clients/tui/README.md
 prune clients/tui/dist
 prune clients/tui/node_modules
+# Include the resource trees so a source build can stage them into
+# vibesys/_resources. Vendored skill repo checkouts stay out of the sdist.
+graft resources/profilers
+graft resources/skills
+prune resources/skills/serving-systems/repos
+global-exclude __pycache__ *.pyc
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ include = ["vibesys*"]
 [tool.setuptools.package-data]
 # ``_tui`` is produced at build time by setup.py's build_py step and injected
 # into the wheel's build tree; the glob keeps it as package data when present.
-vibesys = ["prompts/*.j2", "prompts/**/*.j2", "prompts/**/*.md", "domains/README.md", "_tui/**/*"]
+vibesys = ["prompts/*.j2", "prompts/**/*.j2", "prompts/**/*.md", "domains/README.md", "_tui/**/*", "_resources/**/*"]
 
 [dependency-groups]
 dev = [
--- a/resources_packaging.py
+++ b/resources_packaging.py
@@ -1,0 +1,55 @@
+"""Stage bundled framework resources for inclusion in the Python wheel.
+
+Profiler support packages (``resources/profilers``) and the preset skills
+library (``resources/skills``) are resolved from the repository checkout at
+run time. An installed wheel has no checkout, so the wheel build copies both
+trees into the ``vibesys._resources`` package directory; ``setup.py`` calls
+:func:`stage_resources` from the same custom ``build_py`` step that stages the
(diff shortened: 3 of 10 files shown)


vic-lsh (approved):
I'll merge this and open a new PR that revolves other publishing blockers

PR #327, opened 2026-08-10
Add hypothesis log feature to VibeSys TUI for experimental observations

AyanBinRafaih:
## Problem

VibeSys already runs a scientific loop and already records its results, but the operator never sees the record. Each round the orchestrator proposes one falsifiable hypothesis; the implementer reports a `HypothesisOutcome`; the judge validates it; and `loop.py` resolves the pair into `proven`, `rejected`, or the terminal outcome value. That resolution is persisted three times over, in `rounds.json`, in `active_hypothesis.json`, and as prose in `progress.md`.

None of it reaches the client. Grepping `hypothesis` across `clients/tui/src` and `src/vibesys/server` returns nothing: the protocol carries no hypothesis field and the TUI has no view of one. `/history` returns a flat list of run events and the rounds strip presents rounds as an undifferentiated sequence, so the only way to learn what was tried and whether it held is to open each round and read prose.

Rounds are also the wrong unit, and they are the unit the client opens on. One hypothesis spans one or several rounds through the continuation machinery, so reading round by round scatters the evidence for a single claim across several entries. On a run of 100 rounds the shape of the investigation, which claims were tested, which held, which were abandoned, is invisible even to an operator willing to read everything. The per-round transcript is the right level of detail once you know which hypothesis you care about, and too much to ingest before that.

## Solution

Demonstration:

https://github.com/user-attachments/assets/1f09883a-ddfa-4b6c-9bb9-aedc16fd6178

Group the persisted round state by hypothesis and expose it over the client protocol, then make it the view the client opens on: one row per unit of investigation rather than one per round. The existing per-round transcript becomes what the operator opens by selecting a hypothesis, not what greets them.

**Assembly:** `src/vibesys/server/experiments.py` groups `rounds.json` records by `hypothesis_id` and merges the live plan from `active_hypothesis.json`. `resolved_outcome` is copied from the closing round's `hypothesis_outcome`, never recomputed, so the client cannot drift from the framework's resolution semantics in `loop.py`. `judge_verdict` distinguishes a deferred judge from a failing one: sparse-review policy leaves a round provisional, which is reported as no verdict rather than `Fail`. `kept` is derived from `official_evaluation` and `candidate_disposition`, mirroring the split `CandidateDisposition` already makes between hypothesis truth and checkpoint utility, so a disproven claim can still show `Yes` and a proven one `No`.

**Protocol:** `query.experiments` returns `HypothesisEntry` rows. The addition is backward compatible, so `PROTOCOL_VERSION` stays at 1. `clients/tui/src/generated/` is regenerated from the Python models, as `tests/test_tui.py` enforces.

**Landing view:** the log occupies the main pane from startup, and the controller loads it before the first frame. The rounds strip, agent map, and transcript are per-round chrome, so none of them are drawn while the table is up. `/history` returns to it; `/history rounds` keeps the previous flat list.

**Drill-down:** Enter on a row, or `/open-round`, sets a hypothesis scope, and the ordinary client comes back filtered to that hypothesis's rounds: the transcript shows every round it spans rather than only the latest, the rounds strip covers those rounds alone, and `[`/`]` navigate within them. Escape steps back to the table with the selection intact. `/open-round --N` jumps straight to round N inside whichever hypothesis owns it, which is the fast path when a log line or a `progress.md` entry names a round. Selection is keyed by hypothesis id rather than row index, so a refetch that inserts a row above the current one leaves the operator where they were.

**One route in:** the log is the root view and cannot be dismissed. Opening a hypothesis is the only way to reach per-round output, so the transcript is always entered through the claim it belongs to. `Ctrl+L`, which previously meant "unfiltered live output", now returns to the table, and there is no Escape from the table to anywhere.

**Navigation:** the table body is a scroll box, so the wheel and trackpad move the viewport while the arrow keys move the cursor, and clicking a row selects it. Keyboard selection only nudges the scroll position when the selected row would otherwise be off screen, so the two do not fight.

**Staying current:** the table refetches when an agent phase or a round finishes, which bounds every change to the log, behind a single-flight guard so a burst of phase events costs one request. Rows order by first round and never reshuffle. Columns drop widest first as the terminal narrows; hypothesis, rounds, and outcome always survive. Records written before hypothesis tracking render as `(Unidentified)`, one row each, rather than being merged or dropped.

**Colour:** the outcome is its own renderable so the resolution can carry `success` when a hypothesis held and `error` when it did not, without recolouring the row. Both come from the theme, so the reading holds in all eight. Following the same rule as #265, colour is never the only signal: the outcome is spelled out in the cell.

What the client opens on, at 120 columns. Both frames below are captures from the test renderer; the first is fed the assembled log of a real `--stub-agent` run, which is why the rows are real and the status line is not:

```
VibeSys · connecting · starting · no round yet · experiments
╭─ Experiments ────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│  Hypothesis     Rounds   Implementation Details                               Measured    Verdict  Outcome     Kept  │
│  H-01           1-2      Batching the prefill step removes per-request launc… 1000        Pass     Proven      No  █ │
│  H-02           3-4      A larger KV cache block trades memory for fewer all… +4.5%       Pass     Proven      No  █ │
│  H-03           5-6      Reordering the sampler avoids a redundant device sy… —           Pass     Disproven   No    │
│  H-04           7        Batching the prefill step removes per-request launc… +8.6%       Pass     Continue    No    │
│ 1/4 · ↑↓ or scroll: select · Enter or /open-round: open its rounds                                                   │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
↑↓ or scroll: select · Enter or /open-round: open its rounds · /open-round --N
```

`H-01` has no preceding measurement to compare against, so it shows the absolute metric; `H-03` was resolved without an official evaluation, so it has none to show.

### Architecture

```mermaid
flowchart TD
    LOOP[agent loop] --> RJ[logs/rounds.json<br/>one record per round]
    LOOP --> AH[logs/active_hypothesis.json<br/>live plan only]
    RJ --> B[server/experiments.py<br/>build_experiment_log]
    AH --> B
    B --> BL[apply_baselines<br/>delta vs last prior measurement]
    BL --> Q[query.experiments → HypothesisEntry]
    Q --> C[session-controller<br/>load at start · refetch on phase/round finished]
    C --> M[session-model<br/>selection keyed by hypothesis id]
    M --> T[experiment log: one row per hypothesis<br/>landing view, fills the main pane]
    T -->|Enter sets hypothesisScope| D[transcript · rounds strip · agent map<br/>filtered to that hypothesis]
    D -->|Esc clears the scope| T
```

Note: the claim text is the one field the grouping could not recover. `active_hypothesis.json` holds only the plan currently in flight, so a resolved hypothesis has no claim left on disk. This adds `hypothesis_claim` and `hypothesis_task` to the existing `_RoundRecord`, which is the record the issue already names as a source. No new tracking machinery is added to the loop, and older logs read back as `None`.

## Verification

The two things to check are that the grouping matches what the loop actually recorded, and that the resolution the table shows is the loop's own rather than a second derivation. Both are covered by tests against constructed records and by a real stub run.

These are also specific behaviours in this case for the feature:

| Case | Shown as | Why |
| --- | --- | --- |
| Judge passed, implementer reported `disproven` | `Pass` / `Disproven` | Verdict and outcome are different questions; `loop.py` keeps the implementer's terminal value |
| Round left unreviewed by sparse-review policy | verdict `—` | Provisional, not failed |
| Rejected claim, candidate on the Pareto frontier | `Rejected` / kept `Yes` | Integration is independent of truth |
| Hypothesis started, first round unfinished | row present, outcome `Active` | Work in flight is visible before it lands |
| Run with no hypotheses yet | empty table naming what happens next | The landing view still has to say something useful on a cold start |
| Hypothesis resolved between official evaluations | measured `—` | The framework records a verified metric only when its own gates ran; there is no number to show |
| Record with no `hypothesis_id` | `(Unidentified)`, one row per round | Old logs stay complete; unrelated rounds are not merged |

### Correctness properties

- The client opens on the log, with the transcript, rounds strip, and agent map absent until a hypothesis is opened, and no key dismisses it.
- `Measured` shows the verified metric of the resolving round, as a delta once a preceding measurement exists, and `—` when the sparse evaluation cadence produced none.
- Continuation rounds sharing a `hypothesis_id` collapse into one row with the correct round range, and opening it shows exactly those rounds and no others.
- `resolved_outcome` is the value the loop wrote for the closing round. The client copies it and never re-derives satisfaction.
- An active hypothesis is marked, carries no outcome while mid-continuation, and appears even before its first round finishes.
- `kept` is independent of the verdict in both directions, derived from persisted gate state and disposition rather than inferred.
- Rows order by first round and hold that order as new rounds land; selection survives a refetch that inserts rows above the current one.
- Leaving a hypothesis returns to the table with the previous selection intact, and that is the only navigation between the two views.
- Inside a hypothesis, round navigation and the rounds strip are confined to that hypothesis's rounds.
- The table scrolls with the wheel and the trackpad as well as the keyboard, and the wheel does not move the selection. Header and rows fit the panel exactly at every width the layout degrades through, including the reserved scrollbar gutter, so no column is lost to truncation.
- `/open-round --N` resolves N to the hypothesis that owns it and lands on that round; a round belonging to no hypothesis is reported rather than ignored.
- The view updates in place as phases and rounds finish, without the operator reopening it, and a burst of events costs one request.
- Runs whose records carry no `hypothesis_id` render placeholders instead of failing or dropping rows.
- Every colour comes from the selected theme; no colour literal is added outside `ui/theme.ts`, and the outcome word is always present so colour is never load-bearing.
- The flat round list is unchanged under `/history rounds`. Round-level navigation and the rounds strip keep their behaviour but are now scoped: inside a hypothesis they cover its rounds, and outside one they are not drawn.

### Testing

- `bun --bun vitest run` (in `clients/tui`) — 192 passed, 15 files. Adds `experiment-log.test.ts` (16 cases: column degradation, exact panel fit at seven widths, casing, measured fallbacks, placeholder rendering, outcome colour across all eight themes, selection keying) plus 7 cases in `app.test.ts` (landing view with the per-round chrome absent, opening a trajectory and stepping back out, themed outcome colour in light and dark via `captureSpans`, 120-row keyboard scrolling, wheel scrolling that leaves the selection alone, empty and placeholder states) and 13 in `session-controller.test.ts` (load before the first frame, `/history` routing, `/history rounds`, refetch keeping the selected row, coalesced refetch under a burst, no refetch for events that cannot change the log, the log surviving as the root view, trajectory round trip, query failure, `/open-round` on the selection, `/open-round --N`, an unknown round, and `/open-round` inside a hypothesis). Rendering was also read at 120, 100, and 72 columns against a real run's data.
- `uv run pytest tests/server tests/test_tui.py tests/loops/agent/test_orchestrate.py` — 167 passed. Adds `tests/server/test_experiments.py` (10 cases: grouping, outcome mapping, deferred verdict, ordering, `kept` independence, baseline deltas, service wiring, unreadable state files) and 2 `_RoundRecord` round-trip cases including a legacy record with no plan text.
- `npx tsc -p tsconfig.check.json` — passed. `npx biome check .` — passed, 44 files.
- `uv run ruff check .` and `uv run ruff format --check .` — passed, 311 files. `uv run pyright` — 0 errors.
- End to end without an API key: `./vs --stub-agent --local --max-rounds 7 --headless`. The stub runner now emits a hypothesis every two rounds with every third disproven, and requests an official evaluation on each closing round, so the continuation, proven, disproven, measured, and unmeasured paths are all reachable offline. The assembled log for that run is the table shown above, cross-checked against the `hypothesis_outcome` values in `rounds.json`.
- `grep -rlE '#[0-9a-fA-F]{6}' --include='*.ts'` over non-test, non-generated sources — returns `ui/theme.ts` only, unchanged from #265.

Note: `launcher.test.ts` intermittently times out at its 5s limit. It is not caused by this change. Run in isolation three times on each side, `starts a headless backend, waits for readiness, and runs the frontend` takes 3186/3349/3291ms on this branch and 3481/3163/3213ms on a stashed `main`: the same ~65% of its budget either way, which tips over whenever the machine is busy. The test also spawns a fake frontend that exits immediately, so it never runs the controller code this branch touches.


vic-lsh:
@AyanBinRafaih Could you fix the merge conflicts? Thanks!

AyanBinRafaih:
@vic-lsh I've resolved the conflicts.

vic-lsh:
Thanks Ayan! 

When I enter a slash command from the hypothesis log, the first Enter opens the selected hypothesis and shows its agent messages. I have to press Enter again to run the command. Instead, I think the hypothesis log should be the "home screen" for this TUI, so all slash commands should directly open overlay views on the hypothesis screen.

Hope that makes sense. If not, I can try to re-explain

AyanBinRafaih:
@vic-lsh I've fixed that issue as well as resolved the conflicts. Hope this works fine now! Thank you!

Issue #318, opened 2026-08-10
Microservices: feed trace graphs and critical-path evidence into VibeSys


Ryclic:
### Workstream

Core/runtime

### Problem

VibeSys can collect OpenTelemetry data for microservice evaluations, but raw spans and flat per-service summaries do not directly explain the request path that controls end-to-end latency. Agents need a compact, stable representation of the workload-specific service graph, timing relationships, and critical-path contributors. Without it, they must infer causality from large telemetry dumps and may prioritize services that are expensive in aggregate but are not on the latency-critical path.

The trace graph CLI work establishes a formatted and versioned representation, but critical-path analysis and VibeSys loop integration remain separate follow-up work.

### Desired outcome

For each microservice evaluation window, ServiceBench derives a versioned trace graph from OTel spans, calculates critical-path evidence for complete traces, and emits both machine-readable JSON and a deterministic text rendering. VibeSys consumes the JSON through its profiler path and gives agents concise evidence about which services and operations contribute to end-to-end latency.

The feature is complete when a VibeSys microservice run can cite trace-derived critical-path evidence when proposing an optimization, while benchmark metrics and correctness checks remain the sources of truth for accepting that optimization.

### Acceptance criteria

- [x] ServiceBench emits a versioned trace graph for the benchmark measurement window, with services, operations, call edges, trace counts, errors, and latency statistics.
- [x] The same graph has a deterministic text rendering with service topology, per-operation timing, and readable per-trace timelines.
- [ ] Critical-path analysis accounts for parent-child relationships and overlapping spans without naively summing inclusive durations.
- [ ] The report identifies critical-path services and operations, their latency contribution, and the number of traces supporting each result.
- [ ] Incomplete, disconnected, or ambiguous traces are reported explicitly and excluded from unsupported critical-path claims.
- [ ] VibeSys discovers the versioned JSON artifact through the existing profiler boundary and includes a bounded summary in the microservice profiler agent context.
- [ ] Agent instructions distinguish diagnostic trace evidence from authoritative benchmark scores and correctness results.
- [ ] An end-to-end microservice evaluation demonstrates an agent citing specific critical-path evidence from the generated report.
- [ ] Existing ServiceBench evaluation commands and existing flat OTel summaries remain compatible.
- [ ] Relevant unit, integration, and end-to-end tests pass.

### Scope and non-goals

In scope:

- OTel span normalization into a workload-window trace graph.
- Deterministic JSON and text output.
[...]

PR #360, opened 2026-08-17
Make the VibeSys loop able to consume OTel critical-path evidence

Ryclic:
## Problem

Follow-up to #339, advancing #318.

#339 landed the producer side: `servicebench trace` emits a versioned trace
graph with critical-path evidence, and the otel profiler MCP server exposes
bounded views of it. What it did not do is make that evidence reachable from a
profiler turn. On a Linux host the loop could not consume any of it, and the
failure was silent: the profiler agent ran, found no tools, and fell back to
reasoning from the scalar benchmark number alone.

Three defects, all in the seam between the loop and the evidence.

1. **The profiler MCP server never started.** `mcp_spec` in
   `loops/profiler.py` emits `MCPServerSpec(command="python", ...)`. The CLI
   agent resolves its environment through `bash -i -c env`, so `python` binds
   to whatever the login shell's PATH provides, not the interpreter running
   VibeSys. The server exited with `ModuleNotFoundError: No module named
   'mcp'`, and the agent lost every profiler tool with no error surfaced to
   the loop.

2. **Pinning the interpreter was not sufficient under bubblewrap.** A uv
   virtualenv reaches its base install through an alias directory
   (`cpython-3.14` -> `cpython-3.14.7`), while `sys.base_prefix` is already
   symlink-resolved. Importing only the resolved prefix leaves the alias
   dangling inside the sandbox, so every `sys.executable` exec failed with
   ENOENT.

3. **The bounded tools exposed rankings but no structure.** `critical_path`
   returns ordered contributors; nothing let the agent see call topology or
   per-request timing. Issue #318 asks for an agent that can cite *which*
   services and operations control end-to-end latency, which needs the shape
   of the request, not just a leaderboard.

## Solution

1. `_resolve_mcp_interpreter` in `agents/cli_runner.py` rewrites `python` and
   `python3` MCP server commands to `sys.executable` on host runs. Container
   executors are left untouched: their `python` is the correct one inside the
   image, and the host interpreter path does not exist there.

2. `_interpreter_alias_roots` in `agents/host_resource_declarations.py` walks
   the symlink chain from `sys.executable` and declares any symlinked ancestor
   directories as host resources. This binds the real install under the name
   the interpreter actually walks. Only directories are declared; binding the
   individual hop files makes bwrap fail to create the mount point.

3. A new `trace_breakdown(path, telemetry_path, top)` MCP tool returns a
   depth-ordered call graph (inclusive and exclusive latency per node, edges
   restricted to retained nodes) plus the representative trace's waterfall,
   with explicit omitted-node and omitted-span counts so truncation is never
   silent. It reuses the same workload-identity binding as `summary`,
   `compare`, and `critical_path`: a graph is only reported when it shares
   workload name, hash, and measurement windows with the telemetry report.

4. The otel profiler prompt now names `trace_breakdown` alongside the other
   evidence tools, demotes `reports()`/`trace_graphs()` to discovery, states
   that artifacts written outside the workspace (for example under `/tmp`) are
   not visible to a later turn, explains that high inclusive with low exclusive
   latency means waiting on a callee, and requires the analysis to include the
   representative waterfall as ordered `offset_ms +duration_ms
   service:operation` lines.

### Architecture

No ownership changes. The producer boundary (Go, `servicebench`) and the
consumer boundary (Python, profiler MCP server) are unchanged; this repairs
the process-launch path between the loop and the server, and adds one tool
behind the existing artifact-validation helper.

```mermaid
flowchart LR
  subgraph go["servicebench (Go)"]
    T["trace mode"] --> A["telemetry.json + trace-graph.json"]
  end
  subgraph loop["VibeSys loop"]
    P["profiler round"] --> C["cli_runner"]
    C -->|"MCPServerSpec(command=python)"| R{"_resolve_mcp_interpreter"}
    R -->|"host: sys.executable"| M["otel profiler MCP server"]
    R -->|"container: unchanged"| M
  end
  subgraph sandbox["agent host sandbox"]
    H["_interpreter_alias_roots"] -.->|"binds alias dirs"| M
  end
  A --> M
  M -->|"summary / trace_breakdown / critical_path"| G["profiler agent"]
  G --> S["ProfilerSummary"]
```

### Correctness properties

- A graph is summarized only when it shares workload name, hash, and
  measurement windows with the telemetry report. `trace_breakdown` goes
  through the same `_bound_graph_to_report` helper as the existing tools, so a
  graph from another workload is rejected rather than silently mixed.
- Bounding is disclosed, not hidden. Roots, nodes, and waterfall spans are each
  capped at `top`, and every truncation is accompanied by an
  `omitted_*_count`, so an agent cannot mistake a truncated view for a complete
  one.
- Node retention is by p95 inclusive latency, but presentation is by depth,
  so the returned graph reads as a call structure rather than a ranking. Edges
  are filtered to retained nodes, so no edge references a node that was
  dropped.
- Interpreter pinning is host-only and Python-only. Non-Python MCP commands
  and all container executors keep their declared command verbatim.
- Alias declaration adds directories only, and only those that are actually
  symlinks. An interpreter with no alias in its chain declares nothing extra,
  so the sandbox is not widened on ordinary installs.

### Testing

Automated:

```
uv run pytest -q --no-cov
  4054 passed, 8 skipped, 1 failed in 4:29
```

The single failure is `tests/test_tui.py::test_cli_parse_failure_is_streamed_after_client_attaches`,
which is pre-existing and unrelated: it fails identically on a clean `main`
with these changes stashed.

```
uv run ruff check .        All checks passed
uv run ruff format --check All formatted
```

New tests: `trace_breakdown` returns a call graph and waterfall, bounds its
output while keeping the costly nodes, rejects a graph from another workload,
and returns a structured summary; the profiler tool set assertion now includes
it; interpreter resolution is host-only and Python-only; the runner sandwich
test asserts the installed spec carries the pinned interpreter; alias-root
declaration is covered with and without an alias present. Prompt leakage tests
assert the new tool signature, `representative waterfall`, and
`--trace-graph-json`.

Manual, end to end:

A microservices `evolve` run under a real bubblewrap sandbox reached the
profiler round, the MCP server started, and the agent called `summary`,
`trace_breakdown`, and `critical_path`, then produced a `ProfilerSummary`
citing specific critical-path evidence. Abbreviated from the run log:

```
=== Profiler ROUND START ===
→ mcp__vibesys-otel-profiler__summary(path=".../telemetry.json", top=15)
→ mcp__vibesys-otel-profiler__trace_breakdown(path=".../trace-graph.json",
    telemetry_path=".../telemetry.json", top=15)
→ mcp__vibesys-otel-profiler__critical_path(path=".../trace-graph.json",
    telemetry_path=".../telemetry.json", top=15)
```

From the resulting analysis:

```
Representative waterfall for trace-00062 (total 82.52ms):
  0.00 +82.52 frontend:GET /hotels;  0.63 +71.58 search:Nearby;
  1.19  +8.91 geo:Nearby;            1.41  +7.15 geo:mongo.find;
 10.77 +59.79 rate:GetRates;        11.02  +0.61 rate:memcached.get;
 11.73 +57.18 rate:mongo.find;      72.97  +8.55 profile:GetProfiles;
 ...
This single trace shows rate:mongo.find alone consuming 57.18ms of the
82.52ms root (69%). Aggregated critical-path contribution ranks
rate:mongo.find first: mean 38.89ms per occurrence, firing on 115/240
traces (47.9%), i.e. a memcached miss in the rate service.
```

The agent also disclosed coverage limits without prompting
(`async_relationships_excluded=480`, `unmatched_client_spans=880`) and
declined to convert contribution into a speedup ceiling, citing the absence of
an uninstrumented control. That is the behavior #318 asks for in its final two
acceptance criteria.

Not covered: this end-to-end run replayed a recorded OTLP capture through the
real `cmd/otelcapture` pipeline rather than driving a live Docker cluster, so
the Go producer, MCP tools, prompt, and agent reasoning are all exercised but
cluster spin-up is not. Two environment issues blocked a live run and are
worth separate follow-ups:

- The hotel-reservation task in the DeathStarBench submodule writes
  `--telemetry-output` and `--trace-graph-json` under `/tmp`, which the agent
  sandbox masks with a private tmpfs. The artifacts survive the turn that
  writes them but are invisible to later turns and to `reports()` discovery.
- The agent sandbox binds no Docker socket. `VIBESYS_AGENT_SANDBOX_ALLOW`
  does work for this (a read-only bind of a unix socket still permits
  `connect()`, verified directly), but it is documented nowhere, so the
  microservices domain fails under the default sandbox with an error that
  looks like a Docker problem rather than a sandbox policy one.

### Issue progress

Advances #318 rather than closing it. Lands the two integration criteria
("VibeSys discovers the versioned JSON artifact through the existing profiler
boundary and includes a bounded summary" and "an end-to-end microservice
evaluation demonstrates an agent citing specific critical-path evidence"),
with the caveat above that the end-to-end run used a recorded capture.

### DIFF
--- a/resources/profilers/otel/server.py
+++ b/resources/profilers/otel/server.py
@@ -448,6 +448,64 @@ class CriticalPathEvidence(BaseModel):
     omitted_root_count: int
 
 
+class TraceNodeEvidence(BaseModel):
+    """One call-graph node with its inclusive and exclusive latency split."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    node_id: str
+    path: str
+    service: str
+    operation: str
+    kind: str
+    depth: int
+    inclusive_latency_ms: LatencyDistribution
+    exclusive_latency_ms: LatencyDistribution
+
+
+class BoundedRepresentativeTrace(BaseModel):
+    """Bounded representative waterfall returned to the profiler agent."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    trace_id: str
+    duration_ms: float
+    spans: list[WaterfallSpan]
+    omitted_span_count: int
+
+
+class TraceBreakdownRootEvidence(BaseModel):
+    """Bounded call-graph and waterfall evidence for one root operation."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    service: str
+    operation: str
+    trace_count: int
+    error_count: int
+    latency_ms: LatencyDistribution
+    nodes: list[TraceNodeEvidence]
+    omitted_node_count: int
+    edges: list[TraceGraphEdge]
+    omitted_edge_count: int
+    representative_trace: BoundedRepresentativeTrace
+
+
+class TraceBreakdown(BaseModel):
+    """Bounded schema-v2 structure returned by the ``trace_breakdown`` tool."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    source: str
+    collected_at: str
+    workload_name: str
+    workload_hash: str
+    measurement_windows: list[MeasurementWindow]
+    quality: TraceQuality
+    roots: list[TraceBreakdownRootEvidence]
+    omitted_root_count: int
+
+
 def load_report(path: str) -> TelemetryReport:  # noqa: D103  # tracked: #288
     return TelemetryReport.model_validate_json(Path(path).read_text(encoding="utf-8"))
 
@@ -494,18 +552,7 @@ def summarize_critical_path(
 ) -> CriticalPathEvidence:
     """Return bounded critical-path evidence from a schema-v2 trace graph."""
     _validate_top(top)
-    graph = load_trace_graph(path)
-    telemetry = load_report(telemetry_path)
-    if (
-        graph.workload_name,
-        graph.workload_hash,
-        graph.measurement_windows,
-    ) != (
-        telemetry.workload_name,
-        telemetry.workload_hash,
-        telemetry.measurement_windows,
-    ):
-        raise ValueError(_GRAPH_REPORT_MISMATCH)
+    graph = _bound_graph_to_report(path, telemetry_path)
     roots = []
     for root in graph.roots[:top]:
         critical = root.critical_path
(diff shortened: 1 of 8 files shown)


vic-lsh (changes requested):
Blocked. See comment.

Issue #380, opened 2026-08-19
Express microservice optimization strategies as invocable skills rather than always-loaded prompt text


Ryclic:
### Workstream

Expansion

### Problem

Microservice optimization strategy lives in two places today, both of which are
always-loaded prompt text. `src/vibesys/prompts/domains/microservices/orchestrator.md`
lists seven "good task shapes" (connection pooling, gateway upstream tuning,
cache usage, database indexes, serialization, worker counts, baseline
establishment), and `implementer.md` names a similar set of levers. Every one of
them is parameter tuning inside a fixed service topology.

Two consequences follow.

First, strategies that change the shape of the request graph have nowhere to
live. Moving a call off the synchronous path, batching an N+1 call pattern, or
collapsing a call edge are all permitted by #44's core rule, and the evidence
needed to choose between them exists since #318 wired critical-path attribution
through the OTel profiler MCP boundary. Nothing tells an agent that these are
options, or how to pick between them from that evidence.

Second, the current mechanism does not scale. `README.md` states the
architecture rule that new optimization techniques are added by writing a skill
rather than by modifying the framework, and `orchestrator_plan_prompt.j2`
already asks the orchestrator to recommend installed skill references. Adding
each new strategy to the always-loaded domain prompt instead spends context on
every round of every campaign regardless of relevance, and gives strategies no
room for the evidence requirements and failure modes that make them usable.

The `llm-serving` domain already resolved this: `resources/skills/serving-systems/`
is a routed skill with a topic library, scoped by a `.vibesys.toml` sidecar. The
`microservices` domain has no skill at all.

### Desired outcome

Microservice optimization strategy is expressed as routed Agent Skills that an
agent invokes when the evidence calls for them, rather than as always-loaded
prompt text. Each skill states which role reads which evidence, what the
strategy licenses concretely, and how a change is verified against the
benchmark's `primary_value` rather than against diagnostic telemetry.

The always-loaded microservices domain prompt keeps the low-altitude task shapes
it has today. Skills carry the strategies that need evidence, preconditions, and
failure modes to be usable.

### Acceptance criteria

- [ ] At least one microservices-scoped skill exists under `resources/skills/`
      with a `.vibesys.toml` rule limiting it to `domains = ["microservices"]`.
- [ ] Skill descriptions trigger on strategy-specific evidence and do not
      overlap the task shapes already in `domains/microservices/orchestrator.md`.
[...]

Issue #44, opened 2026-06-08
Online application specialization


vic-lsh:
## Purpose

Track VibeServe expansion scenarios where the target is an online application with externally visible APIs and stateful behavior.

## Core Rule

The implementation may replace the internal architecture, programming languages, service decomposition, storage systems, caches, RPC mechanisms, deployment topology, and internal APIs. The externally visible API behavior and correctness contract are fixed.

## Current Applications

- Social network: #46
- Hotel reservation: #47

## Scope

This issue is the top-level planning container for online application specialization. Application-specific child issues should define shared app semantics. Scenario issues should describe workload and semantic requirements without prescribing optimization strategies or internal architecture.

PR #381, opened 2026-08-19
Add a microservice request-graph optimization skill

Ryclic:
## Problem

Microservice optimization strategy today is always-loaded prompt text, and all
of it is parameter tuning inside a fixed service topology.
`src/vibesys/prompts/domains/microservices/orchestrator.md` lists seven task
shapes: connection pooling, gateway upstream tuning, cache usage, database
indexes, serialization, worker counts, and baseline establishment.

Strategies that change the shape of the request graph have nowhere to live, even
though #44's core rule explicitly permits them ("The implementation may replace
the internal architecture, programming languages, service decomposition, storage
systems, caches, RPC mechanisms, deployment topology, and internal APIs") and the
evidence needed to choose between them has been available since #318 wired
critical-path attribution through the OTel profiler MCP boundary.

`README.md` states the architecture rule that new optimization techniques are
added by writing a skill rather than by modifying the framework, and
`orchestrator_plan_prompt.j2` already asks the orchestrator to recommend
installed skill references. The `llm-serving` domain has `serving-systems`; the
`microservices` domain has no skill at all.

Part of #380. Parent area #44.

## Solution

One microservices-scoped skill, `resources/skills/microservice-optimization/`,
carrying two request-graph strategies:

1. **Change when work happens.** Parallelize independent sequential calls, or
   move work off the synchronous request path.
2. **Change how many cross-service calls happen.** A four-step ladder ordered by
   reversibility: batch, co-locate, merge processes, merge storage.

The split is deliberate: strategy 1 changes call *timing*, strategy 2 changes
call *count*. Batching sits in strategy 2 rather than strategy 1 because it
reduces edge count, which is the same axis as fusion.

The seven existing task shapes are untouched. They are a different altitude
(tuning within a topology) and the skill says so explicitly rather than
duplicating them.

### Architecture

The skill routes through the existing sidecar mechanism, not through new
framework code. One rule added to `resources/skills/.vibesys.toml`:

```toml
[[rule]]
path = "microservice-optimization"
domains = ["microservices"]
```

The load-bearing design point is that the two roles do not see the same
evidence, so the skill states requirements per role:

```mermaid
flowchart LR
    SB["servicebench trace"] -->|"trace-graph.json"| MCP["otel profiler MCP"]
    MCP -->|"trace_graphs()<br/>critical_path()"| PA["Profiler agent"]
    PA -->|"ProfilerSummary<br/>(prose)"| ORCH["Orchestrator<br/>selects strategy"]
    MCP -->|"trace_graphs()<br/>critical_path()"| IMPL["Implementer<br/>confirms the edge"]
    ORCH -->|"plan"| IMPL
```

The orchestrator only ever sees `ProfilerSummary` prose, so the skill tells it
to select a strategy only when the summary names a specific bottleneck, and to
plan a profiling round otherwise. The implementer can call the MCP tools, so the
skill tells it to confirm the edge against `nodes_by_contribution` and the
representative segments before editing.

Two evidence hazards are called out because both produce confident wrong work:

- Overlapping sibling calls are not additive, so parallelizing two calls that
  already overlap gains nothing. Flat span aggregates cannot show this;
  representative segments can.
- Critical-path scope is `synchronous_request`, so work moved off that path
  leaves the critical path by construction. A before/after critical-path
  comparison cannot on its own show a gain.

No guardrail text against collapsing the whole application into one process.
That is a deliberate decision: #44 permits it, and constraining it in a skill
would contradict the area's core rule. The ladder is ordered by reversibility so
the cheap step is the natural first move, but the ordering is guidance, not a
prohibition.

### Correctness properties

- The skill loads only for `DomainName.MICROSERVICES`. Confirmed by resolving
  `SkillMetadata` through `validate_skill_tree`, which reports
  `domains=(DomainName.MICROSERVICES,)` for this directory.
- Frontmatter `name` matches the materialized directory name, which
  `src/vibesys/skills.py` requires so an orchestrator recommendation resolves
  identically across providers.
- The `description` triggers on request-graph evidence only. It does not mention
  cache, index, connection pool, or worker count, so it will not fire on the
  task shapes already in the always-loaded domain prompt.
- The skill states that benchmark `primary_value` is the result and that
  trace-derived evidence is diagnostic, matching the existing boundary in
  `profilers/otel.j2` and `domains/microservices/profiler.md`.
- No framework, evaluator, profiler, or prompt behavior changes. The only
  non-additive edit is one new rule in an existing sidecar.

### Testing

```
.venv/bin/python -m pytest tests/test_skills_wiring.py -q
47 passed

.venv/bin/python -m pytest tests/test_resources_packaging.py tests/test_workspace.py tests/test_schemas.py -q
42 passed
```

Routing checked directly:

```
.venv/bin/python -c "from pathlib import Path; from vibesys.skills import validate_skill_tree; [print(m) for m in validate_skill_tree(Path('resources/skills'))]"
SkillMetadata(skill_dir=.../microservice-optimization, backends=None, domains=(<DomainName.MICROSERVICES: 'microservices'>,), ...)
```

Not run: an end-to-end microservice campaign showing an agent invoking the skill
and the resulting round. That needs agent credentials and a full DeathStarBench
Hotel Reservation run; it is tracked as the last acceptance criterion on #380.

### DIFF
--- a/resources/skills/.vibesys.toml
+++ b/resources/skills/.vibesys.toml
@@ -1,3 +1,7 @@
 [[rule]]
 path = "serving-systems"
 domains = ["llm-serving"]
+
+[[rule]]
+path = "microservice-optimization"
+domains = ["microservices"]
--- a/resources/skills/microservice-optimization/SKILL.md
+++ b/resources/skills/microservice-optimization/SKILL.md
@@ -1,0 +1,135 @@
+---
+name: microservice-optimization
+description: >-
+  Restructuring the request graph of a microservice application to cut
+  end-to-end latency. Activate on serialized cross-service calls,
+  critical-path contribution, sequential dependencies that could run in
+  parallel, work that could move off the request path, batching repeated calls
+  to one service, or collapsing a call edge by co-locating or merging two
+  services.
+---
+
+# microservice-optimization
+
+Two strategies for cutting end-to-end latency by changing the shape of the
+request graph, not by tuning parameters inside a fixed topology.
+
+| Strategy | Changes | Reach for it when |
+| --- | --- | --- |
+| 1. When work happens | Timing of calls relative to the request | The critical path is a chain of calls that need not be a chain |
+| 2. How many calls happen | Count of cross-service calls | The critical path is dominated by call count or per-call overhead |
+
+Tuning within the current topology (connection pools, gateway upstreams, cache
+usage, database indexes, worker counts, serialization) is covered by the
+orchestrator's standing task guidance and is not repeated here. Reach for this
+skill when the evidence points at the graph rather than at one service's
+configuration.
+
+## Establish the evidence first
+
+The two roles see different evidence. Neither should act on the other's.
+
+**Orchestrator.** You see the profiler's prose summary, not the trace graph.
+Select a strategy only when that summary names a specific bottleneck: a
+service, an operation, or an edge with critical-path contribution. A ranking of
+services by aggregate latency is not sufficient. Aggregate cost and
+critical-path contribution are different quantities, and the expensive service
+is frequently not the one holding the request open. When the summary names no
+edge, plan a profiling round rather than guessing one.
+
+**Implementer.** You can read the graph. Confirm the edge or the structure
+before editing:
+
+1. `trace_graphs()` to locate schema-v2 graphs.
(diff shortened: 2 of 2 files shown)


vic-lsh:
Looking forward to your findings! @Ryclic

PR #391, opened 2026-08-20
Add experiment-chat as a permanent pane with the hypothesis log for the TUI

AyanBinRafaih:
## Problem

The client opens on the experiment log, and the chat is wired to a coding agent scoped to the run (#330), but the two cannot be on screen at once. `chatOpen` makes `experimentLogVisible` false, so `/chat`, `/chat <question>`, and any plain text typed into the input all replace the table with a modal. The operator asks "what changed in r41?" and the rows the question is about disappear behind the answer.

That is backwards for the questions this chat is for. "Summarize the last three rounds", "why was H-08 rejected", "what is the implementer doing now" are all questions about what the table is showing. Answering them should not cost the operator the thing they were reading, and closing the answer to look at the table again should not be part of the loop.

#328 gave the client a real layout: visualization output renders in a right pane beside the view instead of over it, and the chat, while open, is confined to the left pane rather than floating across the middle. The layout is there. The chat is still a dialog.

Worse, opening it costs the operator the view they asked from. `experimentLogVisible` required `!chatOpen`, so the moment a question opened the chat the landing view was replaced by the per-round transcript: the rounds strip, the agent strip, and unfiltered log output, none of which the operator asked for, behind a modal answering a question about the table that just disappeared.

## Solution

Demo:

https://github.com/user-attachments/assets/d9ceffed-654d-4804-8837-94277d63c30e

**The chat becomes a column of the landing view rather than a dialog over it.** `chatDocked(state)` is true when the experiment log is the view and the terminal can carry the column. Docked, plain text and `/chat <question>` no longer set `chatOpen` at all: the question and its answer land in the pane and the table stays exactly where it was. Bare `/chat` puts the pane keys on the chat instead of opening a modal over the view it is already part of. Inside a hypothesis, and in a terminal too narrow to dock, the chat is the modal it has always been, so nothing loses a surface it used to have.

**Each column gets the input that writes to it.** The chat has its own box at the foot of its column and the client's command input starts at the boundary between the two columns rather than running the width of the screen under both. The command list rises out of the command input rather than across the chat, because it completes that box and not the other one. `Ctrl+W` moves the cursor along with the pane keys, and the focused box carries the focus border. The chat's instruction line carries the state it needs to: `Ctrl+W to type here` while the cursor is elsewhere, `Ask about this run` when it holds it, and `Awaiting the agent` while a question is in flight. A question typed into either box reaches the same agent and is answered in the same column, so the split is a division of attention rather than of capability.

**`/chat` leaves the command surface where the chat is already a pane.** On the landing view it is absent from `/help` and from the completions: there is nothing for it to open. It is still accepted, and still focuses the pane, because hiding a command is not the same as removing it. Inside a hypothesis the chat is a dialog again, the command comes back, and it opens over the trajectory. The conversation is one conversation throughout, so a question asked from the pop-up is in the column when the operator steps back out.

**Focus becomes three regions instead of two.** `PaneFocus` is `'chat' | 'left' | 'right'`, and `Ctrl+W` cycles one column to the right through whichever columns are actually on screen, wrapping. Page Up and Page Down scroll the focused pane; Escape on the chat hands the keys back to the table, while Escape on the right pane still closes it. The table keeps its own keys throughout: arrows move the selection and Enter opens a hypothesis whether the chat is docked or not. `togglePaneFocus` is renamed `cyclePaneFocus` because with three regions "toggle" is no longer what it does, and `focusPane` is added for the one case that names its destination.

**The chat never replaces the view it was asked from.** `experimentLogVisible` no longer consults `chatOpen`. Docked, the chat is part of the landing view; as a modal it floats over it. Either way the table stays behind it, and a question can no longer drop the operator into a hypothesis they never opened.

**The widths are derived from what the table needs, not guessed.** `experiment-log.ts` now exports the panel widths at which it keeps its claim column and its compact column set, and `chat-pane.ts` sizes the dock from those. The chat asks for the columns left over once the table has its claim, so nothing is taken from a wide table; where there is no surplus it takes its readable minimum of 25 and the table drops columns as it would on any narrow terminal:

| Terminal | Landing view | Why |
| --- | --- | --- |
| < 92 columns | table only, chat as a modal over it | Two columns would both be unreadable |
| 92 to 119 columns | chat + table | The chat takes 25; the table drops columns as it narrows |
| ≥ 120 columns | chat + table, table whole | The surplus over the table's claim width is the chat's, up to 52 |
| ≥ ~180 columns with `/perf` | chat + table + chart | All three keep a readable floor |
| `/perf` on a narrower terminal | table + chart | The chat steps aside; `/chat` still opens it |
| < 100 columns with `/perf` | table, chart as a modal | Unchanged from #328 |

Opening a visualization therefore narrows the chat before it narrows the table, and where all three will not fit the chat is the one that yields, because the chart was asked for explicitly and the table is the view.

**The measurement is reported in rather than guessed at.** Only the renderer knows how many columns there are, so `app.ts` measures and calls `controller.setChatDockFits(fits)`. Where a question's answer is displayed has to agree with what is on screen: if the chat cannot dock, a typed question must open the modal rather than writing the answer into a column that is not being drawn. The layout in that same render pass uses the measurement directly rather than waiting for the state to come back, so a resize never draws a stale row.

Two columns at 140, each with its own input:

```
╭─ Experiment chat ─────────────────────────╮╭─ Experiments ───────────────────────────────────────────────────────────────────────────────╮
│  Ask about this run: progress, a          ││  Hypothesis     Rounds   Implementation Details          Measured    Verdict  Outcome       │
│  failure, or what a hypothesis changed.   ││  H-07           41       Batch the prefill step          —           Pass     Proven        │
│                                           ││                                                                                             │
│                                           ││ 1/1 · ↑↓ or scroll: select · Enter or /open-round: open its rounds                          │
╰───────────────────────────────────────────╯╰─────────────────────────────────────────────────────────────────────────────────────────────╯
Ctrl+W to type here                          ↑↓: select · Enter: open its rounds · Ctrl+W: chat and back · /open-round --N
╭─ Chat ────────────────────────────────────╮╭─ Ask or command ────────────────────────────────────────────────────────────────────────────╮
│ Type a question                           ││ Type a question or /help                                                                    │
╰───────────────────────────────────────────╯╰─────────────────────────────────────────────────────────────────────────────────────────────╯
```

Three at 200, after `/perf`. The command input still starts on the same boundary, now spanning the table and the chart:

```
╭─ Experiment chat ─────╮╭─ Experiments ───────────────────────────────────────────────────────────────────────────╮╭─ Performance · focused ──────────────────────────────────────────────────────────╮
│                       ││  Hypothesis     Rounds   Measured    Verdict  Outcome                                   ││ Performance · tok_s                                                              │
│  ╭─────────────────╮  ││  H-07           41       —           Pass     Proven                                    ││     1135 ┤   ●                                                                   │
│  │ Answer          │  ││                                                                                         ││ best r7 1135 tok_s                                                               │
│  │ Prefill.        │  ││                                                                                         ││                                                                                  │
│  ╰─────────────────╯  ││                                                                                         ││                                                                                  │
│                       ││                                                                                         ││                                                                                  │
│                       ││ 1/1 · ↑↓ or scroll: select · Enter or /open-round: open its rounds                      ││                                                                                  │
╰───────────────────────╯╰─────────────────────────────────────────────────────────────────────────────────────────╯╰──────────────────────────────────────────────────────────────────────────────────╯
Ctrl+W to type here      Ctrl+W: switch pane · PgUp/PgDn: scroll focused pane · Esc on the pane: close it
╭─ Chat ────────────────╮╭─ Ask or command ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Type a question       ││ Type a question or /help                                                                                                                                                    │
╰───────────────────────╯╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

### Architecture

```mermaid
flowchart TD
    INPUT["client input"] --> PARSE[parseInput]
    PARSE -->|plain text| SEND["sendChat"]
    PARSE -->|/chat| OPEN["openChat"]
    PARSE -->|/perf · /history rounds| PANE["openPane"]
    DOCKED{"chatDocked(state)<br/>log is the view · terminal fits"}
    SEND --> DOCKED
    OPEN --> DOCKED
    DOCKED -->|yes| COL["answer lands in the chat column<br/>chatOpen stays false"]
    DOCKED -->|no| MODAL["chat opens over the view<br/>as before"]
    RENDER["app.ts render"] -->|measures terminal| FITS["controller.setChatDockFits"]
    FITS --> DOCKED
    COL --> ROW["main row: chat · log · right pane"]
    PANE --> ROW
    CTRLW["Ctrl+W"] --> CYCLE["cyclePaneFocus over the columns on screen"]
    CYCLE --> ROW
```

The chat transcript itself is untouched: one `chatConversation`, one agent, one queue. Docking changes where it is drawn and whether a modal opens, not what a question does or what comes back.

### Correctness properties

- On the landing view with room to dock, a typed question leaves `chatOpen` false, keeps `experimentLogVisible` true, and puts both the question and the answer in the chat column.
- Where the chat cannot dock, in a terminal too narrow or inside a hypothesis, a question opens the modal, so an answer is never written to a column that is not on screen.
- Asking from the landing view never changes the view. With the chat docked the table does not move; with the chat modal the table stays behind it, and the rounds strip, agent strip, and per-round transcript stay out of a view the operator did not open.
- Bare `/chat` while docked focuses the pane and opens nothing; `/chat <question>` sends the question and still opens nothing.
- The two inputs never overlap: the command input starts at exactly the column where the chat pane ends, and the command list is confined to the column it completes.
- Exactly one input holds the cursor. `Ctrl+W` moves it with the pane focus, the modal takes it back while it is open, and Enter goes to whichever box is focused rather than always to the command input.
- Text typed into the chat input reaches the agent through `submitChat`, so a slash command typed there still runs as a command.
- Enter opens the selected hypothesis only when both boxes are empty, so a question waiting in either one is never discarded by a keystroke meant to send it.
- `/chat` is absent from `/help` and from the completions on the landing view, present in both inside a hypothesis, and accepted either way.
- The transcript is continuous across surfaces: a question asked from the pop-up inside a hypothesis, question and answer both, is in the docked column after `Ctrl+L` returns to the table.
- The chat transcript is one conversation across all of it: docking, undocking, opening and closing a visualization, and stepping in and out of a hypothesis leave it intact.
- Docking never costs the table its claim column. For every terminal width at or above the dock threshold, the table keeps at least the width it needs for the claim; with a visualization open it keeps at least its compact column set.
- The chat column is never narrower than 28 columns or wider than 52, and it yields columns to a visualization before the table does.
- The table keeps its own keys while docked: arrows move the selection, Enter opens the selected hypothesis, and Page Up and Page Down move the selection unless the chat holds focus.
- `Ctrl+W` only ever lands on a column that is drawn, and focus falls back to the table when the dock stops fitting.
- Escape on the chat returns the keys to the table without closing anything; Escape on the right pane still closes the pane.
- `/help`, `/theme`, and errors are still modal, and the visualization pane's own fallback below 100 columns is unchanged.
- No protocol change: no request, response field, or event was added, removed, or reinterpreted, and the Python side is untouched.

### Testing

- `bun test --max-concurrency 1 src` in `clients/tui`: 269 pass, 1 fail, 270 tests across 15 files. The failure is in `launcher.test.ts`, fails identically on unmodified `origin/main` in a second worktree, and is environment-dependent: it shells out to the launcher.
- `pnpm --dir clients/tui check` (`tsc -p tsconfig.check.json`): clean.
- `pnpm --dir clients/tui build`: clean.
- `pnpm check:ts` (`biome check .`): clean across all 44 files.
- `uv run pytest tests/test_tui.py`: 32 passed. No Python file changed; this is the regression check for the protocol the client talks to.

New and changed tests:

| File | Test | Covers |
| --- | --- | --- |
| `ui/chat-pane.test.ts` | `docks only once the table can keep its claim column` | The dock threshold, including that 120 columns still belongs to the table alone |
| `ui/chat-pane.test.ts` | `docks beside a visualization out of the compact table width` | Three columns fit at 200; at 140 with a chart they do not |
| `ui/chat-pane.test.ts` | `never takes the table below the width it needs` | Every width from the threshold to 400, with and without a chart |
| `ui/chat-pane.test.ts` | `narrows itself rather than the table when a visualization opens` | Which surface yields columns first |
| `session-model.test.ts` | `docks beside the log and nowhere else` | Docked on the landing view, not inside a hypothesis |
| `session-model.test.ts` | `gives the pane the keys instead of opening a modal over the table` | `/chat` while docked |
| `session-model.test.ts` | `still opens the modal where the chat cannot dock` | The narrow-terminal fallback |
| `session-model.test.ts` | `cycles focus left to right through the columns on screen` | Three-region `Ctrl+W`, and focus after the pane closes |
| `session-model.test.ts` | `refuses focus for a column that is not on screen` | `focusPane` on a hidden column |
| `session-model.test.ts` | `hands the keys back to the log when the dock stops fitting` | Resize while the chat holds focus |
| `session-controller.test.ts` | `sends ordinary text to the chat docked beside the log` | The table survives a question |
| `session-controller.test.ts` | `opens the chat as a modal where it cannot dock` | Narrow terminal, reported by the renderer |
| `session-controller.test.ts` | `keeps the log as the view when the chat opens over it` | The regression: a question must not swap in the per-round transcript |
| `ui/app.test.ts` | `keeps the table behind the chat modal instead of the round transcript` | The same, in a rendered frame at 84 columns |
| `session-controller.test.ts` | `opens the chat as a modal inside a hypothesis`| Modal there, docked again after `Ctrl+L`, transcript intact |
| `session-controller.test.ts` | `keeps the docked chat beside the log while a visualization is open` | `/perf` plus a question, all three live |
| `ui/app.test.ts` | `lands with the chat docked beside the hypothesis table` | The rendered landing frame, claim column intact, and the command input starting exactly where the chat column ends |
| `ui/app.test.ts` | `routes typing to whichever input Ctrl+W points at` | A question typed after `Ctrl+W` reaches `submitChat`; a command typed after `Ctrl+W` back reaches `submit` |
| `ui/app.test.ts` | `raises the command list out of the command input, clear of the chat` | Suggestion box aligned to the command column |
| `ui/app.test.ts` | `says when the docked chat is waiting on the agent` | The instruction line as the pending indicator |
| `ui/app.test.ts` | `drops /chat from the command surface while the chat is already docked` | Typing `/c` on the landing view completes nothing |
| `commands.test.ts` | `drops /chat where the chat is already a pane of the view` | The list, the help text, and the completions |
| `commands.test.ts` | `still accepts /chat when it is not offered, since the chat is the point` | Hidden, not removed |
| `session-controller.test.ts` | `offers /chat in help only where the chat is not already on screen` | `/help` on the landing view and inside a hypothesis |
| `session-controller.test.ts` | `carries the modal conversation back into the docked pane` | Ask from the pop-up, step back out, find it in the column |
| `ui/app.test.ts` | `answers in the docked chat without covering the table` | A typed question and its answer beside the rows |
| `ui/app.test.ts` | `keeps the chat, the table, and the visualization on screen together` | The three-column frame |
| `ui/app.test.ts` | `gives the row to the table alone when the chat cannot fit beside it` | The narrow frame, and that the measurement reached the model |
| `ui/app.test.ts` | `moves the pane keys onto the docked chat and back with Ctrl+W` | Focus border and title in the frame |
| `ui/app.test.ts` | `leaves the table its own keys while the chat is docked` | Arrow keys still move the selection |

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -24,7 +24,7 @@ available slash commands are:
 | Command | Behavior |
 | --- | --- |
 | `/help` | Show commands and planned controls. |
-| `/chat` | Open a read-only chat about the run; `/chat <question>` asks immediately. |
+| `/chat` | Put the pane keys on the docked chat, or open it as a modal where it cannot dock; `/chat <question>` asks immediately. |
 | | Slash commands work inside the chat too, and do the same thing as in the main input. |
 | `/pause` | Pause after the current agent call finishes. |
 | `/resume` | Resume a paused run. |
@@ -80,18 +80,57 @@ reshuffle. Records written before hypothesis tracking render as
 `(Unidentified)` rather than being dropped. Columns drop widest first as the
 terminal narrows; hypothesis, rounds, and outcome always survive.
 
+The chat is docked beside the table as a permanent column of this view, so a
+question about the run never covers what the run has done. It has its own input
+at the foot of its column, under the instruction `Ask about this run`,
+and the client's command input sits beside it, starting at the boundary between
+the two columns so each box is under the surface it writes to. The command
+list rises out of the command input rather than across the chat.
+
+The cursor starts in the command input, and `Ctrl+W` moves both it and the pane
+keys to the chat and back; the chat's instruction line says so (`Ctrl+W to type
+here`) until it holds them, and the focused input carries the focus border, so
+where a keystroke lands is never a guess. A question typed into either box
+reaches the same agent and is answered in the same column. Page Up and Page Down scroll the focused pane, and
+Escape hands the keys back to the table. Arrows, Enter, and the rest of the
+table's keys are unaffected by the dock. A slash command typed into the chat
+input runs through the same path as anywhere else, and ordinary text typed into
+the command input is still a question for the agent: the two boxes are a
+division of attention, not of capability.
+
+`/chat` leaves the command surface on this view, since the chat is already
+beside the table: it is absent from `/help` and from the completions, though it
+is still accepted and still focuses the pane. Inside a hypothesis, where the
+chat is a dialog again, the command comes back and opens it over the
+trajectory. It is one conversation throughout, so a question asked from the
+pop-up is in the column when the operator steps back out.
+
+The chat asks for the columns left over once the table has enough for its
+claim, so a wide terminal loses no table column to it; where there is no such
+surplus it still takes its readable minimum and the table drops columns as it
+does on any narrow terminal. Below about 92 columns two columns would both be
+unreadable, so the table keeps the row alone and the chat opens over it as a
+modal. Opening a visualization narrows the chat before it narrows the table,
+and where all three will not fit the chat is the one that steps aside.
+
+Whichever way the chat is on screen, it never replaces the view it was asked
+from: the modal floats over the table rather than swapping in the per-round
+transcript behind it.
+
 ### Split panes
 
 Visualization commands render beside the current view rather than over it.
 `/perf` puts its output in a right pane and leaves the transcript, the chat,
-or the experiment log in the left one, both live at the same time. A second
-visualization command replaces the pane's contents rather than stacking
-another surface on top.
+or the experiment log in the left one, both live at the same time. On the
+experiment log that makes three columns, chat, table, and visualization, each
+live. A second visualization command replaces the pane's contents rather than
+stacking another surface on top.
 
-`Ctrl+W` moves focus between the panes; the focused one carries the theme's
-focus border and says so in its title. Page Up and Page Down scroll whichever
-pane has focus, and Escape on the right pane closes it and restores the
-full-width view. Chat and transcript state survive the pane closing.
+`Ctrl+W` moves focus one column to the right and wraps, through whichever
+columns are actually on screen; the focused one carries the theme's focus
+border and says so in its title. Page Up and Page Down scroll whichever pane
+has focus, and Escape on the right pane closes it and restores the full-width
+view. Chat and transcript state survive the pane closing.
 
 Pane widths are computed from the terminal, so a wide terminal gives the
 visualization real room while the left pane keeps a readable floor. Below 100
@@ -115,6 +154,12 @@ Text typed in the chat that starts with `/` is parsed as a slash command
 through the same path as the main input, so `/perf` there does exactly what
 `/perf` does anywhere else. Anything else is a question for the agent, and a
 question containing a slash mid-sentence is still a question.
+
+Where the chat is docked it answers in its column, under its own input; where
+it is not, inside a hypothesis or in a terminal too narrow to dock, it opens
+over the view as before, carrying the same input at the foot of the modal. It
+is one conversation either way: the transcript survives docking, undocking, and
+the pane closing.
 
 Inside a hypothesis the footer shows keyboard navigation. `[` and `]` select rounds, Tab and
 Shift+Tab select agents, Page Up/Page Down scroll the transcript, Ctrl+T expands
--- a/clients/tui/src/commands.test.ts
+++ b/clients/tui/src/commands.test.ts
@@ -1,5 +1,11 @@
 import {describe, expect, it} from 'bun:test';
-import {parseInput, slashCommandRange, suggestSlashCommands} from './commands.js';
+import {
+  availableCommands,
+  helpText,
+  parseInput,
+  slashCommandRange,
+  suggestSlashCommands,
+} from './commands.js';
 
 describe('parseInput', () => {
   it('accepts the intentionally small slash-command surface', () => {
@@ -82,6 +88,28 @@ describe('parseInput', () => {
     expect(parsed.error).toContain('Unknown theme: monokai');
     expect(parsed.error).toContain('catppuccin-mocha');
     expect(parsed.localView).toBeUndefined();
+  });
+});
+
+describe('command surface by view', () => {
+  it('drops /chat where the chat is already a pane of the view', () => {
+    const docked = availableCommands({chatDocked: true}).map(command => command.name);
+
+    expect(docked).not.toContain('/chat');
+    expect(docked).toContain('/perf');
+    expect(helpText({chatDocked: true})).not.toContain('/chat');
+    expect(suggestSlashCommands('/c', {chatDocked: true})).toEqual([]);
+  });
+
+  it('offers /chat everywhere the chat is not already on screen', () => {
+    expect(availableCommands().map(command => command.name)).toContain('/chat');
+    expect(helpText()).toContain('/chat');
+    expect(suggestSlashCommands('/c').map(command => command.name)).toEqual(['/chat']);
+  });
+
+  it('still accepts /chat when it is not offered, since the chat is the point', () => {
+    // Hidden from the list, not removed from the client.
+    expect(parseInput('/chat')).toEqual({localView: 'chat'});
   });
 });
 
--- a/clients/tui/src/commands.ts
+++ b/clients/tui/src/commands.ts
@@ -38,18 +38,40 @@ export const SLASH_COMMANDS: readonly SlashCommand[] = [
   {name: '/theme', description: 'List themes, or switch with /theme <name>'},
 ];
 
-export const HELP_TEXT = [
-  'Available',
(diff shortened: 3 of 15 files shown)


vic-lsh:
Outdated as #395 is already merged.

PR #397, opened 2026-08-23
Fix local OTel profiling for microservice tasks

Ryclic:
## Problem

`--profiler otel` did not work on a local Linux run of a `microservices` task.
A user cloning the repo and running the `hotel-reservation` example got a loop
that planned from source reading alone, with no trace evidence, and no error
explaining why.

Three separate defects produced that outcome:

1. **`auto` resolved to `none`.** The microservices domain defaults `auto` to
   `NONE`, so a task whose benchmark already provisions a collector and
   instrumentation still ran unprofiled unless the user knew to pass
   `--profiler otel` explicitly. Nothing in the CLI surfaced that the task was
   capable of more.
2. **The Profiler could not reach the workload.** A microservices candidate is
   a Docker Compose topology, not a process the agent starts itself. Under the
   default bubblewrap confinement the agent has no Docker socket, and the
   sandbox's private `/tmp` masks the capture directory the benchmark writes
   to. Docker resolves bind-mount sources in the daemon's namespace rather than
   the agent's, so a capture path that does not name the same directory inside
   and outside confinement silently resolves to the wrong place.
3. **The Profiler could not run the benchmark.** The packaged benchmark command
   names `${PACKAGE_ROOT}`, and the Profiler is instructed to run it when no
   capture exists yet. That package lives outside the workspace and was not
   imported, so on a cold start the command died on a missing directory and the
   role returned no evidence at all.

The result was that the OTel path only appeared to work when a capture had been
left behind by some earlier out-of-band run.

## Solution

Fix each defect at its own boundary; no new profiler machinery.

- **Capability-based default.** `InputBundle.provisions_trace_telemetry` reports
  whether a resolved benchmark command names both `--telemetry-output` and
  `--trace-graph-json`. At the CLI boundary, `_apply_bundle_profiler_default`
  upgrades `ProfilerKind.AUTO` to `OTEL` for a microservices bundle with that
  capability. An explicit `--profiler` always wins; tasks without those flags
  stay on `none`, since OTel profiling would have nothing to read.
- **Scoped host-resource widening.** `container_runtime_resources()` and
  `task_scratch_dir()` declare the Docker control socket and
  `/tmp/vibesys-<task>`, bound back through the tmpfs mask at an identical path
  so bind-mount sources resolve identically in both namespaces. This is gated on
  the microservices domain and on host runs only: container backends run the
  agent inside their own image and own resource exposure themselves.
- **Evaluator package import.** The evaluator package root is imported
  read-only so a packaged benchmark command resolves. Read-only because the
  evaluator is trusted, integrity-checked input that no role may edit.
- **Prompt correction.** `otel.j2` carried a stale caveat claiming `/tmp`
  artifacts are invisible to the agent. That is no longer true, and it steered
  the Profiler away from reusing valid captures. Replaced with guidance to reuse
  an existing artifact unless it is missing or older than the edit being
  attributed.

Reviewers should look closely at the gating in `context.py`: reaching the Docker
socket is root-equivalent on the host, so the widening must not leak outside the
microservices domain or into container-backed runs.

### Architecture

`_assemble_run_context` computes the declarations once and hands them to
`build_agent_client`. `_RunContext` retains them so `_assemble_candidate_context`
(the `--loop evolve` path) can give a candidate the same access: a candidate
runs its parent's domain, but its context carries neither the profiler domain
nor the task name, so recomputing there is impossible.

```mermaid
flowchart TD
    A[main: bundle load] -->|provisions_trace_telemetry| B[AUTO becomes OTEL]
    B --> C[_assemble_run_context]
    C -->|domain is MICROSERVICES and not cli_sandboxed| D[agent_host_resources]
    D -->|docker.sock RW, scratch RW, evaluator pkg RO| E[build_agent_client]
    C -->|retained on _RunContext| F[_assemble_candidate_context]
    F -->|parent.agent_host_resources| G[build_agent_client for candidate]
    E --> H[Profiler role]
    H -->|servicebench trace| I[scratch otel dir]
    I --> J[OTel MCP: summary, critical_path, trace_graphs]
```

## Verification

Verified by an end-to-end agentic run from a genuinely cold state, not by code
inspection alone: submodule reset to baseline, `/tmp/vibesys-<task>` removed, no
containers running, no pre-staged capture.

The run reached `need_profile=true`, instantiated the Profiler, found no
telemetry, ran `servicebench trace` itself, captured 265,744 spans across 23,498
traces over 3x30s windows, and called the OTel MCP tools. The resulting plan
cited measured values (`/hotels` p95 26.12ms, profile fetch ~5ms sitting
sequentially after an independent call) rather than source reading. The accepted
change parallelized `profile.GetProfiles` with `reservation.CheckAvailability`
in the frontend; the framework accuracy gate passed and the benchmark reported
2848.17 ops/s. A second profile confirmed activation: `/hotels` p50 16.66ms to
14.87ms, p95 26.12ms to 23.70ms, with the profile span's mean critical-path
contribution dropping 1.33ms to 0.59ms.

Known gap: that run stopped after one completed round on a pre-existing
structured-output retry defect unrelated to this change (the orchestrator emits
its plan wrapped in an envelope key, fails schema validation, and exhausts
retries fatally). Filing separately.

### Correctness properties

- An explicit `--profiler` is never overridden by the capability default.
- Only a microservices bundle naming both `--telemetry-output` and
  `--trace-graph-json` upgrades `AUTO`; every other task keeps its current
  resolution.
- Host-resource widening is inert outside the microservices domain and inert
  when `cli_sandboxed` is set, so container-backed runs are unchanged.
- The evaluator package is imported read-only; no role can edit trusted,
  integrity-checked evaluator input.
- The scratch directory names the same absolute path inside and outside
  confinement, which is what makes a Docker bind-mount source resolve correctly.
- A candidate context receives exactly its parent's declarations, so `--loop
  evolve` and `--loop agent` expose the same surface.

### Testing

```
./scripts/check_format.sh          # All checks passed
./scripts/check_lint.sh            # All checks passed
uv run pytest                      # 4961 passed, 12 skipped, 3 failed
```

The 3 failures (`tests/test_launch_imports.py` x2,
`tests/test_tui.py::test_cli_parse_failure_is_streamed_after_client_attaches`)
reproduce identically on `origin/main` in a clean worktree and are unrelated to
this branch.

`ruff format --check` also flags `src/vibesys/agents/drivers/__init__.py`, which
is likewise pre-existing on `origin/main` and untouched here; left alone to keep
this PR scoped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,10 @@ requires `bubblewrap`, and a kernel that lets it create an unprivileged user
 namespace; macOS includes the required `sandbox-exec` command. Where user
 namespaces are blocked and installing bubblewrap needs root you do not have,
 `VIBESYS_AGENT_SANDBOX=landlock` selects a weaker but root-free backend (see
-the [CLI reference](docs/cli-flags.md) for what it stops enforcing). Then
-install VibeSys:
+the [CLI reference](docs/cli-flags.md) for what it stops enforcing). Tasks whose
+candidate is a container topology, such as the `microservices` examples, also
+need Docker Engine reachable without `sudo` (add your user to the `docker`
+group) and a Go toolchain on `PATH`. Then install VibeSys:
 
 ```bash
 uv tool install vibesys
--- a/docs/cli-flags.md
+++ b/docs/cli-flags.md
@@ -96,6 +96,16 @@ legacy root-level task inputs, or `_evaluator/`, and cannot read
 rejected when a root `.env*` file or `agent.toml` is recoverable from Git refs
 or reflogs.
 
+Local runs in the `microservices` domain import two extra host resources,
+because a microservice candidate is a container topology rather than a process
+the agent can start on its own: the Docker control socket, and the task scratch
+directory `/tmp/vibesys-<task>`. The scratch directory is shared with the host
+rather than masked by the sandbox's private `/tmp`, because Docker resolves a
+bind-mount source in the daemon's namespace, so a capture directory only works
+when the path names the same directory inside and outside confinement. Access
+to the Docker socket is equivalent to root on the host, so it is granted only
+for this domain; use `--docker` to confine the workload to a container instead.
+
 `VIBESYS_AGENT_SANDBOX` selects the Linux mechanism. `auto` (the default) and
 `bwrap` both require bubblewrap. `landlock` opts in to a weaker backend for
 hosts that block unprivileged user namespaces, which is the common reason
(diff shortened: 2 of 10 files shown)


Ryclic:
Superseded by #405: same branch, rebased onto current `main` (the `build_agent_runner` to `build_agent_client` refactor in #403), same three commits. GitHub would not reopen this one after the force-push.

PR #424, opened 2026-08-25
fix(tui): arrow-key navigation for slash suggestions and theme preview (#399, #400)

### LINKED ISSUE #399: TUI: navigate slash-command suggestions with arrow keys before Tab completion
## Observed behavior

When the command input contains a slash prefix with multiple matches, the suggestion panel always highlights the first match. Up and Down do not move that highlight. Instead, those keys continue to navigate the experiment log or round view behind the suggestion panel. Pressing Tab always completes the first match regardless of prior arrow input.

At commit `9069f03`, `clients/tui/src/ui/input.ts` stores only the matching commands, renders index 0 as selected, and completes `matches[0]`. There is no selected-suggestion state or arrow-navigation action. `clients/tui/src/ui/keybindings.ts` routes Up and Down to the underlying view before handling Tab completion.

## Expected behavior

- [ ] While slash-command suggestions are visible, Up and Down move the highlighted suggestion.
- [ ] Tab completes the highlighted suggestion.
- [ ] Suggestion-navigation keys do not change selections in the view behind the input.
- [ ] Selection remains valid when typing changes the filtered matches.
- [ ] Renderer tests cover multiple matches, both arrow directions, and Tab completion of a non-first match.

## Reproduction

1. Launch the interactive TUI.
2. Type `/` in the command input so multiple slash-command suggestions appear.
3. Press Down one or more times.
4. Observe that the suggestion marker does not move and the underlying view may change selection.
5. Press Tab.
6. Observe that the input completes to the first visible match, `/help`, rather than the item targeted with the arrow keys.

## Affected subsystem

CLI and packaging

## Environment

- Commit: `9069f03e434f5918549035e2a636549627606947`
- OS: macOS 27.0, arm64
- Bun: 1.3.9
- Node.js: 24.7.0
- pnpm: 11.11.0
- Hardware, backend, and profiler: not applicable

## Impact

Workaround exists. Users can keep typing until only the intended command matches, but they cannot operate the visible suggestion list as a keyboard selection interface. Arrow input can also mutate the view behind the suggestions.

## Relevant logs

```shell
# Type "/" so several commands are visible.
# Press Down, then Tab.
# The marker remains on the first row and Tab completes "/help".
```

Focused validation:

```shell
bun test --max-concurrency 1 src/ui/app.test.ts src/session-model.test.ts
# 116 pass, 0 fail
```

The existing renderer test verifies only prefix-to-first-match Tab completion, which passes and confirms the current behavior.

Parent roadmap: #284. Slash autocomplete was introduced by #188. The theme picker navigation work in #331 and #337 provides a related keyboard-selection interaction but does not cover slash suggestions.

## Pre-submission checks

- [x] I searched open and closed issues and merged pull requests for an existing report.

### LINKED ISSUE #400: TUI: preview highlighted themes before applying the selection
## Observed behavior

In the `/theme` picker, Up and Down move the highlight but the rendered TUI remains on the active theme. The user cannot see how the highlighted theme looks until pressing Enter, which applies it and closes the picker.

At commit `9069f03`, `moveThemeSelection` updates only `themePicker.selected`; `themeName` changes only when Enter calls `applySelectedTheme`. The app repaints only when `themeName` changes. Existing tests and documentation explicitly cover this apply-on-Enter behavior, so this is a missing preview interaction rather than an intermittent repaint failure.

## Expected behavior

- [ ] Moving the picker highlight immediately previews that theme across the rendered TUI.
- [ ] Enter commits the highlighted theme and closes the picker.
- [ ] Escape closes the picker and restores the theme that was active when the picker opened.
- [ ] Moving repeatedly through the list previews each newly highlighted theme without persisting intermediate choices.
- [ ] Direct `/theme <name>` selection continues to apply immediately.
- [ ] Renderer tests verify preview, commit, and cancellation for both light and dark themes.

## Reproduction

1. Launch the interactive TUI using the default dark theme.
2. Enter `/theme`.
3. Press Down to highlight `light`.
4. Observe that the picker highlight moves to `light`, but the header, canvas, transcript, and input remain rendered with the dark palette.
5. Press Enter and observe that the light palette appears only after the picker closes.

## Affected subsystem

CLI and packaging

## Environment

- Commit: `9069f03e434f5918549035e2a636549627606947`
- OS: macOS 27.0, arm64
- Bun: 1.3.9
- Node.js: 24.7.0
- pnpm: 11.11.0
- Hardware, backend, and profiler: not applicable

## Impact

Workaround exists. Users can apply a theme, reopen the picker, and choose another, but they cannot compare themes before committing a choice.

## Relevant logs

```shell
/theme
# Press Down from "dark" to highlight "light".
# The picker marker moves, but the TUI keeps the dark palette until Enter.
```

Focused validation:

```shell
bun test --max-concurrency 1 src/ui/app.test.ts src/session-model.test.ts
# 116 pass, 0 fail
```

The passing tests include the current contract: moving the selection leaves `themeName` unchanged, and Enter applies the highlighted theme.

Parent roadmap: #284. The theme system was introduced by #260 and #265. Keyboard navigation was added by #331 and #337, which intentionally separated the highlighted row from the active theme but did not provide live preview.

## Pre-submission checks

- [x] I searched open and closed issues and merged pull requests for an existing report.

zatchbell1311-wq:
## Problem

Two keyboard interaction bugs in the TUI made the slash-command suggestion panel and the theme picker feel broken:

**#399 — Slash-command suggestions**: When typing `/` in the command bar, a suggestion panel appears but Up/Down arrow keys had no effect on it. The highlight stayed fixed on the first item, and those keystrokes fell through to the experiment log behind the panel instead. Tab always completed `/help` regardless of which item the user intended.

**#400 — Theme picker preview**: In the `/theme` picker, moving the highlight with Up/Down did not change the visible theme. The user had to press Enter blindly to see how a theme looked, then reopen the picker to try another. There was no live preview.

Closes #399
Closes #400

### #399
Added `selectedIndex` state to `input.ts` to track which suggestion is highlighted. The suggestion list re-renders on every move using a shared `renderSuggestions()` helper. `completeSuggestion()` now completes `matches[selectedIndex]` instead of always `matches[0]`. Two new methods — `moveSuggestion(delta)` and `suggestionsVisible()` — are exposed on the `InputPanel` interface. In `keybindings.ts`, the `experimentLogVisible` block now checks `suggestionsVisible()` before routing Up/Down: if the panel is open, arrow keys move the suggestion selection instead of the experiment log behind it.

### #400
In `app.ts`, the render function now reads `state.themePicker?.selected ?? state.themeName` as the theme to paint with. When the picker is open, the highlighted theme is used for the live render without committing it to state. `themeName` in state only changes on Enter (via `applySelectedTheme`). Escape closes the picker and the next render reverts to `state.themeName` automatically. No state-model changes were needed.

### Architecture

**#399 — Suggestion navigation flow:**
User presses Down
→ keybindings.ts: suggestionsVisible()? yes
→ actions.moveSuggestion(1)
→ input.ts: selectedIndex++ (clamped via modulo)
→ renderSuggestions() repaints the list with new highlight
→ Up/Down never reach experimentLog or moveExperimentSelection

**#400 — Theme preview flow:**
User presses Down in /theme picker
→ moveThemeSelection updates picker.selected (state.themeName unchanged)
→ render() fires
→ previewName = state.themePicker.selected (e.g. "light")
→ applyTheme("light") paints the whole TUI in light
→ Enter → applySelectedTheme() → state.themeName = "light" (committed)
→ Escape → closeThemePicker() → next render uses state.themeName (original)

### Correctness properties
- Arrow keys in the suggestion panel never reach the view behind it when suggestions are visible
- Tab completes whichever suggestion is highlighted, not always the first
- selectedIndex resets to 0 whenever the input value changes and matches are recomputed
- Theme preview is purely a render-time effect — state.themeName is never mutated during navigation
- Escape from the theme picker always restores the previously committed theme

### Testing
bun test --max-concurrency 1 src/ui/app.test.ts src/session-model.test.ts
Result: 79 pass, 0 fail, 256 expect() calls across both test files.
All pre-existing tests pass without modification.

### DIFF
--- a/clients/tui/src/session-model.ts
+++ b/clients/tui/src/session-model.ts
@@ -105,6 +105,7 @@ export interface LayoutState {
 
 export interface ThemePicker {
   selected: ThemeName;
+  original: ThemeName;
 }
 
 export interface UsageMeter {
@@ -392,7 +393,7 @@ export function setTheme(state: SessionState, themeName: ThemeName): SessionStat
 
 /** Opens the theme list as a selection, starting on the active theme. */
 export function openThemePicker(state: SessionState): SessionState {
-  return {...state, overlay: null, themePicker: {selected: state.themeName}};
+  return {...state, overlay: null, themePicker: {selected: state.themeName, original: state.themeName}};
 }
 
 /**
@@ -406,7 +407,7 @@ export function moveThemeSelection(state: SessionState, delta: number): SessionS
   const index = Math.min(THEME_NAMES.length - 1, Math.max(0, current + delta));
   const selected = THEME_NAMES[index];
   if (selected === undefined || selected === picker.selected) return state;
-  return {...state, themePicker: {selected}};
+  return {...state, themePicker: {...picker, selected}};
 }
 
 /** Closes the picker, leaving the theme as it was when it opened. */
--- a/clients/tui/src/ui/app.ts
+++ b/clients/tui/src/ui/app.ts
@@ -126,8 +126,9 @@ export function createOpenTuiApp(renderer: CliRenderer, controller: SessionContr
   let lastState: SessionState = controller.state;
   const render = (state: SessionState): void => {
     lastState = state;
+    const previewName = state.themePicker?.selected ?? state.themeName;
     const releasePreviousStyle =
-      state.themeName === themeName ? undefined : applyTheme(state.themeName);
+      previewName === themeName ? undefined : applyTheme(previewName);
     const showLog = experimentLogVisible(state);
     // A split only happens when the terminal can carry both panes. Narrower
     // than that, a visualization keeps the modal it had before the split
--- a/clients/tui/src/ui/app.ts
+++ b/clients/tui/src/ui/app.ts
@@ -205,6 +205,8 @@ export function createOpenTuiApp(renderer: CliRenderer, controller: SessionContr
   };
   const unbindKeys = bindKeybindings(renderer, controller, viewport, {
     completeInput: () => input.completeSuggestion(),
+    moveSuggestion: (delta) => input.moveSuggestion(delta),
+    suggestionsVisible: () => input.suggestionsVisible(),
     inputIsEmpty: () => input.isEmpty(),
     closeChat: () => controller.closeChat(),
     toggleLatestPrompt: () => conversation.toggleLatestPrompt(),
--- a/clients/tui/src/ui/input.ts
+++ b/clients/tui/src/ui/input.ts
@@ -13,6 +13,8 @@ export interface InputPanel {
   box: BoxRenderable;
   suggestions: BoxRenderable;
   completeSuggestion(): boolean;
+  moveSuggestion(delta: number): boolean;
+  suggestionsVisible(): boolean;
   /** True when nothing is typed, so Enter belongs to whatever pane is behind. */
   isEmpty(): boolean;
   focus(): void;
@@ -42,6 +44,7 @@ export function createInputPanel(
   });
   let syntaxStyle = commandSyntaxStyle(theme);
   let commandStyleId = syntaxStyle.getStyleId('slash-command');
+  const plainTextStyleId = syntaxStyle.getStyleId('plain-text');
   const input = new InputRenderable(renderer, {
     id: 'input',
     width: '100%',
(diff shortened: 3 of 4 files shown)


zatchbell1311-wq:
Splitting into one PR per fix as requested. #399 fix is now at #448. Will open #400 separately. Closing this one.

PR #448, opened 2026-08-26
fix(tui): navigate slash-command suggestions with arrow keys (#399)

### LINKED ISSUE #399: TUI: navigate slash-command suggestions with arrow keys before Tab completion

zatchbell1311-wq:
## Problem

When typing `/` in the command bar, a suggestion panel appears but Up/Down arrow keys had no effect on it. The highlight stayed fixed on the first item, and those keystrokes fell through to the experiment log behind the panel instead. Tab always completed `/help` regardless of which item was highlighted.

Closes #399

## Solution

Added `selectedIndex` state to `command-input.ts` to track which suggestion is highlighted. A shared `renderSuggestions()` helper repaints the list on every move. `completeSuggestion()` now completes `matches[selectedIndex]` instead of always `matches[0]`. A new `navigateSuggestions(direction)` method is exposed on the `CommandInputPanel` interface. In `keybindings.ts`, Up/Down now call `navigateSuggestions()` first — if suggestions are visible, they move the highlight instead of falling through to the experiment log or round view behind them. `app.ts` wires `navigateSuggestions` into the keybinding actions.

### Correctness properties

- Arrow keys in the suggestion panel never reach the view behind it when suggestions are visible
- Tab completes whichever suggestion is highlighted, not always the first
- `selectedIndex` resets to 0 whenever input value changes and matches are recomputed

### Testing

364 pass, 6 fail. The 6 failures are pre-existing launcher test failures unrelated to this change (process spawning tests that fail without a built binary).


vic-lsh:
Pushed 99b653f on top: while testing this locally we found Tab didn't complete the highlighted suggestion when the experiment-log landing view was visible — that view's key handler returned before the shared Tab-completion handler ran. The commit adds a tab branch there delegating to completeInput(), plus two behavioral tests (default highlight and arrow-navigated highlight) that fail without the fix.

PR #469, opened 2026-08-28
fix(cli): narrow TUI rebuild watch set, add progress messages, skip pnpm install

vic-lsh:
## Problem

`src/vibesys/cli.py`'s `_needs_rebuild()` triggers a full, silent TUI rebuild
(`pnpm install --frozen-lockfile` + `generate:protocol` + `build:clients`)
before the interactive TUI process starts, whenever any file under
`_REBUILD_WATCH_DIRS` is newer than `clients/tui/dist/index.js`. That watch
set included all of `src/vibesys/server`, so any backend churn (git pull,
merge, checkout) forced a multi-second-to-minute rebuild on the next
`vibesys` launch, with no indication of what was happening — it reads as "the
app takes forever to load."

## Solution

Three narrowly-scoped changes to `src/vibesys/cli.py`:

1. **Narrow the staleness inputs.** `_REBUILD_WATCH_DIRS` no longer includes
   `src/vibesys/server` wholesale. Instead, `_REBUILD_WATCH_FILES` now lists
   the exact four files that shape the generated protocol schema:
   `schema.py`, `events.py`, `protocol.py`, `diagnostics.py`. I verified this
   set by reading `schema.py`'s `ProtocolDocument` model and its direct
   imports — it only pulls from `events.py` (`RunEvent`) and `protocol.py`
   (`ProtocolRequest`, `Response`, `RunSnapshot`, `ServerMessage`), and
   `protocol.py`/`events.py` both import `Diagnostic` from `diagnostics.py`.
   No other `vibesys` module is imported by any of the three.

   Note: actually running `python -m vibesys.server.schema` imports far more
   than these four files, because Python must execute
   `vibesys/server/__init__.py` (which imports `inspector`, `registry`,
   `runtime`, `service`, `supervisor`, ...) before it can import the
   `schema` submodule, and that package `__init__` transitively pulls in much
   of `vibesys`. None of that extra code feeds the Pydantic models that
   `ProtocolDocument.model_json_schema()` serializes, so it can't change the
   generated schema and is deliberately excluded from the watch set. This is
   documented in a comment above `_REBUILD_WATCH_FILES`.

2. **Tell the user what's happening.** `_run_source_tui` now prints one
   stderr line before a rebuild starts, naming the first offending path
   (`vibesys: TUI bundle is stale (changed: <path>); rebuilding (~30-60s)...`)
   and one after, with elapsed seconds. The pnpm-install step (when it runs)
   gets the same before/after treatment.

3. **Skip `pnpm install` when possible.** New `_needs_install()` only
   requires install when `node_modules` is missing, or
   `pnpm-lock.yaml` is newer than a stamp file
   (`node_modules/.vibesys-install-stamp`) written after the last successful
   install. Otherwise codegen+build run directly. If codegen+build still
   fails (e.g. a partially-removed `node_modules`), `_ensure_source_tui_built`
   falls back to a full install and retries the build once before reporting
   failure.

### Architecture

No ownership/boundary changes; this is all internal to the CLI launcher's
Tier-2 (source-checkout) build path. Refactored `_ensure_source_tui_built`
into three units: `_run_pnpm_install`, `_run_codegen_and_build`, and the
orchestrating `_ensure_source_tui_built` (install-if-needed, build, retry-once
on failure). Staleness detection is split into `_stale_reason` (returns the
first offending path, or `None`) with `_needs_rebuild` as a thin boolean
wrapper kept for existing call sites/tests.

### Correctness properties

- Touching a file outside the narrowed watch set (e.g. any `src/vibesys`
  module other than the four protocol-codegen files) does not trigger a
  rebuild.
- Touching any of `schema.py`/`events.py`/`protocol.py`/`diagnostics.py`, or
  any `clients/{tui,core-state,backend-client}/src` source, or a watched
  config file, still triggers a rebuild (unchanged from before for those
  paths).
- `pnpm install` is skipped when `node_modules` exists and the lockfile is no
  newer than the install stamp; it still runs when `node_modules` is missing,
  the stamp is missing, or the lockfile changed.
- A build failure after a skipped install is retried once with a full
  install before being reported as a failure.
- A rebuild always prints a stderr line before and after, naming the
  triggering path and elapsed time.

### Testing

```
uv run python -m pytest tests/test_cli_launcher.py -q   # 42 passed
uv run ruff check src tests                              # All checks passed
./scripts/check_format.sh                                 # All checks passed
```

Manual: built a throwaway `clients/tui/dist/{index.js,launcher.js}` in this
checkout, then touched `src/vibesys/agents/base.py` (unrelated backend file)
and confirmed `cli._needs_rebuild(root)` stayed `False`; touched
`src/vibesys/server/protocol.py` and confirmed it flipped to `True` with
`_stale_reason(root) == "src/vibesys/server/protocol.py"`. Also drove
`cli._run_source_tui` with `_bun_executable`/`_node_executable`/
`_ensure_source_tui_built` mocked out (did not launch pnpm/bun/the real TUI)
and confirmed the stderr messages read as intended:

```
vibesys: TUI bundle is stale (changed: src/vibesys/server/protocol.py); rebuilding (~30-60s)...
vibesys: TUI bundle rebuilt (0.0s)
```

Did not touch `clients/tui/src/launcher.test.ts` (flaky process tests, out of
scope per the task).

### DIFF
--- a/src/vibesys/cli.py
+++ b/src/vibesys/cli.py
@@ -32,6 +32,7 @@ import re
 import shutil
 import subprocess
 import sys
+import time
 import tomllib
 from dataclasses import dataclass
 from importlib.resources import files
@@ -42,6 +43,18 @@ _MIN_NODE_MAJOR = 20
 
 #: Files and directories whose changes trigger a source-TUI rebuild, relative to
 #: the repository root. Mirrors the staleness check the old ``./vs`` script used.
+#:
+#: Principle: watch only inputs to the *shipped bundle* (``clients/tui/dist``),
+#: not everything that happens to live nearby. On the Python side that means
+#: the modules that actually shape ``ProtocolDocument`` in
+#: ``vibesys.server.schema`` -- ``events.py``, ``protocol.py``, and
+#: ``diagnostics.py`` -- not all of ``src/vibesys/server``. (Running
+#: ``python -m vibesys.server.schema`` also imports the rest of
+#: ``vibesys.server``'s package ``__init__.py``, and transitively much of
+#: ``vibesys``, purely because Python must execute a package's ``__init__``
+#: before importing one of its submodules; none of those extra modules feed
+#: the Pydantic models that get serialized into the schema, so changes to them
+#: cannot change the generated output and are deliberately excluded here.)
 _REBUILD_WATCH_FILES: tuple[str, ...] = (
     "clients/backend-client/package.json",
     "clients/backend-client/tsconfig.json",
@@ -56,12 +69,18 @@ _REBUILD_WATCH_FILES: tuple[str, ...] = (
     "pnpm-lock.yaml",
     "pnpm-workspace.yaml",
     "biome.json",
+    # Inputs to `generate:protocol` (python -m vibesys.server.schema), which
+    # feeds clients/backend-client's generated types and, downstream, the TUI
+    # bundle. See the principle above for why this is the full set, no more.
+    "src/vibesys/server/schema.py",
+    "src/vibesys/server/events.py",
+    "src/vibesys/server/protocol.py",
+    "src/vibesys/server/diagnostics.py",
 )
 _REBUILD_WATCH_DIRS: tuple[str, ...] = (
     "clients/backend-client/src",
     "clients/core-state/src",
     "clients/tui/src",
-    "src/vibesys/server",
 )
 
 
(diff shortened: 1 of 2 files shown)


vic-lsh:
Folded into #468: the StartupTrace launch anchor builds directly on the rebuild-gate changes, and the slow-start work ships as one PR.

Issue #463, opened 2026-08-27
TUI experiment chat: coding agent is slow to respond (cold-start per question, no progress feedback while waiting)


AyanBinRafaih:
### Problem

In the experiment-view chat, each question is answered by a coding agent, and replies take a long time to appear with essentially no progress feedback during the wait. Two compounding causes:
  
  1. Cold start on every question. src/vibesys/agents/client.py:377 forces reuse = kind != "chat" and (...), so chat always takes _run_ephemeral (client.py:493-512): a brand-new session that is closed after a single turn, never cached
     or resumed. The --resume <session_id> path in src/vibesys/_agent_cli/claude.py:176-191 is therefore never used for chat, so every question spawns a fresh claude -p ... subprocess with no in-process context and must reconstruct
     history by re-reading conversation.jsonl + the trajectory. Ahead of that, _ExperimentChatService.ask (src/vibesys/context.py:1436-1497) runs _sync_trajectory() (1541-1565), writing a run-state snapshot and copying logs, on every
     question.
  2. No feedback during the pre-first-token wait. The only guaranteed indicator is static text: the composer title " Message · awaiting agent " (clients/tui/src/ui/chat-composer.ts:231-236), driven by the chatPending boolean — no
     spinner, elapsed timer, or animation. Incremental streaming is wired (clients/core-state/src/core-state.ts:225 routes agent_kind === 'chat' output chunks into the pane), but nothing renders until the agent emits its first
     assistant token, so the entire connect + snapshot I/O + cold subprocess spawn + file re-read window shows no progress at all.
  
  Additionally, follow-ups on the same thread serialize behind a per-service lock (context.py:1440), so a queued question cannot start until the current cold-start turn fully finishes.

### Desired outcome

Asking a question in the experiment chat returns a first visible token quickly and a full answer in a fraction of today's time, with clear, active progress feedback throughout the wait (not just static text).

### Acceptance criteria

- [ ] Time-to-first-token for a chat reply is materially reduced (warm/reused agent session instead of a cold claude -p spawn per question), measured before/after.
- [ ] The chat pane shows active progress while waiting (spinner or elapsed timer, plus streamed partial tokens as soon as they arrive), not only static "awaiting agent" text.
- [ ] Conversation history is carried without re-reading full on-disk transcripts on every question; answers stay correct and contextual across turns.
- [ ] Chat still does not block pause/resume/snapshot or the optimization round (the dedicated connection and participates_in_run_control=False behavior are preserved).
[...]

PR #484, opened 2026-08-28
TUI: Reuse one provider conversation per experiment-chat thread

AyanBinRafaih:
## Problem

Experiment chat throws away its provider session after every question. Each
question therefore pays a full context reload: the whole read-only
investigation prompt plus a transcript re-read, delivered to an agent that was
told exactly the same thing a minute earlier. A CLI process is spawned per
question either way; what is wasted is the conversation, not the process.

Worse, `main` decides which prompt to send from the transcript on disk: any
question after the first gets the short continuation prompt, which points at
`instructions.md` rather than carrying the read-only rules. The rules are
reachable only if the agent chooses to open that file.

The mechanism for fixing this landed separately in #523: `AgentSessionKey`
with a `CHAT` scope, a durable `SessionStore`, and drivers that report
`SessionDisposition.RESET_REQUIRED` whenever they drop and restart the
conversation a session names. This PR is that PR's stated follow-up.

Original diagnosis, per-thread key design, and the first implementation are
@AyanBinRafaih's. The branch has been force-pushed with the session half
rebuilt on #523's contract; the incremental trajectory sync and the composer
spinner it also carried are now #592, which depends on nothing here.

Part of #463. Streaming partial tokens into the chat pane is still deferred:
chat is request/response over `query.chat` and the claude CLI emits an answer
as one message-level chunk, so it needs a new server event kind plus
`--include-partial-messages` driver support.

## Solution

Chat opens its session with `AgentSessionKey(SessionScope.CHAT, thread_id)`
and `reuse_session=True`, so a follow-up continues the same codex/claude
conversation (`codex exec resume` / `claude --resume`) and carries only the
short continuation prompt. The run's default chat, which has no thread ID,
uses the identifier `default`; threads are `uuid4().hex` and cannot collide
with it.

### Shortening a prompt safely

A shortened prompt is only correct inside the conversation that was given the
full instructions. `AgentClient` therefore answers two questions about a key,
not one:

- `provider_session_id(key)` names the conversation the **next** turn
  continues, or `None` when the next turn starts from nothing. A
  driver-reported reset clears it.
- `last_turn_provider_session_id(key)` names the conversation the **last
  completed** turn ran in. A reset does not clear it.

Chat reads the first before its turn to decide whether to shorten, and
compares the second afterwards. This distinguishes the two restarts a reset
reports, which need different responses:

| Restart | When | Response |
| --- | --- | --- |
| Codex thread budget | after the turn it retires | the answer stands; the *next* question cold-starts with the full prompt |
| Missing Codex rollout, refused `claude --resume` | during the turn | the answer came from an agent without the rules; ask once more, in full |

Making `RESET_REQUIRED` itself the re-ask trigger would re-run every second
Codex question in full (three questions, five provider invocations), which is
slower than `main` for no correctness gain.

The re-ask cannot loop: the client evicted the session and cleared the
checkpoint when the driver reported the restart, so the retry cold-starts by
construction and nothing justifies a shortened prompt on the way in.

### Removed

`has_resumable_session`, the `reuse = ... kind != "chat"` special case in
`client.py`, `AgentTurnRequest.cold_start_instructions` with its agentshim
threading, `cold_start_system_prompt` on `invoke_text` and every
implementation, and `experiment_chat_resumed_prompt`. All of them existed to
work around drivers that restarted a conversation without saying so; #523 made
them say so. `ExperimentChatSession._history` and `_load_history` went with
them: nothing reads the loaded transcript now that the prompt choice is driven
by the conversation. `_append_exchange` still writes `conversation.jsonl`,
which is what the agent reads.

### Architecture

Ownership is unchanged from #523: the driver owns the conversation during a
turn and reports restarts; the client owns the live handle and the checkpoint;
the chat session owns the prompt decision. Chat still runs on its dedicated
connection with `participates_in_run_control=False`.

```mermaid
flowchart TD
  Ask["ExperimentChatSession.ask"] -->|"provider_session_id(CHAT:thread)"| Client["AgentClient"]
  Client -->|"named"| Short["continuation prompt"]
  Client -->|"None"| Full["full read-only prompt"]
  Short --> Turn["invoke_text(reuse_session=True, session_key)"]
  Full --> Turn
  Turn --> Driver["AgentShim session"]
  Driver -->|"codex exec resume / claude --resume"| Provider["CLI provider"]
  Driver -->|"restart: budget / rollout / stale resume"| Reset["RESET_REQUIRED"]
  Reset --> Client
  Turn -->|"last_turn_provider_session_id"| Compare{"ran where the<br/>shortened prompt assumed?"}
  Compare -->|yes| Answer["answer stands"]
  Compare -->|no| Reask["re-ask once, full prompt"]
  Reask --> Answer
```

### Correctness properties

- A shortened prompt is sent only into a conversation that a completed turn
  reported, never because a transcript exists on disk. Providers that never
  report a conversation always get the full prompt.
- An answer is returned from a shortened prompt only if that turn ran in the
  conversation that justified it.
- A conversation retired *after* serving a turn costs one answer nothing; the
  next question carries the full instructions.
- A conversation replaced *during* a turn costs exactly one extra invocation,
  never more: the retry cold-starts by construction, so the re-ask cannot
  recur.
- A stale Claude session ends in an answer, not a `RuntimeError`.
- Per-thread keys: two chat threads never share a provider conversation, and
  the default chat's identifier cannot collide with a thread's.
- `CHAT` is a durable scope, so a wired store checkpoints it; chat clients
  currently get the null store, and reuse is in-process. Nothing depends on
  which.
- Non-CLI backends (deepagents, stub) name no conversation and behave exactly
  as before.
- Run-control isolation is untouched: `participates_in_run_control=False`.

### Testing

```
uv run pytest tests/server tests/vibesys/agents -q --no-cov   -> 630 passed
uv run pytest tests/vibesys/loops -q --no-cov                 -> 605 passed
./scripts/check_format.sh                                      -> 481 files already formatted
./scripts/check_lint.sh                                        -> All checks passed
./scripts/check_types.sh                                       -> All checks passed
uv run python scripts/check_doc_links.py                       -> 326 files, no broken links
clients/tui: bun test --max-concurrency 1 src                  -> 495 passed (unchanged by this PR)
pnpm check:clients / check:ts                                  -> clean
```

New coverage in `tests/server/test_chat_session.py`:

- Prompt selection and the re-ask rule against a scripted client: the key and
  `reuse_session` chat asks for; the full prompt with no conversation named;
  the shortened prompt inside one; a re-ask when the turn ran elsewhere; no
  re-ask when the conversation was retired only afterwards; no re-ask for a
  cold turn.
- The Codex renewal, end to end through a real `AgentShimDriver`, the real
  `AgentClient`, and a real `ExperimentChatSession`, using the driver suite's
  own `_FakeAgent`: three questions produce three provider turns whose prompts
  are `[full, continuation, full]`. Checked against a driver patched to renew
  silently (the pre-#523 behavior), which produces
  `[full, continuation, continuation, full]` — the bug this fixes.
- The stale-Claude path, likewise end to end: the shortened prompt, the
  driver's own retry of it from a fresh conversation, then the question again
  with the rules restored, and an answer rather than an exception.

Not covered by the suite: a credentialed live run measuring resumed answer
latency. @AyanBinRafaih's manual smoke on this branch's earlier form measured
30.3s cold against 14.2s resumed on a comparable question, with `--resume`
visible on the follow-up process.

### Docs

`docs/contributing/agent-drivers.md` gains "Retired after a turn, or replaced
during one" under Provider session resume: the two accessors, which restarts
fall on which side, and that the experiment chat is the caller that cares.

### DIFF
--- a/docs/contributing/agent-drivers.md
+++ b/docs/contributing/agent-drivers.md
@@ -62,6 +62,26 @@ nothing about whether the conversation is still resumable, so the client keeps
 the checkpoint and only a driver-reported reset (or a refused adoption) clears
 it.
 
+### Retired after a turn, or replaced during one
+
+A reset says the conversation is gone, not when it went. The two cases differ
+for any caller that shortens a prompt because a conversation already carries
+its instructions, so `AgentClient` answers both questions separately:
+
+- `provider_session_id(key)` names the conversation the **next** turn
+  continues, or `None` when the next turn starts from nothing. A reset clears
+  it.
+- `last_turn_provider_session_id(key)` names the conversation the **last
+  completed** turn ran in. A reset does not clear it.
+
+An over-budget Codex thread is retired after it has answered, so the two
+disagree only from the next turn onward: that answer stands, and the next
+prompt is a cold one. A missing rollout or a refused `claude --resume` is
+replaced mid-turn, so the last turn ran somewhere the caller did not intend,
+and a caller that shortened its prompt has to ask again in full. The
+experiment chat is the caller that does this today
+(`src/server/chat/session.py`).
+
 ## Mock driver
 
 `driver = "mock"` is test infrastructure. It satisfies the same driver
--- a/src/server/chat/factory.py
+++ b/src/server/chat/factory.py
@@ -19,6 +19,7 @@ from server.chat.session import (
 )
 from server.events import ChatThreadCreatedData
 from vibesys.agents import build_agent_client
+from vibesys.agents.session_key import AgentSessionKey, SessionScope
 from vibesys.domains.environment import EnvironmentBindMount
 from vibesys.run import RunLogger
 from vibesys.run.integration import AgentSelection, RunAttachment
@@ -32,6 +33,13 @@ if TYPE_CHECKING:
     from server.controller import RunController
     from server.execution import ExecutionTracker
     from vs_project import Project
+
+
+#: Session-key identifier for the run's default chat, which has no thread ID of
+#: its own. Threads are ``uuid4().hex``, so this cannot collide with one; it is
+#: the same name the client already shows the default thread under
+#: (``DEFAULT_CHAT_THREAD_ID`` in ``clients/core-state``).
+DEFAULT_CHAT_THREAD = "default"
 
 
 class SelectionResolver(Protocol):
@@ -288,6 +296,12 @@ class ExperimentChatFactory:
                 controller=self._controller,
                 executions=self._executions,
                 agent_client=resources.client,
+                # One conversation per thread. The run's default chat has no
+                # thread ID of its own, so it names the identifier the threads
+                # cannot collide with.
+                session_key=AgentSessionKey(
+                    SessionScope.CHAT, DEFAULT_CHAT_THREAD if thread_id is None else thread_id
+                ),
                 workspace=self._workspace,
                 state_dir=state_dir,
                 agent_shared_state_dir=resources.agent_shared_state_dir,
--- a/src/server/chat/session.py
+++ b/src/server/chat/session.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from server.chat.evidence import TrajectoryEvidence
     from server.controller import RunController
     from server.execution import ExecutionTracker
+    from vibesys.agents.session_key import AgentSessionKey
 
 
 class ChatAgentClient(Protocol):
@@ -26,6 +27,14 @@ class ChatAgentClient(Protocol):
 
     def invoke_text(self, **kwargs: Any) -> str:  # noqa: ANN401  # Mirrors agent clients.
         """Invoke the agent with the driver-specific keyword contract."""
+        ...
+
+    def provider_session_id(self, session_key: AgentSessionKey) -> str | None:
+        """Name the provider conversation the next turn on the key continues."""
+        ...
+
+    def last_turn_provider_session_id(self, session_key: AgentSessionKey) -> str | None:
+        """Name the provider conversation the key's last completed turn ran in."""
         ...
 
 
@@ -44,6 +53,9 @@ class ExperimentChatDependencies:
     controller: RunController
     executions: ExecutionTracker
     agent_client: ChatAgentClient
+    #: Identity of this thread's provider conversation. Follow-ups reuse it so
+    #: the agent keeps the context of the questions before them.
+    session_key: AgentSessionKey
     workspace: Path
     state_dir: Path
     agent_shared_state_dir: str
@@ -66,10 +78,11 @@ class ExperimentChatSession:
         dependencies: ExperimentChatDependencies,
         resources: Closeable | None = None,
     ) -> None:
-        """Initialize a serialized chat session and load its transcript."""
+        """Initialize a serialized chat session over one provider conversation."""
         self._controller = dependencies.controller
         self._executions = dependencies.executions
         self._agent_client = dependencies.agent_client
+        self._session_key = dependencies.session_key
         self._workspace = dependencies.workspace
         self._state_dir = dependencies.state_dir
         self._evidence = dependencies.evidence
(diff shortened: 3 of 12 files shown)


vic-lsh:
Review finding:

[P1] Preserve the full chat instructions when stale-session recovery cold-starts Codex (`src/vibesys/context.py:1467-1468`).

The service selects `_resumed_prompt` before invocation whenever the cached turn exposed a provider session ID. If that Codex rollout has since disappeared, AgentShim catches the missing-rollout error, clears the ID, and successfully starts a fresh Codex process with the same short reminder prompt. The fresh process never receives the original read-only rules, evidence paths, or prohibition on editing, even though the reminder says those instructions still apply.

Please make stale-rollout recovery retry with the full/continuation prompt, or surface the reset to the chat service so it can issue a cold prompt. A regression test should cover a cached ID whose rollout disappears before the next question.

vic-lsh:

vic-lsh:
@AyanBinRafaih thanks — the diagnosis here was right and the per-thread key was the right shape. Two things changed since you opened this: #523 landed the durable session contract (`AgentSessionKey`, `SessionStore`, `AgentCapabilities.provider_session_resume`, `AgentSession.resume_provider_session`, and drivers that report `SessionDisposition.RESET_REQUIRED` on every restart), and I've split this PR in two rather than asking for another round on a branch that now sits on top of a different contract.

**The split.** Your incremental trajectory sync and your composer spinner depend on nothing in #523, so they moved to #592 (draft, with the review findings applied and a test that fails on the old wiring). This PR keeps the session half only, force-pushed as a rework on top of `main`. That way the sync and the spinner are not held behind a driver-contract review, and this PR's review is about sessions only.

Here is what changed and why, in the review's terms.

**[P1] The parallel session mechanism is gone.** `has_resumable_session(session_key: str)`, the `reuse = ... kind != "chat"` special case in `client.py`, `AgentTurnRequest.cold_start_instructions` and its agentshim threading, `cold_start_system_prompt` on `invoke_text` and every implementation, and `experiment_chat_resumed_prompt` are all absent. They existed to work around drivers that restarted a conversation without saying so; #523 made drivers say so, so the workaround is not needed and a second mechanism for the same problem would have to be kept in step with the first. Chat now opens its session with `AgentSessionKey(SessionScope.CHAT, thread_id)` and `reuse_session=True`, and `client.py`'s default reuse is unconditional again.

**[P1] The recurring bug: every second Codex question ran without the read-only rules.** The Codex thread budget retires a thread after two turns. Nothing reported that, so `has_resumable_session` stayed true and the shortened prompt went into the replacement thread. There is now a regression test that drives a real `AgentShimDriver(provider="codex")` through the real `AgentClient` and a real `ExperimentChatSession` with the driver suite's own `_FakeAgent`, over three questions, and asserts the provider saw `[full, continuation, full]` — three turns, no re-asks. I checked it against a driver patched to renew silently, the pre-#523 behavior: that produces `[full, continuation, continuation, full]`, i.e. the third question answered with no rules at all.

**[P1] Recovery policy: I did not make `RESET_REQUIRED` the re-ask trigger.** A reset says the conversation is gone; it does not say *when* it went, and the two cases are different for a caller that shortened its prompt:

- The Codex thread budget is evaluated **after** the turn (#523 moved it there deliberately). That turn ran in the conversation the shortened prompt assumed, and its answer is good. Only the *next* question needs the full prompt, which the prompt choice above already gives it.
- A missing rollout, or a `claude --resume` the CLI refuses, is replaced **during** the turn. That answer came from an agent that never saw the rules, and only that case needs a re-ask.

Re-asking on `RESET_REQUIRED` alone would have re-run every second Codex question in full — three questions would cost five provider invocations, which is *slower* than the `main` this PR is trying to speed up, for no correctness gain. So the trigger is the clause in the second half of your recovery rule: the turn ran somewhere other than the conversation that justified the shortened prompt. `AgentClient` exposes both facts:

- `provider_session_id(key)`: the conversation the **next** turn continues; a reset clears it. Chat reads it to decide whether the shortened prompt is justified.
- `last_turn_provider_session_id(key)`: the conversation the **last completed** turn ran in; a reset does not clear it. Chat compares it afterwards.

One accessor cannot answer both: after a renewal the first must say "nothing to continue" while the second must say "it ran in the thread you meant".

**[P2] The stale-Claude path re-asks instead of failing the question.** #523's AgentShim session retries a failed *resumed* Claude turn once from a fresh conversation and reports the reset. Chat now sees that the turn ran elsewhere and asks again with the full instructions, so the operator gets a ruled answer rather than a `RuntimeError`. Tested end to end through the real driver: the provider sees the shortened prompt, the driver's own retry of it, then the question again with the rules restored, and `ask` returns an answer.

The re-ask cannot loop: the client evicted the session and cleared the checkpoint when the driver reported the restart, so the retry cold-starts by construction and `continued` is `None` on the way in.

**[P2] Prompt selection is no longer driven by the transcript.** `main` sent the short continuation prompt whenever `conversation.jsonl` was non-empty, cold spawn or not, leaving the read-only rules reachable only if the agent chose to open `instructions.md`. It is now driven by whether a provider conversation actually carries them. That makes the cold path *stronger* than before, not just faster on the warm path.

**[P2] `DeepAgentsClient` and `StubAgentClient` name no conversation.** Both return `None` from both accessors, so their chat threads keep today's behavior. `DeepAgentsClient` needs the override rather than the inherited implementation for the reason you found: it never runs `AgentClient.__init__`, so it owns neither the session cache nor the store the base version reads. A deepagents "thread" is a LangGraph checkpointer entry, not a provider conversation. `PlainLoopAgentClient` forwards both.

**[P3] Dead state removed.** With prompt selection off the transcript, `ExperimentChatSession._history` and `_load_history` were written and never read, so they are gone. `_append_exchange` still writes `conversation.jsonl`, which is what the agent reads.

**[P3] Docs.** `docs/contributing/agent-drivers.md` gains a "Retired after a turn, or replaced during one" subsection under Provider session resume: the two accessors, which restarts fall on which side, and that the experiment chat is the caller that cares.

**[P3] PR body corrections.** The test paths named commands that do not exist in this repo's layout (`tests/test_context.py`, `tests/agents/`); they are `tests/server`, `tests/vibesys/agents`. "Instead of a cold spawn" is wrong and is now "instead of a full context reload": a CLI process is still spawned per question, and what reuse saves is reloading the conversation, not the process. `Closes #463` is now `Part of #463`, since streaming partial tokens into the chat pane is deferred.

Your per-thread key, the `RESET_REQUIRED`-evicts-the-session rule, and the non-CLI-backends-opt-out rule are all kept; they were the right calls.

Checks on the pushed head: `tests/server` + `tests/vibesys/agents` 630 passed, `tests/vibesys/loops` 605 passed, format / lint / types / doc links clean, and the client-side `bun test` (495), `check:clients`, `check:ts` clean (unchanged by this half).

vic-lsh:
Follow-up on the head above (`d78cf84`): CI's patch-coverage gate wanted the new lines covered, so the push adds tests for the paths the rework introduced rather than lowering the bar — the two conversation accessors on `StubAgentClient`, `DeepAgentsClient`, and `PlainLoopAgentClient`, and `ExperimentChatSession`'s error handling (an agent failure normalized to `RuntimeError`, a cancellation propagated unwrapped). Patch coverage is 100% (`uv run diff-cover coverage.xml --compare-branch=origin/main`), and all checks on this head pass.

Separately, and not from this PR: **`main` is broken for `--stub-agent` runs.** #523 added `implementer_driver=ctx.agent_client.driver_name` at `src/vibesys/loops/agent/loop.py:3240`, and `StubAgentClient` is a standalone class with no `driver_name`, so every stub agent-loop run dies with `AttributeError: 'StubAgentClient' object has no attribute 'driver_name'` right after the first judge. That is what fails the "Verify the installed wheel with a sanitized PATH" step in the release-wheel job on #592 (whose own checks — python-checks, test, tui, typecheck, sandbox, queue-native — all pass). It needs its own fix on `main`; I have not touched it in either PR.

PR #495, opened 2026-08-29
TUI: Fix hypothesis resolution issues and unresolved metrics

### LINKED ISSUE #475: TUI: Hypothesis resolves PROVEN with no measurement, and renders as "Accepted"
### Observed behavior

A hypothesis that was never measured is persisted with `resolution: proven` and rendered in the Experiments pane as a green "Accepted". Observed in a 2-round `queue-mpmc` run:

```
hypothesis_id: lockfree-ring-v1
  resolution      : proven
  declared_outcome: nominated
  review          : pass
  measurement     : None
   round 1 | official_evaluation: False | hypothesis_outcome: nominated | judge_verdict: pass
      perf_metric: None | perf_unit: None | perf_delta_pct: None
      metrics: {}
```

Pane row: `lockfree-ring… | 1 | Lock-free bounded MPMC ring buffer replacing global mutex | — | Accepted | —`

That hypothesis shipped a deadlock. The first official evaluation, one round later, caught it:

```
[framework-accuracy] FAIL: [stderr] FAIL - mpmc: boundary history at capacity 1:
candidate operation timed out after 5s while waiting for a response
```

Mechanism, `src/vibesys/loops/agent/hypotheses.py:42-70`. `nominated` and `supported` are absent from the outcome-to-resolution map at `:53-58`, so control reaches:

```python
elif resolution is None:
    if not evidence.benchmark_expected:
        resolution = HypothesisResolution.PROVEN     # :61-63
elif evidence.declared is HypothesisOutcome.SUPPORTED:
    resolution = HypothesisResolution.PROVEN         # :66-67
```

`benchmark_expected` is wired from `record.official_evaluation` at `:297`, reached via `append_round` at `:203`, which is the loop's own write path. A judge `pass` is the only other precondition. Executed:

```
declared=SUPPORTED, official=False, perf_metric=None -> proven, measurement=None
declared=SUPPORTED, official=True,  perf_metric=None -> proven      (unconditional)
declared=NOMINATED, official=False, perf_metric=None -> proven
declared=NOMINATED, official=True,  perf_metric=None -> inconclusive
judge_verdict='deferred'                             -> None
```

Second defect at the same site: the two callers bind different facts to `benchmark_expected`. `loop.py:3360` passes `framework_benchmark_configured` ("is a benchmark attached at all"); `hypotheses.py:297` passes `record.official_evaluation` (per round). For `nominated` this persists self-contradictory state: `RoundRecord.hypothesis_outcome == 'inconclusive'` alongside `Hypothesis.resolution == 'proven'`.

`'proven'` renders as "Accepted" at `clients/tui/src/ui/experiment-log.ts:617`, colored `theme.success`.

### Expected behavior

A hypothesis with no trusted measurement must never resolve `proven`. Absence of a benchmark is absence of evidence, not proof, so the resolution should remain `inconclusive` or a distinct `unmeasured`. The function's own docstring at `hypotheses.py:45` states the intended rule: "Resolve one declaration only after review and trusted evidence." Both call sites should bind the same fact.

### Reproduction

1. `node clients/tui/dist/launcher.js --project <copy of examples/data-structures/repositories/queue-rs> --task mpmc --profiler none --cli-provider claude --max-rounds 2`
2. Round 1 receives no official evaluation (default `--official-eval-every 3`; `_official_evaluation_reason` returns `None` for ordinary rounds).
3. Experiments pane shows Measured `—`, Outcome `Accepted`.
4. `<project>/.vibesys/state/runs/<run>/agent/state.json` shows `resolution: proven`, `measurement: null`, `official_evaluation: false`.

Without a full run:

```python
state = start_hypothesis(...)                      # declared_outcome=NOMINATED
append_round(state, RoundRecord(official_evaluation=False, perf_metric=None,
                                metrics={}, judge_verdict="pass", ...))
assert state.hypotheses[0].resolution is HypothesisResolution.PROVEN   # passes
assert state.hypotheses[0].measurement is None                          # also passes
```

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell
grep benchmark_expected tests/ -> only test_hypotheses.py:360,371, both =True
_correct_legacy_resolution is gated on judge_verdict is None (hypotheses.py:311);
loop.py:3466 always sets a verdict, so it never runs on loop-written records.
reproject_run_evidence on resume reproduces the same 'proven'.
```

#284

### Pre-submission checks

### LINKED ISSUE #476: TUI: `hypothesis_id` schema fails runs for re-used IDs, asserting invalid input
### Observed behavior

A 2-round `queue-mpmc` run terminated at round 2:

```
VibeSys · failed · orchestrator · round-2-plan
Run failed: Run received invalid input
Detail: ValueError: hypothesis ID 'lockfree-ring-v1' was already used
```

The designer reused the round-1 id, which is exactly what the schema instructs. `src/vibesys/schemas.py:623-629`:

```python
hypothesis_id: str = Field(
    default="",
    description=(
        "Stable short identifier for the hypothesis. Reuse it across rounds while "
        "the same implementer investigation remains active."
    ),
)
```

`src/vibesys/loops/agent/loop.py:1107-1110`:

```python
if state.by_id(plan.hypothesis_id) is not None:
    raise ValueError(f"hypothesis ID {plan.hypothesis_id!r} was already used")
```

`AgentRunState.hypotheses` only ever grows (`hypotheses.py:153`) and is never pruned, so uniqueness is permanent over run history, not "no two active at once". `tests/loops/agent/test_orchestrate.py:1119-1144` pins this: round 1's hypothesis has already completed and round 2 reusing its id still raises.

The reuse the schema asks for is structurally unreachable. `loop.py:2527-2532` and `:2630-2641`: when an investigation is active the designer is skipped entirely ("continuing {id}; designer invocation skipped"); when a plan is requested the investigation is by construction not active, so any reuse is fatal. `hypothesis_updates` is not a continuation channel either: `apply_strategy_updates` (`hypotheses.py:229-264`) rejects both the active hypothesis and incomplete ones. Compounding it, `_MAX_CONTINUATION_ROUNDS_WITHOUT_DESIGN_REVIEW = 2` (`loop.py:156`) re-invokes the designer while the implementer investigation is still active, which is precisely the state the field description describes.

Neither `orchestrator_plan_prompt.j2` nor `orchestrator_pre_round_prompt.j2` mentions `hypothesis_id` or uniqueness, so the field description is the model's only guidance and it points the wrong way.

The failure is uncaught and unretried. `_validate_orchestrator_plan_state` is called bare at `loop.py:1085` inside the round loop, whose enclosing `try` has only `finally: ctx.close()`. The sole handler is `runtime.py:113-116`, which records and re-raises. Structured-output repair does not apply: it fires for schema failures inside the invocation (observed twice in a later run, both repaired), while this validator runs after an already-schema-valid object is returned. A malformed plan is repaired; a semantically invalid one kills the run.

### Expected behavior

The prompt contract and the validator must agree. Either the validator permits reuse for a continuing investigation, or the field description states the real rule. Given the framework already skips the designer for continuing work, the description should read approximately:

> Fresh unique identifier for a newly proposed hypothesis. Must never repeat an > id used earlier in this run. Continuing an in-flight investigation is handled > by the framework without a new plan; use `hypothesis_updates` only to park or > abandon previously completed hypotheses.

An invalid plan should also be reprompted once rather than terminating the run, matching how malformed output is already handled.

### Reproduction

Non-deterministic end to end, because it depends on whether the designer mints a fresh id. Two consecutive identical 2-round `queue-mpmc` runs: the first reused `lockfree-ring-v1` and died at `round-2-plan`; the second minted `cacheline-pad-v1` and completed.

Deterministic at the unit level:

```python
_validate_orchestrator_plan_state(
    OrchestratorPlan(hypothesis_id="h1", ...),
    state_that_already_contains_completed_h1,
)   # ValueError: hypothesis ID 'h1' was already used
```

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell
[child] calling the real orchestrator-plan validator
[child] run_server raised ValueError: hypothesis ID 'lockfree-ring-v1' was already used
seq=3 type=run_failed status=failed  text='Run received invalid input'
```

#284

### Pre-submission checks

### LINKED ISSUE #477: TUI: Evaluator-declared metric direction is resolved and then discarded, so protocol-v2 tasks can never resolve a hypothesis
### Observed behavior

`_read_protocol_benchmark` (`src/vibesys/loops/agent/loop.py:1917-1959`) correctly resolves the metric direction as `configured or declared`, reading `hello.metrics[name].direction` from the evaluator. The caller at `:2084-2094` copies `metric_name`, `metric_value` and `row` and never reads `outcome.metric_direction`, which has zero readers repo-wide. `_run_framework_benchmark` then rebuilds direction from `objectives` alone at `:2154-2165`, with a legacy-only fallback `or ("max" if result_spec is not None else None)`. For `result_protocol = 2`, `result_spec` is `None`, so with no `objectives.toml` the direction is `None`.

Executed against the real functions:

```
_read_protocol_benchmark(stream, objectives=[]) ->
  FrameworkBenchmarkOutcome(metric_name='total_ops_per_sec', metric_value=9500000.0,
                            metric_direction='max', ...)

_run_framework_benchmark(result_protocol=2, objectives=[]) ->
  FrameworkBenchmarkOutcome(..., metric_direction=None)      DROPPED

control A: objectives=[total_ops_per_sec:max] -> 'max'
control B: legacy [benchmark.result] path     -> 'max'

direction='max'  baseline=2,000,000 -> 9,500,000 (+375%)
    _resolve_metric_evidence -> proven        scalar_candidate_retained -> True
direction=None   baseline=2,000,000 -> 9,500,000 (+375%)
    _resolve_metric_evidence -> inconclusive  scalar_candidate_retained -> None
```

`_resolve_metric_evidence` short-circuits at `hypotheses.py:77` and `scalar_candidate_retained` at `:100` when direction is not in `{"max","min"}`. Direction is persisted as `RoundRecord.perf_direction` at `loop.py:3440`, so `None` propagates.

All four shipped `result_protocol = 2` tasks lack an `objectives.toml`: `queue-rs/.vibesys/tasks/{mpmc,mpsc,spsc}` and `examples/microservices/train-ticket`. `find . -name objectives.toml` returns only the seven `examples/model-serving/*` paths, and `objectives` is loaded exclusively from that filename (`main.py:2127-2133`). The queue evaluator does declare direction: `resources/evaluators/queue/result_stream.go:38-42`, `schema.Number("total_ops_per_sec", Unit("ops/s"), Direction(Max))`. That declaration reaches the reader and is discarded.

Live corroboration from a 2-round `queue-mpmc` run: `run.json` has `"objectives": []`, and the recorded measurement is

```
{'metric': 'total_ops_per_sec', 'value': 9337789.413431, 'direction': None,
 'baseline_value': None, 'delta_pct': None}
resolution: inconclusive
```

Note the inversion: the 11 legacy `[benchmark.result]` tasks get `"max"` for free, so migrating a task to the newer protocol silently loses its verdict.

### Expected behavior

A direction declared by the evaluator in its `hello` record must reach every comparison that decides improvement. `_read_protocol_benchmark` already resolves it correctly; the caller should propagate `outcome.metric_direction` instead of recomputing it from `objectives` alone.

### Reproduction

1. `--project <queue-rs copy> --task mpmc --profiler none --cli-provider claude --max-rounds 2`
2. Let a round record a measurement.
3. `agent/state.json` shows `perf_direction: null` and `resolution: inconclusive` despite a finite metric.

Or directly: call `_run_framework_benchmark` with `result_protocol=2`, `objectives=[]` and a well-formed stream whose `hello` declares `direction: max`; the returned `metric_direction` is `None` while the inner reader had `'max'`.

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell
No test asserts metric_direction on the real protocol path. The two occurrences in
tests/loops/agent/test_orchestrate.py:3349,3354 are hand-supplied values in a test
that fully mocks _run_framework_benchmark. Real-path protocol tests always pass a
matching objectives=[...], so the buggy recomputation coincidentally returns the
right answer.

Direction cannot be recovered downstream: every perf_direction read
(hypotheses.py:295,426,479-482) falls back to _configured_legacy_direction, which
returns None for empty legacy_directions and additionally short-circuits on
record.judge_verdict is not None (:503), so it can never rescue a modern record.
```

#284

### Pre-submission checks

### LINKED ISSUE #478: TUI: Performance baseline never resolves at any `official_eval_every` above 1
### Observed behavior

`metric_baseline` (`src/vibesys/loops/agent/hypotheses.py:110-133`) filters candidate records to `official_evaluation and perf_metric is not None and perf_unit == metric`, then matches `item.commit == parent_commit`. `parent_commit` is round N-1's commit. For any cadence k >= 2, a round being official implies round N-1 was provisional, so the exact-commit match can never hit on an official round.

Replaying `loop.py:3342-3398` over 8 rounds at the default `official_eval_every=3`:

```
rnd official      reason  parent_commit  baseline  delta_pct  resolution
  1    False        None     head-sha-1      None       None  proven
  2    False        None             c1      None       None  proven
  3     True     cadence             c2      None       None  inconclusive
  4    False        None             c3     115.0       None  proven
  5    False        None             c4      None       None  proven
  6     True     cadence             c5      None       None  inconclusive
  7    False        None             c6     130.0       None  proven
  8     True final_round             c7      None       None  inconclusive

official rounds:                  [3, 6, 8]
rounds where a baseline resolved: [4, 7]      <- never an official round
rounds where delta_pct is None:   [1..8]      <- all of them
```

Sweeping the cadence: `delta_pct` is `None` in every round for `official_eval_every` in {2, 3, 4}. Only `=1` produces deltas. The baseline resolves only on rounds that have no official metric to compare it against.

Two further contributors. Branches 2 and 3 of `metric_baseline` are dead from the agent loop: both callers (`loop.py:3342`, `hypotheses.py:454-460`) pass a commit that `loop.py:2610-2613` guarantees non-`None` via a `ctx.git.current_sha()` fallback. And the `perf_unit == metric` filter runs before the commit match, so a renamed metric drops an otherwise valid official parent.

Separately, no baseline is ever re-measured. `_run_framework_benchmark` has one caller, `_run_framework_gates`, which has two mutually exclusive call sites, both passing `candidate_revision=ctx.git.current_sha()` (post-edit HEAD). Repo-wide search for `measure_baseline|rebaseline|baseline_run|benchmark_parent|evaluate_commit|reference_run` returns zero hits. Every parent-commit checkout materializes a tree for editing and then measures the modified tree.

There is no operator-visible signal for an unresolved baseline. `experiments.py:50-52` projects only `perf_metric`, `perf_unit` and `perf_delta_pct`; `baseline_round`, `baseline_commit` and `baseline_value` never reach the wire. `experiment-log.ts:635-643` `formatMeasured` falls back to the raw absolute metric when `delta_pct` is null, so an unresolved baseline is indistinguishable from a deliberate absolute display. `hypotheses.py:74-79` collapses "no baseline" and "genuinely within noise" into the same `INCONCLUSIVE`.

Live corroboration: the one measured round of a 2-round run recorded `value: 9337789.41` with `baseline_value: None`, `delta_pct: None`, and `Kept: —`.

### Expected behavior

An official round should resolve a baseline. Either the lookup should accept the most recent official record rather than requiring an exact parent-commit match, or the framework should re-measure the parent commit in the same session so a directly comparable number exists. Failing that, "baseline could not be resolved" must be distinguishable from "measured, no improvement", both in the round record and in the projection.

### Reproduction

1. Run any task with `--official-eval-every` >= 2 for enough rounds to produce at least two official evaluations.
2. Inspect `agent/state.json`: every `perf_delta_pct` is `null`, including on official rounds.
3. `formatMeasured` displays the absolute metric with no indication that no comparison was made.

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell
tests/loops/agent/test_hypotheses.py:327
  test_metric_baseline_prefers_exact_parent_commit_and_fails_closed
asserts parent_commit="missing" -> None. Failing closed on an unrelated commit is
intentional. What is not designed is that the loop then supplies a commit that
structurally can never match.
```

#284

### Pre-submission checks

### LINKED ISSUE #479: TUI: `official_evaluation` is set for metrics the agent reported about itself
### Observed behavior

When a task declares neither `[benchmark.result]` nor `result_protocol`, `_run_framework_benchmark` short-circuits (`loop.py:2024-2025`), `framework_perf_metric` stays `None`, and `loop.py:3236-3247` falls through to the implementer agent's own self-reported number:

```python
elif implementation_metric is not None:
    perf_metric = implementation_metric               # ImplementerResponse.perf_metric
    accepted_metrics = dict(implementation.metrics)   # agent-declared row
```

`official_evaluation` is still set `True`, because `loop.py:3315-3320` requires only `bool(ctx.judge_accuracy_command) or framework_benchmark_configured`. Executed through the repo's orchestrate harness:

```
NO RESULT CONTRACT:   perf_metric = 321.5   official_evaluation = True
                      metrics = {'aggregate_throughput': 321.5, 'p99_latency_ms': 87.0}
                      <- agent self-reported ImplementerResponse.perf_metric = 321.5

WITH RESULT CONTRACT: perf_metric = 99.0    official_evaluation = True
                      <- framework measured; correctly overrides the agent's 321.5
```

The two records are structurally indistinguishable. `RoundRecord` has 30 fields and no provenance field; `experiments.py:59-71` projects both identically. This contradicts the field's own contract comment at `libs/vs-loop-state/src/vs_loop_state/agent.py:72-74`: "Only framework-owned gates can set this."

13 of 28 shipped `vibesys.input.toml` declare neither contract, so this is the default path for them, including all of `examples/model-serving/*` and `examples/data-structures/kvstore-default`.

Second, larger half: the evolutionary loop has no framework benchmark at all.

```
'benchmark_result':          agent-loop param=True  evolve-loop param=False
'benchmark_result_protocol': agent-loop param=True  evolve-loop param=False

src/vibesys/loops/agent/loop.py:2006  def _run_framework_benchmark(...)
src/vibesys/loops/agent/loop.py:2253      benchmark = _run_framework_benchmark(
evolve/ call sites: (none)
```

`run_evolve_loop` does not accept a result contract, so a task that declares one silently loses it under `--outer-loop evolve`. Fitness comes straight from the profiler agent at `evolve/loop.py:586-590`. `run_accuracy_gate` is the only framework gate in that loop.

### Expected behavior

`official_evaluation` should mean "produced by a framework-executed, agent-inaccessible measurement". An agent-reported number should either not be recorded as official, or `RoundRecord` should carry provenance so the projection and the operator can tell the two apart. The evolve loop should honour a declared result contract rather than dropping it.

### Reproduction

1. Use any task declaring neither `[benchmark.result]` nor `result_protocol`, for example `examples/model-serving/Llama-3-8B`.
2. Run a round to completion.
3. `agent/state.json` shows `official_evaluation: true` with `perf_metric` equal to the value in the implementer's `ImplementerResponse`.

For the evolve half: run any `result_protocol` task with `--outer-loop evolve` and observe that no framework benchmark executes.

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell

```

#284

### Pre-submission checks

### LINKED ISSUE #480: TUI: Framework benchmark result file is a shared `/tmp` path with no per-invocation identity
### Observed behavior

`src/vibesys/loops/agent/loop.py:2032` builds the result path as a pure function of round and retry:

```python
output_path = f"/tmp/vibesys-framework-benchmark-{round_number}-{retry}.json"
command = (
    f"{execution_base} {shlex.quote(output_argument)} {shlex.quote(output_path)}"
    f" && printf '\n{_FRAMEWORK_BENCHMARK_MARKER}\n'"
    f" && cat {shlex.quote(output_path)}"
    f" && printf '\n{_FRAMEWORK_BENCHMARK_END_MARKER}\n'"
)
```

There is no `rm -f` before the run, no nonce, no run id, and no check that the file's mtime postdates the invocation. On the local backend the sandbox shares the host's `/tmp` (`backends/local.py:85-91`; `stat` confirms the same inode as host `/tmp`, owner root, mode `drwxrwxrwt`).

Files from a different user are present on this machine right now under exactly that template:

```
-rw-r--r-- 1 ryanc175 ryanc175    32 Aug 18 06:17 /tmp/vibesys-framework-benchmark-1-0.json
-rw-r--r-- 1 ryanc175 ryanc175   625 Aug 16 20:45 /tmp/vibesys-framework-benchmark-1-2.json
-rw------- 1 ryanc175 ryanc175 38329 Aug 25 04:53 /tmp/vibesys-framework-benchmark-1-3.json
-rw-r--r-- 1 ryanc175 ryanc175   625 Aug 16 21:04 /tmp/vibesys-framework-benchmark-2-1.json
```

Two failure modes reproduced through the real parser. Same-operator concurrency, two runs at `(round=1, retry=1)`, run A writes 1000.0 and then reads:

```
--- A.out (run A's recovered stream) ---
__VIBESYS_FRAMEWORK_BENCHMARK_JSON__
{"kind":"hello","protocol":2,"metrics":{"total_ops_per_sec":{"unit":"ops/s","direction":"max"}}}
{"kind":"result","label":"","values":{"total_ops_per_sec":7777.7}}
__VIBESYS_FRAMEWORK_BENCHMARK_JSON_END__
```

Run A recorded 7777.7, which is run B's number. Stale read, evaluator exits 0 without writing (faithful: `vseval.OpenPath("")` returns a discard report with no file and no error, `sdk/vs-evaluator/vseval/report.go:103-106`):

```
markers found: True
REAL PARSER VERDICT -> feedback: None
                      metric   : total_ops_per_sec = 7777.7
```

Both produce well-formed streams the parser accepts. There is no cross-check available: the protocol schema is closed (`records.py:15-19,37-55`, `ConfigDict(extra="forbid", strict=True)`), so a run id cannot be carried in the record stream. `VIBESYS_CANDIDATE_REVISION` is outbound-only and is read by no evaluator in the tree.

No outer concurrency lock exists. The only `fcntl.flock` is `sandbox/modal_evaluator.py:80-88`, scoped per-container. `create_run` does a bare existence check; `set_current_run` is last-writer-wins.

Docker and Modal are unaffected: each gets a per-container `/tmp` under a `vibesys-{uuid4}` name, so collision is impossible there.

### Expected behavior

A parsed benchmark result must provably be the output of that invocation. The path should carry per-invocation identity (run id plus a nonce), the file should be removed before the evaluator runs, and a missing or stale file should be a hard error rather than a silently accepted parse.

### Reproduction

1. Start two VibeSys runs as the same user on the same host, both using the local backend.
2. Let both reach round 1 retry 1 of a framework benchmark.
3. Whichever reads second parses the other's file. Both runs report a plausible, well-formed measurement.

Stale-read variant: run once to populate `/tmp/vibesys-framework-benchmark-1-1.json`, then run again with an evaluator that exits 0 without writing its output file. The previous run's numbers are recorded as current.

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Profiler: none
Python: 3.14.7 (repo .venv)

### Impact

Blocks normal use

### Relevant logs

```shell

```

#284

### Pre-submission checks

AyanBinRafaih:
## Problem

Six related defects let hypothesis resolutions and evolve fitness rest on evidence that was missing, untrusted, misread, or racy:

- #475: a supportive declaration resolved `PROVEN` with no framework measurement at all. `resolve_hypothesis_outcome` trusted the agent's word whenever `official_metric` was absent, so "proven" carried no empirical weight.
- #476: the `hypothesis_id` schema description said "Reuse it across rounds", but the lifecycle validator raises `ValueError` on reuse after resolution, and the loop let that exception kill the whole run. The schema invited the exact input that terminated the run.
- #477: `_read_protocol_benchmark` resolved `metric_direction` from the configured objective or the evaluator's hello declaration, then the caller discarded it and recomputed direction from objectives alone. All four shipped `result_protocol = 2` tasks configure bare metric names without directions, so direction was always `None` and every metric comparison resolved `INCONCLUSIVE`.
- #478: `metric_baseline` demanded an exact parent-commit match among official rounds. With `official_eval_every >= 2` the parent of an official round is provisional (never in the comparable set), so baseline and delta never resolved. The comparable filter also required `perf_unit == metric`, dropping baselines whose headline metric was renamed even when `metrics` still carried the value.
- #479: rounds that never ran the framework benchmark still wrote `official_evaluation=True` with an agent self-reported `perf_metric`, and nothing recorded who produced the number. Downstream consumers (resolution, baselines, measurements, retention) could not tell a framework measurement from a self-report. Separately, the evolve loop silently ignored a declared benchmark result contract: `main.py` never wired it through, and candidate fitness came from the profiler agent's summary even when the task declared a framework-owned benchmark.
- #480: the framework benchmark wrote to the fixed path `/tmp/vibesys-framework-benchmark-{round}-{retry}.json`. Two runs on one host collide (cross-user /tmp is shared, and another user's stale file makes the write fail with permission denied), and a stale file from a crashed prior run could be read back as this run's result.

## Solution

One theme: a metric may only influence resolution, baselines, or fitness when the framework itself produced it, and every consumer can see where a number came from.

- Resolution (#475): `resolve_hypothesis_outcome` never resolves `PROVEN` without a trusted measurement. A supportive declaration with no trusted `official_metric` resolves to the new `HypothesisResolution.UNMEASURED`, distinct from `INCONCLUSIVE` (which now always means a trusted measurement existed but could not decide). With a measurement, the measurement decides as before. `UNMEASURED` is a resolution, not an outcome: it does not count as a failed hypothesis for strategy pruning, and the TUI renders it through the existing sentence-case fallback with no client change.
- Reprompt instead of terminate (#476): the schema description now states the real contract (a resolved hypothesis id is retired and never reused), and `_run_orchestrator_plan` validates the plan before writing any artifacts. The first invalid plan triggers one corrective reprompt naming the violation; a second consecutive invalid plan raises as before. Checkpoint and round labels from the reprompt carry a `-retry` suffix so artifacts stay distinguishable.
- Direction propagation (#477): protocol benchmark parsing moved to `vibesys.loops.gates` (`read_protocol_benchmark`), and `run_benchmark_gate` returns the full `FrameworkBenchmarkOutcome` including `metric_direction` resolved from the configured objective first, then the evaluator's hello declaration. The agent loop stores that direction on the round instead of recomputing it.
- Baseline fallback (#478): `metric_baseline` still prefers the exact parent-commit match, but when the parent is provisional it falls back to the newest trusted official measurement with `round_number <= parent_round`, so comparisons never invert cause and effect. The comparable filter accepts a record when `perf_unit` matches or the metric is present in the record's `metrics` row, and `_baseline_metric_value` reads the row value for a renamed headline metric.
- Provenance (#479): `RoundRecord` gains `perf_provenance: Literal["framework", "implementer"] | None`. The loop stamps `"framework"` when the benchmark gate produced the number and `"implementer"` when the agent self-reported it; `None` marks legacy records, which stay trusted so reprojection does not rewrite historical resolutions. `_trusted_measurement` gates resolution evidence, `HypothesisMeasurement` projection, baseline selection, and retention comparisons: an implementer-reported metric can never prove a hypothesis, seed a baseline, or appear as an official measurement. `official_evaluation` keeps its meaning (the framework gates ran). The field is loop state only; the wire protocol is unchanged (generated protocol bindings regenerate with zero diff).
- Evolve honours the contract (#479): `_run_evolve` now passes `benchmark_result`, `benchmark_result_protocol`, and the benchmark timeout into `run_evolve_loop`. A new `_run_candidate_gates` runs the accuracy gate and, when a contract is declared, the framework benchmark gate for both bootstrap seeds and children; `_candidate_fitness` takes fitness from the gate's outcome (name, value, full metric row) and falls back to the profiler summary only when no contract is declared. A failed benchmark gate fails the candidate with the gate's feedback before any profiling.
- Collision-proof result path (#480): `run_benchmark_gate` writes to `/tmp/vibesys-framework-benchmark-{slug}-{12-hex-nonce}.json`, removes the path (`rm -f --`) before executing, and reads the result back between explicit output markers from the same invocation. A missing or unparseable result is a hard gate failure, never a silent fallback to a stale file. This also replaces a host-local workaround for the cross-user /tmp collision.

### Architecture

```mermaid
flowchart LR
    G[gates.run_benchmark_gate<br/>nonce path, markers,<br/>direction from objective/hello] --> O[FrameworkBenchmarkOutcome<br/>name, value, direction, row]
    O --> A[agent loop round record<br/>perf_provenance=framework]
    S[agent self-report] --> A2[round record<br/>perf_provenance=implementer]
    A --> H[hypotheses.py<br/>_trusted_measurement]
    A2 -. never trusted .-> H
    H --> R[resolve_hypothesis_outcome<br/>PROVEN / FAILED / INCONCLUSIVE<br/>or UNMEASURED without evidence]
    H --> B[metric_baseline<br/>exact parent, else newest<br/>trusted <= parent_round]
    O --> E[evolve _run_candidate_gates<br/>_candidate_fitness]
```

The gate layer (`vibesys.loops.gates`) owns benchmark execution and parsing for both loops; the agent loop owns round records and provenance stamping; `hypotheses.py` owns resolution and only ever consumes trusted evidence; the evolve loop owns fitness and defers to the gate when a contract is declared.

## Verification

Automated: full pytest suite, ruff format and lint, `ty check`, doc links, launcher smoke test, generated-protocol drift check, and the full TypeScript pipeline all pass on top of current `main`. Manual: a real `--cli-provider claude` 2-round run on the queue mpmc task verifying each fix end to end (details below).

### Correctness properties

- A hypothesis never resolves `PROVEN` without a trusted framework measurement; without one a supportive declaration resolves `UNMEASURED`, and `INCONCLUSIVE` always implies a measurement existed.
- `UNMEASURED` is not a failed outcome: it does not feed strategy pruning or hypothesis retirement heuristics.
- A reused `hypothesis_id` costs one corrective reprompt, not the run; two consecutive invalid plans still raise, and no artifacts are written for a plan that failed validation.
- `metric_direction` on a protocol benchmark round always reflects the configured objective when present, else the evaluator's declaration; consumers never recompute it.
- A baseline never postdates the round it is compared against: exact parent commit first, else the newest trusted official measurement at or before the parent round.
- `perf_provenance` is stamped on every new round record; implementer-reported metrics never enter resolutions, measurements, baselines, or retention comparisons; legacy records (`None`) keep their historical resolutions.
- With a declared benchmark result contract, evolve fitness comes only from the framework gate; a failed gate fails the candidate before profiling; without a contract, profiler fitness is unchanged.
- Every benchmark gate invocation reads only a file it created itself: unique nonce path, pre-run removal, marker-delimited readback, and a hard failure when the file is missing or invalid.
- The wire protocol is unchanged: regenerating the TypeScript protocol bindings produces a zero diff.

### Testing

- `uv run pytest tests/loops/agent/test_hypotheses.py tests/loops/agent/test_orchestrate.py tests/loops/evolve/test_evolutionary_loop.py tests/loops/agent/test_prompt_snapshots.py tests/test_schemas.py -q --no-cov`: all pass. New and updated coverage: supportive declarations without measurement resolve `UNMEASURED` (parametrized); implementer-reported metrics never prove or measure; framework provenance still measures and resolves; the provisional-parent chain now resolves via the fallback baseline; untrusted and renamed-metric baseline selection; reused hypothesis id recovers after one reprompt and still raises after two; invalid plans write no artifacts; nonce result path and `rm -f --` prefix asserted by regex on the executed command; hello-declaration direction resolution; benchmark contract owns seed and child fitness in evolve; contract failure fails the candidate before profiling; no contract keeps profiler fitness.
- `uv run python -m pytest -q -n auto --dist loadgroup --cov-report=xml --cov-report=json --cov-fail-under=75`: 5267 passed, 27 skipped, 15 failed. The 15 failures (tests/sandbox/test_modal_evaluator.py and tests/skypilot/test_evaluator_helper.py) are pre-existing shared-host artifacts: those tests use fixed /tmp paths owned by another user on this machine, they fail identically on a pristine checkout of current `main`, and they pass on CI's clean runners. `scripts/check_coverage_floor.py coverage.json` passes (every module over the 40% floor), and `diff-cover coverage.xml --compare-branch=main --fail-under=80` reports 92% patch coverage.
- `./scripts/check_format.sh`, `./scripts/check_lint.sh`, `uv run ty check`, `python3 scripts/check_doc_links.py`, `uv run vibesys --headless --help`: all clean.
- `pnpm --dir clients/backend-client generate:protocol` then `git diff` on the generated files: zero diff (no wire change). `pnpm build:clients`, `check:clients`, `check:ts-architecture`, `bun test` (clients/tui), `biome check`: pass.
- Merge safety: `git merge-tree --write-tree` simulations against current `main` and all four open PRs (#484, #485, #487, #488), pairwise and as sequential chains in both orders: all clean, no conflicts.
- Manual, real provider: a full 2-round `--cli-provider claude` run on the queue mpmc task (TUI attached), exercising every fix live.
  - #475: round 1's hypothesis (lock-free MPMC ring) was nominated and passed the judge with no framework measurement. It resolved `unmeasured`, not `PROVEN`; the TUI renders "Unmeasured" with no measured value through the existing fallback.
  - #476-adjacent reprompt machinery: round 2's third attempt was a corrective retry after the framework rejected an artifact, with `-retry` labels and no run termination.
  - #477: the queue evaluator declares no configured objective direction. The round 2 record stores `perf_direction: "max"` taken from the evaluator's protocol hello declaration, which the caller previously discarded.
  - #479: the round 2 record carries `perf_provenance: "framework"` alongside `official_evaluation: true` and `perf_metric: 8789681.66` (`total_ops_per_sec`), while round 1, which reported no metric, has `perf_provenance: null`.
  - #480: the executed benchmark command was `rm -f -- /tmp/vibesys-framework-benchmark-<slug>-<12-hex-nonce>.json && <evaluator> --json-output <path> && cat <path>` between output markers; the marker-delimited JSON echoed into the gate record is the readback of the nonce file from that same invocation.
  - Bonus: the run also demonstrated the accuracy gate doing its job. Attempt 1 of round 2 failed the official accuracy gate with a genuine candidate bug (a Vyukov sequence-stride aliasing hang at capacity 1). The loop fed the failure back, the implementer root-caused and fixed it, and the retried gates then passed (accuracy pass, benchmark pass at 8.79M ops/s). The second hypothesis resolved `inconclusive` because a trusted measurement existed but no earlier trusted official baseline did in a 2-round run, which is exactly the corrected meaning of `INCONCLUSIVE` versus `UNMEASURED`.
- Prompt budget note: the orchestrator fallback prompt sat 12 bytes under its 13,200-byte budget on `main`; the corrected `hypothesis_id` contract needs ~60 more bytes even after tightening, so the budget rises to 13,300 with a comment recording why.

## Additional findings while solving these issues

- `main.py` never passed the declared benchmark result contract into `run_evolve_loop` at all, so the evolve loop could not have honoured it even if it had the plumbing; the wiring is part of the #479 fix.
- Moving benchmark execution into `vibesys.loops.gates` left a dead private copy of `_publish_subprocess_output` in the agent loop; it is removed here (caught by the type checker).
- The orchestrator fallback prompt byte budget was saturated on `main` (12 bytes of headroom), so any future one-line schema description change would have tripped it; the bump above records the trigger.
- The modal and skypilot evaluator code has the same defect class as #480: `tests/sandbox/test_modal_evaluator.py` exercises a fixed `/tmp/vibesys-modal-evaluator.lock` and the skypilot helper tests use fixed `/tmp/pending-*` paths. On a shared host where another user owns those paths, 15 tests fail with `PermissionError`; the same collision would hit production runs of those evaluators. Left unfixed here to keep this PR scoped to the six issues; worth a follow-up issue applying the same per-invocation nonce treatment.

Closes #475. Closes #476. Closes #477. Closes #478. Closes #479. Closes #480.


AyanBinRafaih:
Split into individual PRs for each issue. Superseded by #532, #533, #534, #535, #536 and #537.

PR #496, opened 2026-08-29
TUI: Reconnect the event stream and discard 2 answer-cases in experiment chat

### LINKED ISSUE #471: TUI: keystrokes are not contained to the focused surface, so the command input loses characters and modals leak keys
### Observed behavior

### Expected behavior

### Reproduction

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
Theme: dark
Backend: cpu
Profiler: none
Provider: cli / claude

### Impact

Blocks normal use

### Relevant logs

```shell
```

#284

### Pre-submission checks

### LINKED ISSUE #472: TUI: event stream never reconnects after a disconnect, so live updates stop for the rest of the session
### Observed behavior

client.subscribe() is called exactly once, in SessionController.start() (clients/tui/src/session-controller.ts:192). The onDisconnect callback (194-204) only calls markEventStreamUnavailable and reportCaughtError to paint a banner. There is no re-subscribe, retry, or backoff anywhere in the controller or the client, and #eventSubscription is set once and cleared only in stop().
  
After any transient drop of the subscribe socket, the TUI shows "Supervision event stream disconnected" and then receives no further round, agent, or experiment events for the remainder of the session, even though the backend keeps producing them and still answers request() RPCs normally. eventStreamAvailable is also never reset to true, so visibleActiveExecutions returns an empty list and the live-activity strip stays blank permanently. The only recovery is to kill and relaunch the TUI.

### Expected behavior

A dropped event stream is transient. The client re-subscribes with backoff, resumes from its last applied sequence so no events are missed or double-applied, clears the banner and the eventStreamAvailable flag on success, and surfaces a persistent error only after repeated failures.

### Reproduction

1. Start a run: uv run vibesys --project /tmp/vibesys-runs/queue-mpmc --task mpmc --profiler none --cli-provider claude --max-rounds 2
  2. Wait for the experiment log to populate and rounds to start streaming.
  3. Interrupt the subscribe connection (the stream socket only, not the backend process).
  4. Observe the disconnect banner, then observe that no further round or agent updates arrive and the live-activity strip stays empty, while commands that use request/response (for example /perf, /help) still work.
  5. Only relaunching restores live updates.

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
  Terminal: 200x50
  Theme: dark
  Backend: cpu
  Profiler: none
  Provider: cli / claude

### Impact

Blocks normal use

### Relevant logs

```shell

```

#284

### Pre-submission checks

### LINKED ISSUE #473: TUI experiment chat: the agent's answer renders twice for a single question
### Observed behavior

Asking one question in the experiment chat produces two identical answer blocks in the transcript. Observed directly in a real run using the claude CLI provider; the two blocks are exact replicas.
  
**Cause:** the chat agent runs inside supervisor.presentation_scope(agent_kind='chat', round_label='experiment-chat') (src/vibesys/context.py:1461-1476). For the CLI provider, AgentLogger.log_text publishes streamed assistant-text deltas on the assistant channel (src/vibesys/agents/callbacks.py:459-473, via _publish to output_sink().agent_output to publish_presentation), tagged agent_kind='chat'. On the client those become chat-thread assistant entries (clients/core-state/src/core-state.ts applyChatEvent:421-423 via eventToTranscriptEntry:639-655). The run then emits a single terminal ChatData event (src/vibesys/server/supervisor.py:598-605) whose full answer is appended as a second assistant "Answer" entry (:629-637).
  
Nothing equates the streamed prose with the final answer. The only dedup, at clients/tui/src/session-controller.ts:685, guards the client's own local fallback append (it is skipped when response.events contains a chat-kind event) and does not touch entries already added from the stream. By contrast, tool-channel chunks do get filtered (chatTypedToolEvents / legacyToolChunk, core-state.ts:409-417), so the assistant channel is the gap. 

### Expected behavior

 One answer per question. Either the streamed assistant chunks render as the answer and the terminal ChatData event reconciles into that same entry, or the streamed chunks are treated as progress and replaced by the final answer. Either way the transcript ends with exactly one answer block.

### Reproduction

1. Start a run with the CLI provider: uv run vibesys --project /tmp/vibesys-runs/queue-mpmc --task mpmc --profiler none --cli-provider claude --max-rounds 2
  2. Press Ctrl+W to focus the chat and ask any question, for example "summarize round 1".
  3. Wait for the answer to complete.
  4. Observe two identical assistant blocks for that one question, one built from the streamed chunks and one labeled "Answer".

### Affected subsystem

CLI and packaging

### Environment

Commit: 3a2fd8b (main)
  Terminal: 200x50
  Theme: dark
  Backend: cpu
  Profiler: none
  Provider: cli / claude

### Impact

Blocks normal use

### Relevant logs

```shell

```

#284

### Pre-submission checks

AyanBinRafaih:
## Problem

Three TUI client defects, all in the keystroke-to-state path:

- #471: `[`, `]`, Left, and Right were handled globally with no `inputIsEmpty()` guard, unlike Enter. Typing `/steer fix arr[0]` silently dropped both brackets and jumped round focus mid-command. The theme picker intercepted only its navigation keys and let every other key leak to the input and global bindings behind it, and the /help and round-detail overlays handled only Escape, so all other keys acted on the UI underneath an open overlay.
- #472: `subscribe()` was called exactly once at startup. The disconnect callback only painted a banner; nothing ever redialed, and `eventStreamAvailable` was never reset. One transient socket drop left the TUI permanently frozen on stale state for the rest of the run.
- #473: every experiment chat answer rendered twice. The streamed assistant deltas fold into one transcript entry keyed by `turnId` (the invocation id), and then the terminal `chat` event, which carries no turn id, appended a second entry with the identical full text under the label "Answer".

## Solution

- Keystroke containment (#471, `keybindings.ts`): `[`, `]`, Left, and Right now fire round navigation only when the command input is empty, the same contract Enter already had; with text present they fall through to the input untouched. The theme picker and the overlays are modal: each handles its own keys (navigation, apply, Escape) and `preventDefault`s everything else, so no key reaches the input, round focus, or any global binding while one is open. Escape on an overlay returns to the live tail.
- Stream reconnect (#472, `session-controller.ts`, `session-model.ts`): on disconnect the controller paints the transport banner as before, then schedules a reconnect on a finite backoff schedule (500 ms, 1 s, 2 s, 4 s, 8 s). Each attempt resubscribes from `CoreState.sequence`, the client's own applied cursor, with no tail, so the server replays exactly the events the client missed; the reducers' existing `sequence <= state.sequence` guard makes any boundary replay idempotent. On success a new `markEventStreamAvailable` clears the flag and retires the transport-scope banner (banners from other scopes are untouched), and the attempt counter resets so the next outage gets the full schedule. If the schedule runs out the banner stays up as the persistent answer. Reconnect is suppressed once the run is terminal, after a stream protocol error, and after `stop()`, which also cancels any pending timer: a server closing the socket after the final event is lifecycle cleanup, not an outage.
- One answer per chat turn (#473, `core-state.ts`): the terminal chat answer now folds over its own streamed entry in place (`foldChatAnswer`). It replaces the last entry when that entry is an assistant entry with a defined turn id, keeps the streamed entry's id so `reconcileChatTranscript`'s replace-by-id stays stable, takes the terminal event's fields (label "Answer", authoritative full text), and drops the turn id to mark the turn closed. The assistant-merge rule in `foldTranscriptEntry` additionally requires a defined turn id, so two complete answers (both with no turn id) can never glue into one block. Answers that were never streamed still append as their own entry.

### Reconnect flow

```mermaid
flowchart LR
    D[onDisconnect] -->|not terminal, not stopped,<br/>no protocol error| B[banner + eventStreamAvailable=false]
    B --> S[backoff timer<br/>500ms..8s, 5 tries]
    S --> R[resubscribe from<br/>CoreState.sequence, no tail]
    R -->|ok| C[clear banner, reset tries,<br/>fold batches at existing history floor]
    R -->|refused| S
    S -->|schedule exhausted| P[persistent banner]
```

A resumed subscription declares no history floor of its own (`history_after_sequence` is 0 on every batch). Taking that literally would mark history complete and silently break scroll-back backfill, so resumed batches keep the floor bookkeeping the boot subscription established and fold at the floor already in state.

## Verification

Automated: the full TypeScript pipeline passes on top of current `main`. Manual: a live `--cli-provider claude` run with the TUI attached, exercising #471 and #473 end to end (details below).

### Correctness properties

- A key pressed with text in the command input always goes to the input; round navigation via `[`, `]`, Left, Right fires only when the input is empty, matching Enter's existing guard.
- An open theme picker or overlay captures every key: its own keys act, everything else is swallowed, and nothing reaches the surfaces behind it. Escape closes and returns to the live tail.
- After a disconnect the client resumes from its own applied sequence: no event is missed and none is applied twice (the reducer skips `sequence <= state.sequence`).
- A resumed subscription never lowers the history floor the boot subscription established, so scroll-back backfill is unaffected by reconnects.
- Reconnect attempts stop exactly when one of these holds: an attempt succeeded (banner cleared, counter reset), the run is terminal, a stream protocol error occurred, the controller stopped, or the schedule is exhausted (banner stays up).
- Exactly one transcript entry exists per chat turn: streamed chunks merge into it, the terminal answer replaces it in place keeping its id, unstreamed answers append once, and two complete answers never merge.
- The prefix (scroll-back backfill) fold and the in-batch fold agree on chat transcripts even when a turn's events straddle the chunk boundary.
- The wire protocol is unchanged: regenerating the TypeScript protocol bindings produces a zero diff.

### Testing

- New unit tests. Keybindings (`app.test.ts`, 3 tests): typing and Enter are contained while the theme picker is open (the typed command is discarded, the highlighted theme applies); an open overlay swallows `[`, Left, and a typed command, and Escape restores the live tail with the input untouched; brackets and cursor keys stay in a non-empty command input, and `[` still navigates rounds once the input is empty. Reconnect (`session-controller.test.ts`, 4 tests with a `ReconnectTransport` fake): resume from the client's own cursor with no tail, idempotent boundary replay, banner cleared, and attempt counter reset for the next outage; the history floor survives a resumed stream that declares floor 0; an exhausted schedule stops dialing and leaves the transport banner; no reconnect after a terminal event or after `stop()`. Chat fold (`core-state.test.ts`, 2 tests): the final answer folds over its own streamed chunks keeping the streamed id; multiple turns, including an unstreamed answer landing between two streamed turns, stay separate entries, asserted identically through `reduceEventBatch` and sequential `reduceEvent`.
- Two pre-existing tests asserted the old behavior and were updated: one app test pinned the theme picker's Enter leak (now asserts containment), and one prefix test pinned consecutive chat answers concatenating into one entry (now asserts two entries; the boundary-consistency property it exists for still holds and is also asserted by the new dual-path test).
- Full pipeline: `pnpm check:ts` (biome), `pnpm check:ts-architecture`, protocol drift (`generate:protocol` then `git diff`: zero), `pnpm check:clients` (tsc), `pnpm test:clients` (backend-client 16, core-state 66, tui 474, 0 failures), `pnpm build:clients`, and both CI TUI package smokes (packed tarball self-test and deployed workspace self-test) all pass.
- Merge safety: the branch is based on the current `main` tip, and of the open PRs (#484, #485, #487, #488, #495) only #488 touches a shared file (`session-controller.ts`) in functions this PR does not modify, so either merge order is clean.
- Manual, real provider: a live `--cli-provider claude` run on the queue demo task with the TUI attached.
  - #471: `/steer fix arr[0]` renders both brackets in the input and Left moves the cursor without touching round focus; with the input empty `[` still navigates rounds. The open theme picker swallows typed letters and closes on Escape. The open /help overlay swallows `[`, Left, and a typed command; Escape returns to the live tail with the command input still empty.
  - #473: two questions asked in experiment chat each produced exactly one Answer block, and the second answer stayed its own block. Before the fix the terminal event appended a duplicate of the streamed text, and consecutive answers merged into one block. No duplicate appeared after later event-driven reconciliations either.
  - #472 is covered by the unit suite; a mid-run transport drop is not stageable in the manual harness.

## Additional findings while solving these issues

- Pre-existing bug found and fixed alongside #473: consecutive chat answers glued into a single transcript entry, because the assistant-merge rule treated two missing turn ids as equal. This affected any thread with more than one answer, independent of the duplication bug. The tightened rule is safe beyond chat: streamed entries always carry a turn id, so only complete answers are affected.
- Resume semantics subtlety worth recording: the server answers a no-tail subscription with `history_after_sequence` 0 on every batch, meaning "I replayed from your cursor", not "history is complete". A naive reconnect that trusted it would silently break scroll-back backfill; with a tail instead, it would trigger the rebootstrap path and discard live state. The `#resumedStream` flag exists to route around both, and a dedicated test pins the floor's survival.
- The disconnect callback also fires when the server closes the socket after a finished run. Reconnecting then would redial a finished run until the schedule ran out and replace the run's terminal status with a spurious outage banner, which is why terminal state suppresses the whole path.

Closes #471. Closes #472. Closes #473.

### DIFF
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -153,11 +153,12 @@ describe('prefix merges across the chunk boundary', () => {
         model: 'opus',
       },
     ]);
-    // Consecutive chat answers carry no turn id, so the fold concatenates them
-    // exactly as it does inside one batch. The point is that it still does when
-    // the two answers land on opposite sides of the boundary.
+    // Each chat answer is a complete turn and stays a separate entry. The point
+    // is that the prefix fold agrees with the in-batch fold even when the two
+    // answers land on opposite sides of the boundary.
     expect(merged.chatTranscripts['thread-a']?.map(entry => entry.content)).toEqual([
-      'first answersecond answer',
+      'first answer',
+      'second answer',
     ]);
   });
 
--- a/clients/core-state/src/core-state.test.ts
+++ b/clients/core-state/src/core-state.test.ts
@@ -400,6 +400,49 @@ describe('core state projection', () => {
     expect(state.transcript).toEqual([]);
   });
 
+  it('folds the final chat answer over its own streamed chunks', () => {
+    let state = initialCoreState();
+    state = reduceEvent(state, chatStreamEvent(1, 'The queue ', 'exec-1'));
+    state = reduceEvent(state, chatStreamEvent(2, 'is lock-free.', 'exec-1'));
+    state = reduceEvent(state, chatAnswerEvent(3, 'The queue is lock-free.'));
+
+    // One answer block, under the streamed entry's id so consumers tracking
+    // entries by id update in place, closed to any further folding.
+    expect(state.chatTranscript).toHaveLength(1);
+    expect(state.chatTranscript[0]).toMatchObject({
+      id: '1',
+      kind: 'assistant',
+      label: 'Answer',
+      content: 'The queue is lock-free.',
+    });
+    expect(state.chatTranscript[0]?.turnId).toBeUndefined();
+  });
+
+  it('reconciles each chat turn separately and appends unstreamed answers', () => {
+    const events = [
+      chatStreamEvent(1, 'first ', 'exec-1'),
+      chatStreamEvent(2, 'answer', 'exec-1'),
+      chatAnswerEvent(3, 'first answer'),
+      chatAnswerEvent(4, 'unstreamed answer'),
+      chatStreamEvent(5, 'second answer', 'exec-2'),
+      chatAnswerEvent(6, 'second answer'),
+    ];
+    const batched = reduceEventBatch(initialCoreState(), events);
+    let single = initialCoreState();
+    for (const item of events) single = reduceEvent(single, item);
+
+    // Both fold paths agree: a finalized turn cannot swallow the next answer,
+    // and an answer that never streamed still lands as its own entry.
+    for (const state of [batched, single]) {
+      expect(state.chatTranscript.map(item => [item.id, item.content])).toEqual([
+        ['1', 'first answer'],
+        ['4', 'unstreamed answer'],
+        ['5', 'second answer'],
+      ]);
+      expect(state.chatTranscript.every(item => item.turnId === undefined)).toBe(true);
+    }
+  });
+
   it('replays the thread list from creation events after the implicit default', () => {
     let state = initialCoreState();
     state = reduceEvent(state, threadCreatedEvent(1, 'thread-a', 'claude'));
@@ -796,6 +839,15 @@ function chatAnswerEvent(sequence: number, answer: string, threadId?: string): R
     round_label: 'experiment-chat',
     ...(threadId === undefined ? {} : {chat_thread_id: threadId}),
     data: {kind: 'chat', answer},
+  };
+}
+
+/** One assistant-channel chunk of a chat turn, as the chat agent streams it. */
+function chatStreamEvent(sequence: number, content: string, invocationId: string): RunEvent {
+  return {
+    ...outputEvent(sequence, content, invocationId),
+    agent_kind: 'chat',
+    round_label: 'experiment-chat',
   };
 }
 
--- a/clients/core-state/src/core-state.ts
+++ b/clients/core-state/src/core-state.ts
@@ -685,7 +685,7 @@ function applyChatEvent(
   }
   const entry = eventToTranscriptEntry(event);
   if (entry === null || (entry.kind !== 'assistant' && entry.kind !== 'result')) return next;
-  return appendChatTranscript(next, threadId, entry, folder);
+  return appendChatTranscript(next, threadId, entry, folder, data?.kind === 'chat');
 }
 
 /** Registers a replayed thread, or refreshes the record it already has. */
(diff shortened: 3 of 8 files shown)


vic-lsh:
@AyanBinRafaih Thanks for the PR. It looks like this addresses 3 issues -- can you break this into 3 smaller PRs (or let me know if this is not feasible)? Thanks!

AyanBinRafaih:
@vic-lsh. I've created 3 separate PR (one per issue). Superseded by PR #524, #525 and #526.

PR #497, opened 2026-08-30
TUI: Per-round design log added in hypothesis drill-down detail view

AyanBinRafaih:
## Problem

The TUI shows what each hypothesis claims and how it scored, but not what the
agent actually did to the system under optimization. To see which files a round
touched, the operator has to leave the TUI and walk the run branch with git by
hand, mapping round checkpoints to commits themselves. There is no per-round
view of design changes anywhere in the client.

Original implementation by @AyanBinRafaih. A maintainer pass reworked the
contract, the git path, and the surface; the findings it addressed are listed
in a comment on this PR.

## Solution

A server-derived per-round design log, surfaced in two existing places with no
new pane type:

- `/design` renders a summary in the right pane, beside the current view, the
  way `/perf` does: one heading per round (`Round N · hypothesis-id · title`,
  plus the measured perf when recorded) and one file line
  (`+2 ~1  a.rs, b.rs, +3 more`).
- The hypothesis drill-down (Enter on a row, `/open-round`) gains a
  `ROUND N CHANGES` section that follows the round cursor: the full changed-file
  list (`+` added, `-` deleted, `→` renamed with the old name, `~` modified).

Key choices:

- **The design query publishes only what the experiment log does not have.**
  `DesignRound` is `round`, `commit`, `files`. The two round-level facts that
  were genuinely missing, `judge_verdict` and `perf_delta_pct`, are now on
  `HypothesisRound`, so every per-round fact crosses the protocol exactly once
  and clients join the two answers by round number.
- **File lists are cached by commit range.** A round's `(base, head)` range is
  immutable once the checkpoint exists, so nothing invalidates the cache; it is
  bounded rather than expired. The service holds one projection per attached
  run.
- **The diff goes through `GitTracker`.** It is the module that owns git for a
  run: it strips inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`, injects
  `safe.directory`, and logs failures.
- A hypothesis's first round diffs from its recorded `parent_commit` (correct
  under orchestrator reverts); later rounds diff from the same hypothesis's
  previous round (correct for resumed hypotheses); hypotheses without a parent
  fall back to the chronologically previous checkpoint, anchored at the
  manifest's `trusted_input_baseline`.
- Framework bookkeeping is filtered out so the list shows only the system under
  optimization. The path set is derived from the modules that own those
  layouts, not restated: `issue_board.framework_memory_paths` for both memory
  layouts and their artifact and Pareto roots, and `vs_project`'s
  `is_project_state_path` for generated run state. A rename is filtered on both
  its old and its new path.
- An unresolvable commit range yields `files: null` ("File changes are not
  recorded"), distinct from `[]` ("No workspace files changed").

### Architecture

The backend derives; the client joins and lays out.

Two queries answer about the same round, so exactly one of them owns each fact.
`query.experiments` owns every stage fact (`passed`, `reviewed`,
`judge_verdict`, `hypothesis_outcome`, `official_evaluation`,
`candidate_disposition`, `perf_metric`, `perf_unit`, `perf_delta_pct`,
`commit`). `query.design` owns the file list and nothing else. The client joins
them by round number in `designRoundViews`, which is why the pane and the
drill-down cannot state a round differently.

The design projection caches one `git diff --name-status` per `(base, head)`.
Both endpoints are immutable checkpoints, so a cached range stays correct
forever; the cache is bounded (512 ranges, oldest evicted) rather than
invalidated, and lives on `RunApi` keyed by the attached run so it survives
across requests. Failures are deliberately not cached, so a transient git fault
does not turn into a permanent blank.

```mermaid
flowchart LR
  A[AgentRunState rounds] --> E[experiments.py<br/>every stage fact]
  A --> D[design.py DesignLog]
  T[GitTracker.diff_name_status<br/>read-only, timed, logged] --> D
  D -->|cache by base,head| D
  E --> S[service.py]
  D --> S
  S --> C[session-model<br/>designRoundViews join by round]
  C --> P[/design right pane/]
  C --> V[drill-down ROUND N CHANGES]
```

Files:

- `src/server/api/protocol.py`: `DesignQuery` (`query.design`), a narrowed
  `DesignRound`, `DesignFileChange`, `Response.design` / `design_ready` (same
  bootstrap-versus-empty convention as `experiments_ready`), and
  `HypothesisRound.judge_verdict` / `perf_delta_pct`. `hypothesis_outcome` and
  `candidate_disposition` are now the repo's own StrEnums, so the generated
  TypeScript union is closed.
- `src/server/api/design.py` (new): `DesignLog`, the projection plus its cache.
- `src/server/api/experiments.py`: projects the two new round fields and reads
  the two enum fields.
- `src/server/api/service.py`: dispatches `DesignQuery`, owns the per-run
  projection, and journals git failures once per distinct message.
- `src/vibesys/run/git_tracker.py`: `diff_name_status(base, head)`, read-only,
  timeout-bounded, rejecting anything that is not a plain object name.
- `src/vibesys/loops/agent/issue_board.py`: `framework_memory_paths`.
- `clients/tui/src/session-model.ts`: `designLog` state, `designRoundFor`, and
  the `designRoundViews` join; `PaneView` gains `design`.
- `clients/tui/src/session-controller.ts`: `#requestPane` dispatches by view;
  the design refresh single-flights on its own rather than inside the
  experiments promise.
- `clients/tui/src/ui/design-log.ts` (new): pure formatting over the joined
  view.

### Correctness properties

- Every per-round fact has one publisher. The design query cannot restate a
  stage fact, because `DesignRound` has no field for one.
- A round's file list covers exactly the range that round produced: first round
  of a hypothesis from its parent checkpoint (verified under revert by a
  dedicated test), later rounds from the same hypothesis's previous round.
- A cached range is only ever a range whose two endpoints are recorded
  checkpoints, which are immutable; a failed diff is retried rather than cached.
- `files: null` (range unresolvable) is never conflated with `files: []`
  (nothing outside framework bookkeeping changed).
- Framework bookkeeping paths never appear in a change list, on either side of
  a rename.
- Unvalidated commit strings from run state never reach the git argument
  vector: the projection filters them, and `GitTracker.diff_name_status`
  rejects them again.
- `hypothesis_outcome` and `candidate_disposition` are closed sets across the
  boundary. A value from neither vocabulary projects as "not recorded" rather
  than failing the whole log.
- Boot does not wait on git. The design refresh is started by the experiments
  answer but is not awaited by it, so the landing view's readiness is unchanged
  by this PR. (The previous revision awaited it inside `#requestExperiments`,
  which is inside the boot barrier; that is fixed here.)
- Before a run attaches, `/design` explains itself instead of showing an empty
  log; the drill-down shows no design section until the log has loaded, so
  stale or missing data degrades silently.

### Testing

- `uv run pytest tests/server -q --no-cov`: 252 passed.
- `uv run pytest tests/vibesys/test_git_tracker.py tests/vibesys/loops/agent -q --no-cov`:
  352 passed.
- `./scripts/check_format.sh`, `./scripts/check_lint.sh`,
  `./scripts/check_types.sh`, `uv run python scripts/check_doc_links.py`: clean.
- `pnpm --dir clients/backend-client generate:protocol` then `git status`: no
  drift; regeneration is idempotent.
- `clients/tui`: `bun test --max-concurrency 1 src`: 509 pass, 0 fail.
- `pnpm check:clients`, `pnpm check:ts`, `pnpm check:ts-architecture`: clean.
- Performance, 60-round repository, `query.design` end to end:
  uncached ~585ms per request (three runs: 624ms, 576ms, 555ms); with the cache
  the first request after attach costs the same ~545ms and every later one
  ~0.3ms.
- New tests: a non-hex commit never reaching git and `diff_name_status`
  rejecting a revision expression; the `OSError` and `TimeoutExpired` paths and
  a nonzero exit, each asserting the failure was logged; a rename whose old path
  is framework-owned; cache hit, cache miss after eviction, and retry after a
  failed range; the closed-enum projection including a retired value; and a
  `createTestRenderer` test at 100x30 with 10 rounds asserting the newest round
  is on screen, which the fixed-height modal this replaces could not show.


vic-lsh:

vic-lsh:
Thanks @AyanBinRafaih — the derivation here was the hard part and it is right: diffing a hypothesis's first round from its recorded `parent_commit` rather than the chronologically previous checkpoint is the detail that makes this correct under reverts, and the `files: null` versus `[]` distinction is the right contract. I took the branch and pushed a maintainer pass on top of your commit rather than asking for another round of changes. Here is what changed and why.

**[P2] `DesignRound` duplicated the experiment log.** Twelve of its fifteen fields (`hypothesis_id`, `title`, `claim`, `task`, `passed`, `hypothesis_outcome`, `official_evaluation`, `candidate_disposition`, `perf_metric`, `perf_unit`, and the two below) already crossed the protocol on `HypothesisRound`. The drill-down rendered the same round's judge state from two fetches, and they disagreed: the round row said "Judge pending" while the design line beside it said "Judge deferred". The design query now returns only what the experiment log does not carry (`round`, `commit`, `files`). The two round-level facts that were genuinely missing, `judge_verdict` and `perf_delta_pct`, moved onto `HypothesisRound`, so the experiment log is the single source and the client joins the two answers by round number (`designRoundViews`). The row itself now reads its verdict from `judge_verdict` instead of re-deriving one from `reviewed` and `passed`, which is where the contradiction came from.

**[P2] `query.design` ran one uncached git subprocess per round per request.** On the shared request socket, refiring on every `experiments_changed` (twice per round). Each per-round diff is immutable once its checkpoint exists, so `DesignLog` now caches by `(base, head)` and the service keeps one projection per attached run. Measured over a 60-round repository: ~585ms per request before (three runs: 624/576/555ms), ~545ms for the first request after attach and ~0.3ms for every later one. The cache is bounded at 512 ranges rather than invalidated, since checkpoints do not change. Failures are deliberately not cached, so a transient git fault does not become a permanent blank.

**[P2] The git call bypassed `GitTracker`.** It now goes through a new read-only `GitTracker.diff_name_status(base, head)`, so it gets the stripped inherited `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE`, the injected `safe.directory`, and logged failures instead of collapsing to `files: None` in silence. `GitTracker.run` grew an optional `timeout`. Failures reach the run journal once per distinct message.

**[P2] The `/design` overlay clipped the newest rounds.** Up to 24 lines went into a fixed 60%-height box with no scroll, and at 100x30 the footer overpainted content. `clients/tui/README.md` says visualizations render in the scrollable right pane, not the modal, so `/design` now goes through the same `paneView` mechanism as `/perf` and the separate `designView` flag in `commands.ts` is gone — one mechanism, and `#requestPane` dispatches by view. Nothing is dropped to fit a box any more, so the summary lists every round instead of the last 10. There is a `createTestRenderer` regression test at 100x30 with 10 rounds asserting the newest round is on screen.

**[P3] The framework-path filter was a hand-copied frozenset.** It is now derived: `issue_board.framework_memory_paths` covers both memory layouts and their artifact and Pareto roots, and `vs_project.is_project_state_path` covers generated run state. A rename is filtered on both its old and its new path, which the frozenset version missed.

**[P3] `hypothesis_outcome` and `candidate_disposition` crossed as bare `string`.** They now use the repo's StrEnums so the generated TypeScript union is closed. Worth knowing: a round record can hold either vocabulary, the implementer's `HypothesisOutcome` or the framework's `HypothesisResolution`, so the protocol field is the union of both. A value from neither projects as "not recorded" rather than failing the whole log.

**[P3] Test and documentation gaps.** Added: a non-hex commit never reaching git and `diff_name_status` rejecting a revision expression, the `OSError`/`TimeoutExpired`/nonzero-exit paths each asserting the failure was logged, a rename whose old path is framework-owned, cache hit / eviction / retry-after-failure, and the renderer test above. `/design` is in the README command table.

**Body.** Fixed the file paths (they are `src/server/api/*`, not `src/vibesys/server/*`) and dropped the "boot behavior unchanged" claim, which was not true: the design fetch was awaited inside `#requestExperiments`, which is inside the boot barrier, so boot waited on git. It now single-flights on its own and is started, not awaited, by the experiments answer.

**#529 and #539 will need a rebase onto this shape.** What changed for them:

- `DesignRound` lost `hypothesis_id`, `title`, `claim`, `task`, `passed`, `hypothesis_outcome`, `judge_verdict`, `official_evaluation`, `candidate_disposition`, `perf_metric`, `perf_unit`, and `perf_delta_pct`. #539 reads `perf_delta_pct` off the design record through `designRoundFor`; it should read `HypothesisRound.perf_delta_pct` from the experiment log instead, or go through `designRoundViews`, which joins the two.
- `ParsedCommand.designView` is gone. `/design` now parses to `{request: {type: 'query.design'}, paneView: 'design'}`, so in #529's registry it is an ordinary pane command alongside `/perf` rather than a special case.
- `PaneView` is now `'perf' | 'design'` and `PANE_TITLES` has a `design` entry.
- `HypothesisRound` gained `judge_verdict` and `perf_delta_pct`, and `hypothesis_outcome` / `candidate_disposition` are closed unions rather than `string | null`.

I have not marked this ready or merged it; it is on the same branch, so nothing about your PR's identity changed.

PR #500, opened 2026-08-30
fix(tui): append unit to absolute metric in measured column (#465)

### LINKED ISSUE #465: TUI experiment log: Measured column shows a bare, unlabeled number - add unit/metric/direction so the value is interpretable
### Problem

In the default experiment-log landing view, the "Measured" column shows a bare number like 55434.2 with no unit, metric name, or direction, so a reader can't tell what it represents (ops/sec? ms? is higher better?).
  
Grounding: formatMeasured() (clients/tui/src/ui/experiment-log.ts:635-643) returns ±N% when perf_delta_pct is set, else the absolute perf_metric via trimNumber (experiment-log.ts:719-721), else —. The header is just "Measured", padded to MEASURED_WIDTH = 11 (experiment-log.ts:515,545), with no unit. The bare number is the absolute headline metric (for mpmc, the median total_ops_per_sec), not a ratio or normalized score.

Metadata to interpret it exists but isn't shown: 
- Unit — perf_unit is already carried to the TUI on HypothesisEntry (clients/backend-client/src/generated/protocol.generated.ts:836) and is even rendered in the drill-down detail (roundMetadata, experiment-log.ts:665-668), but formatMeasured drops it in the column.
- Metric name (total_ops_per_sec) and direction (max/min) live on the domain model HypothesisMeasurement (src/vibesys/loops/agent/model.py:52,55) but are not projected into the HypothesisEntry DTO by _entry() (src/vibesys/server/experiments.py:50-52) and have no protocol field, so the TUI can't display them. (The objective defines them: .../tasks/mpmc/OBJECTIVE.md:3 — "total_ops_per_sec (maximize)".)
- Baseline value also lives on the model (baseline_value, model.py:58) but only delta_pct is projected, so the column can show +12% yet has no absolute reference.

Image

### Desired outcome

The Measured column is self-describing: the number carries its unit, and the metric and its direction (higher- vs lower-is-better) are discoverable, so a reader can interpret a value without prior knowledge of the task. Delta percentages stay, but the raw value's meaning is no longer lost.

### Acceptance criteria

- [ ] When the column shows an absolute value, it includes the unit (e.g. 55434.2 ops/s), reusing the perf_unit already on HypothesisEntry.
- [ ] The metric name and direction (max/min) are projected from HypothesisMeasurement through the protocol into HypothesisEntry and surfaced to the user (in the column, header, or a legend/detail, as width allows), so it's clear what the metric is and which direction is better.
- [ ] ±N% delta values remain; where space allows, the absolute value + unit stays discoverable (the delta shouldn't fully hide the raw number).
- [ ] Direction is not conveyed by color alone (glyph/word), legible in all themes.
- [ ] Degrades gracefully at narrow widths: the existing responsive drop order (Kept→Claim→Measured, Measured drops at body width 62) and column alignment are preserved.
- [ ] "—" still shown when there is no measurement.
- [ ] TUI build/lint/tests pass, including the experiment-log formatting tests.

### Scope and non-goals

In scope:
- Rendering the unit in the Measured column (client-only — formatMeasured already has entry.perf_unit).
- Plumbing metric name + direction (and optionally baseline_value) from HypothesisMeasurement through _entry() and the protocol into HypothesisEntry, then surfacing them.
  
Not in scope:
- Changing how metrics are measured or how objectives are defined.
- Adding a hard-coded success threshold (there is no 1.4x target field in the data; that figure was an agent plan note, not a metric).
- Redesigning the rest of the log table.

- Unit is a client-only change; metric-name/direction/baseline require adding fields to the generated protocol (protocol.generated.ts HypothesisEntry) and the server projection _entry() (experiments.py), then regenerating (pnpm --dir clients/backend-client generate:protocol).
- The cell is narrow (MEASURED_WIDTH = 11); a long unit or the full metric name may not fit inline — consider a header suffix, a per-panel legend, or the detail view for the full name, keeping the cell as number+unit.
- Preserve existing breakpoints/drop order and trailing-space border alignment.

### Verification approach

- Unit tests for formatMeasured (extend experiment-log.test.ts): absolute value with unit, delta %, missing measurement, narrow-width drop.
- Protocol round-trip: regenerate protocol; server test that a hypothesis with metric/direction/baseline projects the new fields.
- Manual on a real --cli-provider claude mpmc run: confirm the cell reads e.g. 55434.2 ops/s and the metric/direction is discoverable; width sweep for graceful drop; theme sweep so any direction glyph is legible without relying on color.

Parent #284

zatchbell1311-wq:
## Problem
`formatMeasured` returned a bare unlabeled number (e.g. `9337789`) when no delta was available, giving no indication of what the number meant. The unit was already present on `HypothesisEntry.perf_unit` and used correctly in `roundMetadata` but not in the hypothesis-level fallback path.

## Solution
Append `entry.perf_unit` to the absolute metric return in `formatMeasured`, matching the pattern already used in `roundMetadata`. A test case is added for the unit-bearing path.

### Testing
`bun test --cwd clients/tui`: 141 pass, 16 fail. The 16 failures are pre-existing on `main` (stale `core-state` dist, unrelated to this change). New assertion: `formatMeasured` with `perf_unit: 'ops/s'` returns `'2412.5 ops/s'`.

Closes #465.

### DIFF
--- a/clients/tui/src/ui/experiment-log.test.ts
+++ b/clients/tui/src/ui/experiment-log.test.ts
@@ -119,7 +119,9 @@ describe('experiment log rows', () => {
   it('falls back from a delta to an absolute metric and then to a placeholder', () => {
     expect(formatMeasured(entry({perf_delta_pct: -2}))).toBe('-2.0%');
     expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: 2412.5}))).toBe('2412.5');
-    expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: 2412.5, perf_unit: 'ops/s'}))).toBe('2412.5 ops/s');
+    expect(
+      formatMeasured(entry({perf_delta_pct: null, perf_metric: 2412.5, perf_unit: 'ops/s'})),
+    ).toBe('2412.5 ops/s');
     expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: null}))).toBe('—');
   });
 
--- a/clients/tui/src/ui/experiment-log.test.ts
+++ b/clients/tui/src/ui/experiment-log.test.ts
@@ -119,6 +119,7 @@ describe('experiment log rows', () => {
   it('falls back from a delta to an absolute metric and then to a placeholder', () => {
     expect(formatMeasured(entry({perf_delta_pct: -2}))).toBe('-2.0%');
     expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: 2412.5}))).toBe('2412.5');
+    expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: 2412.5, perf_unit: 'ops/s'}))).toBe('2412.5 ops/s');
     expect(formatMeasured(entry({perf_delta_pct: null, perf_metric: null}))).toBe('—');
   });
 
--- a/clients/tui/src/ui/experiment-log.ts
+++ b/clients/tui/src/ui/experiment-log.ts
@@ -640,7 +640,7 @@ export function formatMeasured(entry: HypothesisEntry): string {
     const sign = delta > 0 ? '+' : '';
     return `${sign}${delta.toFixed(delta >= 10 || delta <= -10 ? 0 : 1)}%`;
   }
-  if (typeof entry.perf_metric === 'number') return trimNumber(entry.perf_metric);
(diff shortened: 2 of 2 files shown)


zatchbell1311-wq:
Closing in favour of #487 which covers #465 more completely — adds perf_metric_name, perf_direction, and perf_baseline_value to the wire protocol, direction glyph in the column header, and full drill-down metadata.

PR #523, opened 2026-09-01
TUI: Persist provider session IDs so a quit run resumes mid-hypothesis

### LINKED ISSUE #520: TUI: persist coding-agent session IDs and implementor state so a fully quit run resumes from where it stopped
### Workstream

Core/runtime

### Problem

Quitting the TUI mid-run throws away everything about the in-flight implementor except its files. Confirmed live and in code: the launcher SIGTERMs the whole backend process group on TUI exit (`clients/tui/src/launcher.ts:208-224`), killing any in-flight codex/claude subprocess, and `--resume` restarts the interrupted round from its beginning with a brand-new CLI session. The provider session ID that would allow true continuation is produced but never consumed: `AgentTurnResult.provider_session_id` (`src/vibesys/agents/contracts.py:138`) is populated by the agentshim and omnigent drivers and stored nowhere (not in the event schema, not in `usage.jsonl`, not in `RoundRecord`, not in `Hypothesis`/`AgentRunState`), so after a restart `CliAgentRunner._agents` and `AgentClient._sessions` are empty, `CliAgentBase.session_id` is `None`, and the first call takes the fresh-command branch (`_agent_cli/cli_agent.py:119-123`) instead of `claude --resume <id>` or `codex exec resume <id>`, even though both CLIs support resuming and the provider transcripts survive on disk. Verified in this session: after kill and `--resume`, the new implementer process ran `codex exec` with no resume argument while the pre-kill thread `01a03fa1-0d7c-...` was simply abandoned, and the implementer re-read the objective and source tree from scratch. Session continuity is requested by the loop (`reuse_session=True, session_key=f"hypothesis:{id}"` at `loops/agent/loop.py:1283-1284`) but that key resolves to an in-memory dict only. Also lost across exit: pause/steer state (`supervisor.py:157-159`; a paused run silently resumes as `running` on attach, supervisor.py:262), queued steer messages, the in-flight `RoundRecord` (constructed only at round completion, loop.py:3392/3499), and the ephemeral session dir. What does persist today: completed rounds in `agent/state.json` (WAL-protected), git snapshots per phase, uncommitted working-tree edits (no reset on resume), the durable event log, chat transcripts, and the run manifest.

### Desired outcome

A run interrupted by a full TUI/backend exit resumes from the same logical point instead of replaying the round. Concretely: the active provider session ID for each run-control agent (keyed by hypothesis/session_key, with model and turn count) is persisted durably at turn boundaries, and on `--resume` the implementor is reconstructed with `--resume <id>`/`exec resume <id>` when the provider and model match and the session is still valid, falling back cleanly to a fresh session (current behavior) when it is not. Implementor attribution becomes part of the persisted record (which driver/provider/model/session implemented each round attempt), so the TUI and the resume path both know who was doing what. Paused state and queued steer messages survive restart: a run paused at exit comes back paused with its steer queue intact rather than silently running.

### Acceptance criteria

- [ ] Provider session IDs are persisted at agent-turn boundaries for run-control executions, keyed so a resumed process can look them up per hypothesis/role.
- [ ] `--resume` after a mid-round kill relaunches the implementor via the provider resume command when valid (verified for codex and claude), and the agent demonstrably retains prior-turn context.
- [ ] Invalid/stale sessions (provider renewal thresholds, deleted rollouts, model change) fall back to a fresh session without failing the run.
- [ ] Round attempt records include implementor attribution (driver, provider, model, session reference).
- [ ] Pause flag and pending steer queue are persisted and restored; a run paused at exit resumes paused and the TUI shows it (with issue 2's status surfacing).
- [ ] Existing resume tests pass; new tests cover kill-mid-implementer, resume-with-session, and resume-with-stale-session.

### Scope and non-goals

In scope: 
`_agent_cli/cli_agent.py` and driver session extraction, `agents/cli_runner.py`/`agents/client.py` session caches, agent-loop state model (`loops/agent/model.py`, `state.py`), supervisor pause/steer persistence, and the resume path in `main.py`/`context.py`. 

Not in scope: 
resuming mid-agent-turn (a turn killed in flight is re-run; only cross-turn conversation continuity is required), the evolve loop's generation-counter reset (worth a separate issue), and sandbox changes beyond exposing the session store path to the driver.

Session IDs are machine-local (provider transcripts live under the invoking user's home), so they belong in the machine-local state namespace, not the portable `agent/state.json`; the store must record provider and model to guard against resuming with a mismatched configuration, and the existing codex thread-renewal logic (`cli_runner.py:435-451`) must invalidate the persisted ID when it rotates sessions. Restoring the paused flag interacts with `attach()` forcing `run_status = "running"` (supervisor.py:262).

### Verification approach

Integration test: start a stub-or-real run, kill the process group mid-implementer turn, resume, and assert the provider was invoked with its resume command and the agent's first response references prior context; plus unit tests for the session store lifecycle (persist, lookup, invalidate on renewal/model change) and pause/steer restore. Manual validation with a real codex run on queue-rs as reproduced here.

#284

AyanBinRafaih:
## Problem

Quitting the TUI mid-run throws away everything about the in-flight implementer except its files. On TUI exit the launcher SIGTERMs the whole backend process group (`clients/tui/src/launcher.ts:208-224`), killing any in-flight codex/claude subprocess, and `--resume` restarts the interrupted round from its beginning with a brand-new CLI session. The provider session ID that would allow true continuation is produced but consumed nowhere: `AgentTurnResult.provider_session_id` is populated by the agentshim and omnigent drivers and stored in no durable place (not the event schema, not `usage.jsonl`, not `RoundRecord`), so after a restart `AgentClient._sessions` is empty, the underlying `session_id` is `None`, and the first call takes the fresh-command branch instead of `claude --resume <id>` or `codex exec resume <id>`, even though both CLIs support resuming and the provider transcripts survive on disk. Session continuity is already requested by the loop (`session_key=f"hypothesis:{id}"`) but that key resolves to an in-memory dict only, so it never survives a process exit. Also lost today: which driver/provider/model implemented each round attempt, because attribution was never part of the persisted `RoundRecord`.

Making a session resumable across processes exposes a second, pre-existing gap. Nothing in `src/` ever produced `SessionDisposition.RESET_REQUIRED`, so `AgentClient` had no way to learn that a driver had silently dropped and restarted the conversation a session key names. The AgentShim session does exactly that in three places (retiring an over-budget Codex thread, retrying a missing Codex rollout, and — once resume is possible — meeting a Claude session the CLI refuses to resume). Without a reported restart the last one is fatal: a refused `claude --resume` surfaces as a bare nonzero exit that propagates out of the first resumed turn and ends the run.

This implements issue #520.

## Solution

A durable, machine-local checkpoint of provider session IDs, plus the typed driver contract that makes adopting one safe.

### Driver contract

Two members carry cross-process resume, both typed, neither probed:

- `AgentCapabilities.provider_session_resume` says whether a driver can adopt a conversation created by an earlier process. `session_reuse` only ever promised reuse within one process.
- `AgentSession.resume_provider_session(session_id) -> bool` offers one checkpoint and returns whether it was **actually adopted**. AgentShim delegates to the agent; the mock adopts and echoes; Omnigent returns `False` and declares `provider_session_resume=False`, because Omnigent 0.10 owns its executor's conversation lifecycle internally and exposes no attach point.

Underneath, `CodingAgent` now declares `session_id`, `supports_session_resume` (set by Codex and Claude Code, the two providers with a resume flag), `resume_from(session_id) -> bool`, and `forget_session()`. That removes the `getattr(session, "seed_provider_session", None)` capability probe, the `hasattr(self._agent, "session_id")` checks, and both `cast("CodexCodingAgent", ...)` writes.

### Drivers report every restart

The AgentShim session returns `SessionDisposition.RESET_REQUIRED` whenever it drops the conversation its session named, so `AgentClient` evicts the live session and clears the checkpoint instead of later claiming continuity with history that no longer exists:

- **Codex thread budget.** Now evaluated *after* the turn rather than before it, so the decision reads the usage the turn just produced and the caller learns about the restart from that turn's result instead of discovering it on the next one. The renewal cadence is unchanged.
- **Missing Codex rollout.** The existing retry now reports the restart it performs.
- **Stale Claude session (new).** A resumed Claude turn that fails is retried once from a fresh conversation. Claude Code reports a refused `--resume` as a bare nonzero exit with no distinguishing message, so any failed *resumed* turn is retried; a fresh turn's failure is the agent's own and propagates unchanged. Timeouts never reach this path (they raise `subprocess.TimeoutExpired`). This is what stops a stale checkpoint from killing a run.

### Typed session keys and scoped persistence

`AgentSessionKey(scope: SessionScope, identifier: str)` replaces free-text keys through `AgentClient`, the runner shims, and the loop's two construction sites. It serializes to the same `"hypothesis:H-01"` strings the map already wrote, so the on-disk form is unchanged, and stored keys are parsed back on load.

Only `HYPOTHESIS` and `CHAT` opt into persistence. `ROLE` does not, and `ROLE` is exactly the fallback scope every unscoped `invoke` lands on, so judge, perf_eval, profiler, and evolve-candidate conversations are never checkpointed and never resumed across restarts. The store drops non-durable keys rather than trusting the caller to remember.

### Checkpoint rules

- A checkpoint is adopted only when the **full session-spec fingerprint** matches — the same comparison in-process reuse makes when it decides whether to evict a live session — not just provider and model. The fingerprint is persisted alongside the ID.
- It is dropped on a driver-reported reset, or when the driver reports it did not adopt the ID.
- It is **kept** through an ordinary turn failure. A timeout or a cancelled run says nothing about whether the conversation is still resumable; only a driver that finds it unusable says so, and it says so with `RESET_REQUIRED`.
- Every store failure (malformed file, schema mismatch, `OSError`) is logged through the run log and swallowed. A broken map degrades this run to no persistence and is overwritten by the next completed turn. A completed turn is never discarded because its checkpoint could not be written.

### Round attribution

`RoundRecord` gains `implementer_driver`, `implementer_provider`, and `implementer_model`, all defaulting to `None` so pre-existing records still load under `extra="forbid"`. They describe the *latest* attempt at the round: a resumed run that re-attempts the same round overwrites the entry, so they are not an append-only audit log. The session key field was dropped (derivable from `hypothesis_id`), and so was the persisted turn count: the Codex thread budget is enforced by the driver from the turns and usage it observes, and a resumed thread's first turn immediately reports its real input-token count, which is the signal the budget actually cares about.

### Wiring

`context.py` builds the store over `RunState.local(RunStateNamespace.AGENT).slot("sessions.json", AgentSessionState)` and hands it the run logger. Evolve candidate clients deliberately get no store: a candidate names no conversation narrower than an agent role, so all its turns land on role-scoped keys that are never checkpointed anyway, and candidates share the parent's run ID and local namespace while running concurrently, so one per-run map would alias them.

### Architecture

Who owns the provider session ID at each moment:

- **The driver**, during a turn. It is the only thing that knows whether the conversation still exists, and the only thing allowed to restart it. It reports a restart through the turn's disposition.
- **The store**, between processes. A checkpoint, not a handle: it is advisory, refusable, and losing it costs a resume, never a turn.
- **The client cache**, within a process. The live handle. A cached session's conversation is always newer than any checkpoint, which is why `resume_provider_session` refuses to overwrite it.

```mermaid
flowchart TD
    Loop["run_agent_loop"] -->|"AgentSessionKey(HYPOTHESIS, id)"| Client["AgentClient.run"]
    Loop -->|"driver / provider / model"| Record["RoundRecord (portable, no session ID)"]
    Client -->|"cache hit: live handle"| Cached["_sessions[key]"]
    Client -->|"cache miss: offer checkpoint"| Resume["_resume_checkpoint"]
    Resume -->|"fingerprint must match"| Store[("DurableSessionStore")]
    Resume -->|"resume_provider_session -> bool"| Session["AgentShim session"]
    Session -->|"codex exec resume / claude --resume"| Provider["CLI provider"]
    Session -->|"restart: budget / missing rollout / stale resume"| Reset["RESET_REQUIRED"]
    Reset --> Client
    Client -->|"reset or refused adoption: clear"| Store
    Client -->|"completed turn: record id + fingerprint"| Store
    Store -->|"machine-local, never snapshotted"| Slot["sessions.json"]
```

### Correctness properties

- **Drivers report every restart.** Whenever the AgentShim session drops the conversation its key names, that turn's result is `RESET_REQUIRED`; the client then evicts the live session and clears the checkpoint. A turn that merely raises is not a restart.
- **A refused resume clears the checkpoint.** If `resume_provider_session` returns `False` — the provider cannot resume, or a live conversation is already attached — nothing will ever adopt that ID, so it is dropped immediately.
- **Only opted-in scopes persist.** `HYPOTHESIS` and `CHAT` are checkpointed; `ROLE`, the fallback for every unscoped call, is not, and the store refuses to write it even if asked.
- **Configuration drift never resumes.** A checkpoint is adopted only when the full session-spec fingerprint matches, so any change that would evict a live session also refuses a checkpointed one. The stale entry is dropped, and the turn writes its own.
- **The store never fails a turn.** Malformed JSON, a schema mismatch, and filesystem errors are logged and degrade to no persistence; the next completed turn overwrites the map.
- **Session IDs stay machine-local.** They live only in the local, never-Git-snapshotted namespace. `RoundRecord` carries driver/provider/model and no session identity at all.
- **On-disk compatibility.** Keys serialize to the same `"<scope>:<identifier>"` strings and are parsed back on load; `RoundRecord`'s new fields default to `None`, so older records still parse under `extra="forbid"`.

### Testing

```
uv run pytest tests/vibesys/agents tests/vibesys/_agent_cli tests/vibesys/test_context.py -q --no-cov  -> 538 passed, 1 skipped (macOS sandbox-exec)
uv run pytest tests/vibesys/loops/agent -q --no-cov                                                   -> 331 passed
uv run pytest tests/server libs/vs-loop-state/tests -q --no-cov                                       -> 308 passed (with tests/vibesys/test_context.py)
./scripts/check_format.sh                                                                             -> 479 files already formatted
./scripts/check_lint.sh                                                                               -> All checks passed
./scripts/check_types.sh                                                                              -> All checks passed
uv run python scripts/check_doc_links.py                                                              -> 326 Markdown files, no broken links
```

New or reworked coverage:

- `tests/vibesys/agents/drivers/test_agentshim_driver.py`: `resume_provider_session` adopts a checkpoint on a cold session, refuses when a conversation is already live, and refuses when the provider declares no resume support; the thread-budget renewal and the heavy-turn renewal each report `RESET_REQUIRED`; the missing-rollout retry and the stale-Claude fallback both retry from a fresh conversation and report the reset; a failed *fresh* Claude turn is not retried.
- `tests/vibesys/_agent_cli/test_codex.py`: the real `CodingAgent.resume_from` / `forget_session`, including a provider without a resume flag refusing to adopt.
- `tests/vibesys/agents/test_session_store.py`: scope durability and key parsing; store lifecycle across instances; a role-scoped key writing nothing at all; a malformed map, a foreign-key-contract map, and an unwritable slot each degrading without raising, with a completed turn surviving a broken store; and the client rules — checkpoint on completion, offer on a cold session, never re-offer to a live one, refuse on fingerprint drift, drop on refused adoption, drop on `RESET_REQUIRED`, keep through a failed turn.
- `tests/vibesys/test_context.py`: the run's client gets a `DurableSessionStore` whose map lands in the machine-local namespace outside the repository, and evolve candidate clients get none. (The author's P1 test, kept.)
- `libs/vs-loop-state/tests/test_vs_loop_state_agent.py`: attribution round-trip and legacy-record loading.

Not covered by the suite: credentialed live `codex exec resume` / `claude --resume` against a real quit-and-restart run.

### Docs

`docs/contributing/agent-drivers.md` documents the new capability and method, which providers can resume, that cross-process resume is AgentShim-only today, and the rule that drivers must report every restart.

### Non-goals (follow-up)

Deferred, tracked under #520 acceptance criterion 5: persisting the pause flag and the pending steer queue so a run paused at exit resumes paused. Also out of scope, per the issue: resuming mid-agent-turn (a turn killed in flight is re-run), the evolve loop's generation-counter reset, and Omnigent executor-thread resume. #484's chat session reuse should be rebuilt on this contract (a `CHAT`-scoped key plus `RESET_REQUIRED` handling) rather than a parallel mechanism.

Closes #520.

### DIFF
--- a/docs/contributing/agent-drivers.md
+++ b/docs/contributing/agent-drivers.md
@@ -16,6 +16,52 @@ driver = "omnigent"
 The `omnigent` package is a base dependency (pinned exactly in
 `pyproject.toml`), so every install carries it; `uv sync` is enough.
 
+## Provider session resume
+
+`AgentClient` keeps one live session per session key and, for keys whose scope
+opts into durability, checkpoints that session's provider conversation ID in
+the run's machine-local state. A resumed process offers the checkpoint to the
+first session it builds for that key, so a quit run continues the
+implementer's conversation instead of replaying the round.
+
+Two contract members carry this:
+
+- `AgentCapabilities.provider_session_resume` says whether a driver can adopt a
+  conversation created by an earlier process. `session_reuse` only promises
+  reuse within one process.
+- `AgentSession.resume_provider_session(session_id) -> bool` offers one
+  checkpoint and returns whether it was adopted. A driver returns `False` when
+  its provider cannot resume, or when the session already holds a live
+  conversation whose history is newer than the checkpoint. A `False` answer
+  tells the client the checkpoint is dead, so it drops it.
+
+Cross-process resume is AgentShim-only today, and within AgentShim only for the
+providers whose CLI has a resume flag: `codex exec resume <thread>` and
+`claude --resume <session>`. Gemini and opencode declare no resume support, so
+they always refuse a checkpoint. Omnigent 0.10 owns its executor's conversation
+lifecycle internally and exposes no attach point, so `OmnigentSession` reports
+`provider_session_resume=False` and always refuses.
+
+### Drivers must report restarts
+
+A driver that drops and restarts the conversation a session names must return
+`SessionDisposition.RESET_REQUIRED` on that turn's `AgentTurnResult`. The client
+then evicts the live session and clears the checkpoint, so nothing later claims
+continuity with history that no longer exists. The AgentShim session reports a
+reset for all three of its restarts:
+
+- retiring an over-budget Codex thread (turn count or heavy-turn usage),
+  evaluated after the turn so the decision reads the usage it just produced;
+- retrying a turn whose Codex thread has no rollout on disk;
+- retrying a resumed Claude turn once from a fresh conversation, because Claude
+  Code reports a refused `--resume` as a bare nonzero exit that would otherwise
+  end the run.
+
+A turn that merely raises is not a restart. Timeouts and cancellations say
+nothing about whether the conversation is still resumable, so the client keeps
+the checkpoint and only a driver-reported reset (or a refused adoption) clears
+it.
+
 ## Mock driver
 
 `driver = "mock"` is test infrastructure. It satisfies the same driver
--- a/libs/vs-loop-state/src/vs_loop_state/agent.py
+++ b/libs/vs-loop-state/src/vs_loop_state/agent.py
@@ -126,6 +126,16 @@ class RoundRecord:
     # run's persisted metric space.
     perf_comparison: MetricComparison | None = None
 
+    # Implementer attribution: which driver/provider/model produced this round
+    # attempt. The provider session ID itself is machine-local and never lives
+    # on this portable record; ``vibesys.agents.session_store`` owns it, keyed
+    # by the round's hypothesis ID. These three fields describe the *latest*
+    # attempt at the round: a resumed run that re-attempts the same round
+    # overwrites the manifest entry, so they are not an append-only audit log.
+    implementer_driver: str | None = None
+    implementer_provider: str | None = None
+    implementer_model: str | None = None
+
     @model_validator(mode="after")
     def _normalize_review(self) -> RoundRecord:
         """Keep the review state expressible only through ``judge_verdict``.
--- a/libs/vs-loop-state/tests/test_vs_loop_state_agent.py
+++ b/libs/vs-loop-state/tests/test_vs_loop_state_agent.py
@@ -66,7 +66,6 @@ def test_implementer_attribution_round_trips_and_defaults_to_none() -> None:
     assert bare.implementer_driver is None
     assert bare.implementer_provider is None
     assert bare.implementer_model is None
-    assert bare.implementer_session_key is None
 
     attributed = RoundRecord(
         round_number=2,
@@ -77,11 +76,10 @@ def test_implementer_attribution_round_trips_and_defaults_to_none() -> None:
         implementer_driver="agentshim",
         implementer_provider="codex",
         implementer_model="gpt-5.6-sol",
-        implementer_session_key="hypothesis:H-01",
     )
     payload = serialize_round_record(attributed)
     assert payload["implementer_provider"] == "codex"
-    assert payload["implementer_session_key"] == "hypothesis:H-01"
+    assert payload["implementer_model"] == "gpt-5.6-sol"
     assert parse_round_record(payload) == attributed
 
 
@@ -89,16 +87,11 @@ def test_legacy_record_without_attribution_still_loads() -> None:
     # A record written before attribution existed has none of the new keys;
     # ``extra="forbid"`` rejects unknown keys, not missing defaulted ones.
     legacy = serialize_round_record(_record(1, "a" * 40))
-    for key in (
-        "implementer_driver",
-        "implementer_provider",
-        "implementer_model",
-        "implementer_session_key",
-    ):
+    for key in ("implementer_driver", "implementer_provider", "implementer_model"):
         legacy.pop(key, None)
     restored = parse_round_record(legacy)
     assert restored.implementer_provider is None
-    assert restored.implementer_session_key is None
+    assert restored.implementer_driver is None
 
 
 def test_parse_round_record_applies_declared_defaults() -> None:
--- a/libs/vs-loop-state/tests/test_vs_loop_state_agent.py
+++ b/libs/vs-loop-state/tests/test_vs_loop_state_agent.py
@@ -61,6 +61,46 @@ def test_public_round_record_codec_round_trips_current_schema() -> None:
     assert parse_round_record(payload) == record
 
 
+def test_implementer_attribution_round_trips_and_defaults_to_none() -> None:
+    bare = _record(1, "a" * 40)
+    assert bare.implementer_driver is None
+    assert bare.implementer_provider is None
+    assert bare.implementer_model is None
+    assert bare.implementer_session_key is None
+
+    attributed = RoundRecord(
+        round_number=2,
+        commit="c" * 40,
+        perf_metric=None,
+        perf_unit=None,
+        passed=True,
+        implementer_driver="agentshim",
+        implementer_provider="codex",
+        implementer_model="gpt-5.6-sol",
+        implementer_session_key="hypothesis:H-01",
+    )
+    payload = serialize_round_record(attributed)
+    assert payload["implementer_provider"] == "codex"
+    assert payload["implementer_session_key"] == "hypothesis:H-01"
+    assert parse_round_record(payload) == attributed
+
+
+def test_legacy_record_without_attribution_still_loads() -> None:
+    # A record written before attribution existed has none of the new keys;
+    # ``extra="forbid"`` rejects unknown keys, not missing defaulted ones.
+    legacy = serialize_round_record(_record(1, "a" * 40))
+    for key in (
+        "implementer_driver",
+        "implementer_provider",
+        "implementer_model",
+        "implementer_session_key",
+    ):
+        legacy.pop(key, None)
+    restored = parse_round_record(legacy)
+    assert restored.implementer_provider is None
+    assert restored.implementer_session_key is None
+
+
 def test_parse_round_record_applies_declared_defaults() -> None:
     record = parse_round_record(
         {
--- a/src/vibesys/_agent_cli/base.py
+++ b/src/vibesys/_agent_cli/base.py
@@ -99,6 +99,22 @@ class CodingAgent(ABC):
     back to the prompt-level schema instruction.
     """
 
+    supports_session_resume = False
+    """Whether this provider's CLI can continue an earlier conversation.
+
+    Set by the providers that override :meth:`CLICodingAgent._get_resume_command`
+    (Codex and Claude Code today). It is the declared capability behind
+    :meth:`resume_from`, so callers never have to probe for a resume flag or
+    branch on the provider name.
+    """
+
+    session_id: str | None = None
+    """The provider conversation the next :meth:`generate` continues.
+
+    ``None`` means the next turn starts a fresh conversation. Concrete CLI
+    agents assign it after each turn from the provider's own transcript ID.
+    """
+
     # Declared (not assigned) here: every concrete provider sets these in its
     # ``__init__``.  ``env`` is the subprocess environment the CLI is spawned
     # with; ``executor`` is the agentshim ``CommandExecutor`` (host, docker,
@@ -274,3 +290,25 @@ class CodingAgent(ABC):
         Idempotent. Default implementation is a no-op.
         """
         return
+
+    def resume_from(self, session_id: str) -> bool:
+        """Continue *session_id* on the next turn, reporting whether it stuck.
+
+        Returns ``False`` when the ID was not adopted, which happens when the
+        provider has no resume flag or when a live conversation is already
+        attached: an in-flight conversation is never replaced, because its
+        history is newer than any checkpoint a caller could hold. Callers use
+        the return value to decide whether the checkpoint is still usable.
+        """
+        if not self.supports_session_resume or self.session_id is not None:
+            return False
+        self.session_id = session_id
+        return True
+
+    def forget_session(self) -> None:
+        """Drop the current conversation so the next turn starts fresh.
+
+        Idempotent. The caller is responsible for telling its own callers that
+        the conversation restarted; this only mutates the agent.
+        """
+        self.session_id = None
--- a/src/vibesys/_agent_cli/claude.py
+++ b/src/vibesys/_agent_cli/claude.py
@@ -60,6 +60,9 @@ class ClaudeCodeCodingAgent(CLICodingAgent[ClaudeGenerationSession]):
     """Coding agent implementation using the Claude Code CLI tool."""
 
     supports_native_output_schema = True
+    # ``claude --resume <session-id>`` continues a stored session, so a
+    # checkpointed session ID can be adopted before the next turn.
+    supports_session_resume = True
     # Claude Code's ``--json-schema`` flag takes the schema *inline*, not a
     # path, so the schema file must be read at command-build time. The runner
     # therefore hands this provider an absolute host path (readable on both the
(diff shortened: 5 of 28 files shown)


vic-lsh:
Review finding:

[P1] Invalidate a persisted session ID when a seeded session fails (`src/vibesys/agents/client.py:471-478`).

The exception path evicts only the live cached session. The durable record remains, so the next attempt creates a fresh session and `_seed_session` injects the same failing ID again. This is concrete for a deleted or otherwise invalid Claude session: Claude has no Codex-style missing-rollout fallback, so every retry and restart can repeat `--resume <stale-id>` indefinitely.

Please clear or otherwise invalidate the durable ID on the relevant failure path, or add provider-specific stale-session recovery. A test should seed a session, make `run_turn` raise, then verify the next attempt does not receive the same seed.

vic-lsh:
Note to self: double check the new feature's code architecture. Is there a cleaner way to encapsulate agent run session IDs?

vic-lsh:

vic-lsh:
@AyanBinRafaih thanks for this — the shape was right and the durability argument (machine-local namespace, never the portable snapshot) was the correct call. I've pushed a rework onto this branch rather than asking for another round, since the changes cross the driver contract. Your three commits are preserved; mine sits on top, rebased onto `main`. The PR body is rewritten to describe the result.

Here is what changed and why, in the review's terms.

**[P1] A stale Claude session killed the run.** `claude --resume <id>` on a session Claude will not resume exits nonzero, which surfaced as a `RuntimeError` out of the very first resumed turn. Codex had a missing-rollout fallback; Claude had none, so the feature turned a recoverable restart into a dead run. The AgentShim session now retries a *resumed* turn once from a fresh conversation. Only resumed turns are retried, and only once, so a real agent failure still propagates; timeouts never reach the path because they raise `subprocess.TimeoutExpired`.

**[P1] Nothing reported that a driver had restarted the conversation.** Nothing in `src/` ever produced `SessionDisposition.RESET_REQUIRED`, yet the AgentShim session silently drops its `session_id` in three places. With a durable checkpoint in play that becomes a correctness bug: the client keeps believing a conversation exists after the driver replaced it. All three now report `RESET_REQUIRED` — the Codex thread-budget renewal, the missing-rollout retry, and the new stale-Claude fallback — so the client evicts and clears the checkpoint. The thread-budget check also moved from before the turn to after it, so the decision reads the usage that turn produced and the caller learns from that turn's result; the renewal cadence is unchanged.

**[P1] `seed_provider_session` was a `getattr` capability probe.** `getattr(session, "seed_provider_session", None)` plus `hasattr(self._agent, "session_id")` plus `cast("CodexCodingAgent", ...)` is exactly the untyped reach-through the coding guide rules out, and it could not express "I was offered an ID and did not take it". Replaced with a typed contract: `AgentCapabilities.provider_session_resume`, and `AgentSession.resume_provider_session(session_id) -> bool` returning whether the ID was actually adopted. `CodingAgent` now declares `session_id`, `supports_session_resume` (Codex and Claude Code), `resume_from`, and `forget_session`, so the probe, the `hasattr` checks, and both casts are gone.

**[P1] A failed seeded turn discarded the checkpoint on any `BaseException`.** A timeout or a cancelled run says nothing about whether the conversation is still resumable, so a slow turn permanently threw away a valid session. The checkpoint is now dropped only when the resume was *refused*: a driver-reported `RESET_REQUIRED`, or the driver reporting it did not adopt the ID. Ordinary turn failures leave it intact.

**[P2] Free-text session keys, and role keys leaking into durable state.** `cache_key = session_key or kind` meant every unscoped call — judge, perf_eval, profiler — landed on a bare-role key that the store happily persisted, so a restart could resume an unrelated role conversation. Replaced with a frozen `AgentSessionKey(scope, identifier)` over a `SessionScope` StrEnum, serializing to the same `"hypothesis:H-01"` strings (on-disk form unchanged, parsed back on load). Only `HYPOTHESIS` and `CHAT` opt into persistence; `ROLE` never does, and the store refuses to write it even if asked. The loop's two f-string sites use the constructor.

**[P2] The seed check compared only provider and model.** In-process reuse evicts on *any* spec change (reasoning effort, skills, MCP servers, policy, workspace), so a checkpoint had a weaker rule than a live session. It now compares a fingerprint of the whole `AgentSessionSpec`, persisted alongside the ID — same rule, same inputs.

**[P2] Store failures could fail a turn.** A malformed `sessions.json`, a schema mismatch, or an `OSError` propagated out of `record()` and would have discarded a turn that had already completed. `DurableSessionStore` now logs every such failure through the run log and degrades to no persistence; the next completed turn overwrites the map. Tested with a corrupt file, a foreign-key-contract file, and an unwritable slot.

**[P2] Wiring.** `context.py` uses `RunState.local(RunStateNamespace.AGENT)` instead of reaching into `project_state.local_namespace(...)`, and there is now a focused wiring test asserting the map lands outside the repository. Evolve candidate clients stay on the null store deliberately, with the reason in a comment: a candidate names no conversation narrower than a role, so all its turns are role-scoped and never checkpointed, and candidates share the parent's run ID and local namespace while running concurrently, so one per-run map would alias them.

**[P3] `RoundRecord` fields.** `implementer_session_key` dropped — it is derivable from `hypothesis_id` through the key constructor, and a field whose value is a function of an existing field is one more place copies can drift. `implementer_driver` / `provider` / `model` stay, with a comment noting these describe the latest attempt: a resumed run re-attempting the round overwrites them, so they are not an append-only audit log. Of the two options for `turn_count`, I dropped the field rather than feeding it back into the session: doing the latter would have widened `resume_provider_session` to carry a counter, and the thread budget already reacts to a resumed thread's real input-token count on its first turn, which is the signal it actually cares about.

**[P3] Docs.** `docs/contributing/agent-drivers.md` now covers the capability, the method, which providers can resume, that cross-process resume is AgentShim-only today, and the rule that drivers must report every restart.

Your P1 test (the store landing in machine-local state) is kept, extended into the wiring test.

One forward note: **#484's chat session reuse should be rebuilt on this contract** — a `CHAT`-scoped `AgentSessionKey` plus `RESET_REQUIRED` handling — once this lands, rather than growing a parallel mechanism for the same problem.

Checks on the pushed head: `tests/vibesys/agents` + `tests/vibesys/_agent_cli` + `tests/vibesys/test_context.py` (538 passed, 1 skipped), `tests/vibesys/loops/agent` (331 passed), `tests/server` + `libs/vs-loop-state/tests` (308 passed), format, lint, types, and doc links all clean.

vic-lsh:
@AyanBinRafaih FYI: #594 follows up on this PR. The round-record attribution it added reads `driver_name`/`provider`/`model_for_kind` off the agent client, which only the CLI `AgentClient` defines, so every `--stub-agent` run (including the release-wheel job's installed-wheel smoke) hit `AttributeError: 'StubAgentClient' object has no attribute 'driver_name'`. #594 makes that attribution part of an explicit client protocol every backend implements, so the gap is a type error rather than a runtime one. No action needed.

PR #524, opened 2026-09-01
TUI: Contain keystrokes for modal surfaces and a non-empty command input

### LINKED ISSUE #471: TUI: keystrokes are not contained to the focused surface, so the command input loses characters and modals leak keys

AyanBinRafaih:
## Problem

Three overlapping keystroke-routing defects in the TUI client, all in `keybindings.ts`, let keys reach the wrong surface. First, `[`, `]`, Left, and Right were handled globally with no `inputIsEmpty()` guard, unlike Enter. Typing a command that contains those characters (for example `/steer fix arr[0]`) silently dropped both brackets and jumped round focus mid-command instead of inserting them into the command input. Second, the theme picker intercepted only its own navigation keys and let every other key leak through to the command input and the global bindings behind it, so typing while the picker was open both edited the hidden input and triggered pane or round actions. Third, the `/help` and round-detail overlays handled only Escape, so every other key acted on the UI underneath an open overlay. In all three cases a key that should have been either inserted as text or swallowed by a modal instead mutated round focus or pane state.

## Solution

Make the guarded bindings and the two modal surfaces obey the same contract Enter already had. `[`, `]`, Left, and Right now fire round navigation and pane focus only when the command input is empty; with text present they fall through to the input untouched, so cursor keys and brackets belong to a non-empty command. The theme picker and the overlays are treated as modal: each handles its own keys (navigation, apply, Escape) and `preventDefault`s everything else, so no key reaches the command input, round focus, or any global binding while one is open. Escape on an overlay returns to the live tail. The change is confined to the key-dispatch order in `bindKeybindings`; no controller state, event flow, or rendering path is touched.

### Architecture

`keybindings.ts` owns the single global key handler and is the only place that decides, per keystroke, whether a key belongs to a modal surface, the command input, or a global binding. The controller and the views are unchanged: they still expose the same actions (`applySelectedTheme`, `closeThemePicker`, `live`, `focusRound`, `selectNextRound`, `selectPreviousRound`) and the same `inputIsEmpty()` predicate the handler consults. The fix reorders and guards the dispatch so a modal short-circuits first, then the empty-input guard gates round and pane navigation, then text falls through to the input.

```mermaid
flowchart TD
    K[keydown] --> M{theme picker or<br/>overlay open?}
    M -->|yes| MK{key the modal<br/>handles?}
    MK -->|yes| A[act: navigate / apply / Escape to live tail]
    MK -->|no| S[preventDefault: swallow<br/>never reaches input or globals]
    M -->|no| E{command input<br/>empty?}
    E -->|no| I[insert into command input:<br/>brackets, Left, Right stay as text]
    E -->|yes| G[global nav: focusRound,<br/>selectPreviousRound / selectNextRound]
```

## Verification

Automated: the TUI TypeScript pipeline (`tsc`, biome) and the TUI test suite pass on top of current `main`; the new assertions fail without the change and pass with it. Manual: a live `--cli-provider claude` run with the TUI attached, exercising the three cases end to end.

### Correctness properties

- A key pressed with text in the command input always goes to the input; round navigation and pane focus via `[`, `]`, Left, and Right fire only when the input is empty, matching Enter's existing guard.
- An open theme picker or overlay captures every key: its own keys act, everything else is swallowed, and nothing reaches the command input, round focus, or any global binding behind it.
- Escape on an open overlay closes it and returns to the live tail with the command input untouched.

### Testing

- New unit tests in `app.test.ts` (3 tests): typing and Enter are contained while the theme picker is open (the typed command is discarded and the highlighted theme applies); an open overlay swallows `[`, Left, and a typed command, and Escape restores the live tail with the input untouched; brackets and cursor keys stay in a non-empty command input, and `[` still navigates rounds once the input is empty.
- One pre-existing test that pinned the theme picker's Enter leak was updated to assert containment.
- `pnpm --filter @vibesys/tui check` (tsc), biome over the changed files, and `pnpm --filter @vibesys/tui test` all pass; the only failing test is the pre-existing environment-specific `launcher` case, which fails identically on unmodified `main`.
- Manual, real provider: a live `--cli-provider claude` run on the queue demo task. `/steer fix arr[0]` renders both brackets in the input and Left moves the cursor without touching round focus; with the input empty `[` still navigates rounds. The open theme picker swallows typed letters and closes on Escape. The open `/help` overlay swallows `[`, Left, and a typed command; Escape returns to the live tail with the command input still empty.

Closes #471.

### DIFF
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -628,6 +628,51 @@ describe('OpenTUI presentation', () => {
     expect(early).toContain('[ r1 ]');
   });
 
+  it('leaves brackets and cursor keys to a typed command', async () => {
+    const testRenderer = await createTestRenderer({width: 120, height: 20});
+    const controller = new FakeController({
+      ...initialSessionState(),
+      core: {
+        ...initialSessionState().core,
+        rounds: [
+          {number: 1, status: 'completed' as const},
+          {number: 2, status: 'active' as const},
+        ],
+        transcript: [
+          {
+            id: 'live',
+            kind: 'assistant',
+            label: 'Agent',
+            content: 'live output',
+            roundNumber: 2,
+          },
+        ],
+      },
+    });
+    const app = createOpenTuiApp(testRenderer.renderer, controller);
+    registerCleanup(testRenderer.renderer, app);
+    await testRenderer.waitForFrame(value => value.includes('live output'));
+    const focusBefore = controller.state.roundFocus;
+
+    // With text in the command input, brackets are characters and the cursor
+    // keys stay in the input: nothing navigates rounds or moves pane focus.
+    await testRenderer.mockInput.typeText('/steer fix arr[0]');
+    testRenderer.mockInput.pressKey('ARROW_LEFT');
+    const typed = await frameAfter(testRenderer);
+    expect(typed).toContain('arr[0]');
+    expect(controller.state.selectedRound).toBeNull();
+    expect(controller.state.roundFocus).toBe(focusBefore);
+
+    testRenderer.mockInput.pressEnter();
+    await testRenderer.waitForFrame(() => controller.submissions.length === 1);
+    expect(controller.submissions).toEqual(['/steer fix arr[0]']);
+
+    // With the input empty again, the same key is round navigation.
+    testRenderer.mockInput.pressKey('[');
+    await frameAfter(testRenderer);
+    expect(controller.state.selectedRound).toBe(1);
+  });
+
   it('filters the transcript to an agent node that is clicked', async () => {
     const testRenderer = await createTestRenderer({width: 150, height: 24});
     const controller = new FakeController({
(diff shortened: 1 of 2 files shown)


vic-lsh:
Review finding:

[P2] Route modal input before the focused docked-chat branch (`clients/tui/src/ui/keybindings.ts:119-148`, with the overlay branch at 169-177).

When docked chat has focus and a global command opens `/theme` or `/help`, `chatPaneFocused()` returns before either modal block. Printable keys can therefore reach the hidden composer, Up/Down can navigate chat suggestions, and Escape focuses the left pane instead of consistently closing the modal. The reported containment bug remains on this common path.

Please move modal routing ahead of focused-pane routing, or otherwise make modal state authoritative, and add coverage for opening each modal while docked chat is focused.

PR #525, opened 2026-09-01
TUI: Reconnect the event stream after a transport drop

### LINKED ISSUE #472: TUI: event stream never reconnects after a disconnect, so live updates stop for the rest of the session

AyanBinRafaih:
## Problem

The TUI subscribed to the event stream exactly once, at startup. The disconnect callback only painted a transport banner; nothing ever redialed, and `eventStreamAvailable` was never reset back to true. One transient socket drop left the TUI frozen on stale state for the rest of the run: active executions and the transcript stopped updating, and the only signal was a banner that never cleared. There was no path back to a live subscription short of restarting the client.

## Solution

On disconnect the controller paints the transport banner as before, then schedules a reconnect on a finite backoff schedule (500 ms, 1 s, 2 s, 4 s, 8 s). Each attempt resubscribes from `CoreState.sequence`, the client's own applied cursor, with no tail, so the server replays exactly the events the client missed; the reducers' existing `sequence <= state.sequence` guard makes any boundary event delivered twice idempotent. On success a new `markEventStreamAvailable` clears the flag and retires the transport-scope banner (banners from other scopes are untouched), and the attempt counter resets so the next outage gets the full schedule again. Reconnect is suppressed once the run is terminal, after a stream protocol error, and after `stop()`, which also cancels any pending timer: a server closing the socket after the final event is lifecycle cleanup, not an outage. If the schedule runs out the banner stays up as the persistent answer.

A resumed subscription declares no history floor of its own (`history_after_sequence` is 0 on every batch). Taking that literally would mark history complete and silently break scroll-back backfill, so resumed batches keep the floor bookkeeping the boot subscription established and fold at the floor already in state. A `#resumedStream` flag routes resumed batches down this path instead of the bootstrap or rebootstrap paths.

### Architecture

`session-controller.ts` owns the subscription lifecycle: it opens the boot stream, detects disconnects, and now schedules and drives reconnects; `session-model.ts` owns the presentational flag and banner and gains `markEventStreamAvailable`, the inverse of the existing `markEventStreamUnavailable`. The reducers in `core-state` are unchanged and already idempotent on sequence, which is what makes resume-from-cursor safe. The wire protocol is unchanged: resume is just a subscribe from a non-zero cursor with no tail.

```mermaid
flowchart LR
    D[onDisconnect] -->|not terminal, not stopped,<br/>no protocol error| B[banner + eventStreamAvailable=false]
    B --> S[backoff timer<br/>500ms..8s, 5 tries]
    S --> R[resubscribe from<br/>CoreState.sequence, no tail]
    R -->|ok| C[clear banner, reset tries,<br/>fold batches at existing history floor]
    R -->|refused| S
    S -->|schedule exhausted| P[persistent banner]
```

## Verification

Automated: the full TUI TypeScript pipeline (`tsc`, biome, architecture cruise) and the TUI test suite pass on top of current `main`; the new reconnect tests fail without the change and pass with it. A mid-run transport drop is not stageable in the manual harness, so #472 is covered by the unit suite.

### Correctness properties

- After a disconnect the client resumes from its own applied sequence: no event is missed and none is applied twice (the reducer skips `sequence <= state.sequence`).
- A resumed subscription never lowers the history floor the boot subscription established, so scroll-back backfill is unaffected by reconnects.
- Reconnect attempts stop exactly when one of these holds: an attempt succeeded (banner cleared, counter reset), the run is terminal, a stream protocol error occurred, the controller stopped, or the schedule is exhausted (banner stays up).
- A successful resubscribe retires only the transport-scope banner; banners raised by other scopes are left in place.

### Testing

- New unit tests in `session-controller.test.ts` (4 tests, driven by a `ReconnectTransport` fake): resume from the client's own cursor with no tail, idempotent boundary replay, banner cleared, and attempt counter reset for the next outage; the history floor survives a resumed stream that declares floor 0; an exhausted schedule stops dialing and leaves the transport banner up; no reconnect after a terminal event or after `stop()`.
- `pnpm --filter @vibesys/tui check` (tsc), biome over the changed files, `pnpm run check:ts-architecture`, and `pnpm --filter @vibesys/tui test` all pass; the four new reconnect tests pass and the only failing test is the pre-existing environment-specific `launcher` case, which fails identically on unmodified `main`.

Closes #472.


vic-lsh:
Review findings:

[P1] Mark a subscription as resumed before its first callback can run (`clients/tui/src/session-controller.ts:378-383`).

`#resumedStream` is set only after `await client.subscribe(...)`, but the production `ServerClient.subscribe` invokes `onMessage` before its promise resolves and supports `subscribed` plus `event_batch` in one socket chunk. The first resumed batch therefore takes the bootstrap branch, consumes the declared floor `0`, and can lower `historyAfterSequence` to zero, disabling scroll-back. The current fake emits after `subscribe` settles and misses this ordering. Please bind resume mode into the callback or set it before dialing, and test synchronous first-batch delivery.

[P2] Close reconnect subscriptions that resolve after `stop()` (`clients/tui/src/session-controller.ts:365-367`).

Once the retry timer fires, `#eventSubscription` is null while `subscribe()` is pending. `stop()` cannot cancel it, and if the promise later resolves the new subscription is stored; `#resumeEventStream` then returns on `#stopped` without closing it. The socket can survive shutdown and may deliver state changes after stop. Please add cancellation/generation ownership and close late subscriptions, with a test that stops during a deferred reconnect handshake.

vic-lsh:
@AyanBinRafaih  Good PR idea. I wonder if we can create an abstraction that provides the semantics of a persistent connection. Handling re-connection is handled entirely within that abstraction. Then the session controller could be a lot simpler, without having to think about reconnecting?

AyanBinRafaih:
@vic-lsh Yes, I think we could do something along those lines. Right now, the reconnection for a session is mixed up with the session controller so I guess we could separate them. 

We could do something like a persistent event stream (which would be in the backend-client) and this could wrap around the subscription (handling everything for persisting the state through the socket such as connection, detecting when it got disconnected etc), and then we could have the controller just call it. We could definitely do this. Let me know what you think!

vic-lsh:
> @vic-lsh Yes, I think we could do something along those lines. Right now, the reconnection for a session is mixed up with the session controller so I guess we could separate them. 
> 
> 
> 
> We could do something like a persistent event stream (which would be in the backend-client) and this could wrap around the subscription (handling everything for persisting the state through the socket such as connection, detecting when it got disconnected etc), and then we could have the controller just call it. We could definitely do this. Let me know what you think!
> 
> 

Sounds great! Let's do that in this PR?

AyanBinRafaih:
@vic-lsh Done in this PR.

PR #526, opened 2026-09-01
TUI: Render one answer per experiment chat turn

### LINKED ISSUE #473: TUI experiment chat: the agent's answer renders twice for a single question

AyanBinRafaih:
## Problem

Every experiment chat answer rendered twice. The streamed assistant deltas of one turn fold into a single transcript entry keyed by `turnId` (the invocation id), and then the terminal `chat` event, which carries no turn id, appended a second entry with the identical full text under the label "Answer". A related pre-existing bug compounded it: consecutive chat answers glued into a single transcript entry, because the assistant-merge rule in `foldTranscriptEntry` treated two missing turn ids as equal, so any thread with more than one answer merged its answers into one block independent of the duplication.

## Solution

The terminal chat answer now folds over its own streamed entry in place (`foldChatAnswer`) rather than appending. It replaces the last entry when that entry is an assistant entry with a defined turn id, keeps the streamed entry's id so `reconcileChatTranscript`'s replace-by-id stays stable, takes the terminal event's fields (label "Answer", authoritative full text), and drops the turn id to mark the turn closed. The assistant-merge rule in `foldTranscriptEntry` additionally requires a defined turn id, so two complete answers (both with no turn id) can never glue into one block. Answers that were never streamed still append as their own entry. The change is confined to the chat-transcript fold in `core-state.ts`; the event contract and every other reducer are unchanged.

### Architecture

`core-state.ts` owns the chat transcript reduction, and both the streamed-chunk fold and the terminal-answer fold live there, so the two paths agree on what one turn is. `appendChatTranscript` now consults `foldChatAnswer` on the terminal answer before falling back to a plain append, on both the streaming (`TranscriptFolder`) path and the direct path; `foldTranscriptEntry` carries the tightened turn-id predicate that keeps two complete answers separate. The server, the streaming protocol, and the TUI views are unchanged: the same events arrive in the same order and one entry per turn now comes out.

```mermaid
flowchart TD
    S[streamed assistant chunks<br/>share turnId] --> E[one open entry keyed by turnId]
    T[terminal chat event<br/>no turnId, full text] --> F{foldChatAnswer:<br/>last entry assistant<br/>with a turnId?}
    E --> F
    F -->|yes| R[replace in place: keep entry id,<br/>take Answer label + full text, drop turnId]
    F -->|no| A[append as its own entry]
    R --> O[exactly one entry per turn]
    A --> O
```

## Verification

Automated: the `core-state` TypeScript pipeline (`tsc`, biome) and the `core-state` test suite pass on top of current `main`; the new assertions fail without the change and pass with it. Manual: a live `--cli-provider claude` run with the TUI attached, exercising the chat path end to end.

### Correctness properties

- Exactly one transcript entry exists per chat turn: streamed chunks merge into it, the terminal answer replaces it in place keeping its id, unstreamed answers append once, and two complete answers never merge.
- The prefix (scroll-back backfill) fold and the in-batch fold agree on chat transcripts even when a turn's events straddle the chunk boundary.
- The wire protocol is unchanged: no event shape or ordering assumption is altered.

### Testing

- New unit tests in `core-state.test.ts` (2 tests): the final answer folds over its own streamed chunks keeping the streamed id; multiple turns, including an unstreamed answer landing between two streamed turns, stay separate entries, asserted identically through `reduceEventBatch` and sequential `reduceEvent`.
- One pre-existing prefix test that pinned consecutive chat answers concatenating into one entry was updated to assert two entries; the boundary-consistency property it exists for still holds and is also asserted by the new dual-path test.
- `pnpm --filter @vibesys/core-state check` (tsc), biome over the changed files, and `pnpm --filter @vibesys/core-state test` (66 tests, 0 failures) all pass.
- Manual, real provider: two questions asked in experiment chat each produced exactly one Answer block, and the second answer stayed its own block. No duplicate appeared after later event-driven reconciliations.

### Additional findings while solving this issue

- Pre-existing bug found and fixed alongside the duplication: consecutive chat answers glued into a single transcript entry, because the assistant-merge rule treated two missing turn ids as equal. This affected any thread with more than one answer, independent of the duplication bug. The tightened rule is safe beyond chat: streamed entries always carry a turn id, so only complete answers are affected.

Closes #473.

### DIFF
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -162,6 +162,26 @@ describe('prefix merges across the chunk boundary', () => {
     ]);
   });
 
+  it('folds a tail answer over a streamed chat turn left in the chunk', () => {
+    const events = [
+      threadCreatedEvent(1, 'thread-a'),
+      chatChunkEvent(2, 'thread-a', 'partial '),
+      chatChunkEvent(3, 'thread-a', 'stream'),
+      chatEvent(4, 'thread-a', 'complete answer'),
+    ];
+
+    // The turn's streamed chunks fall below the floor; only its terminal answer
+    // lands in the tail. A full replay folds the answer over its streamed turn
+    // and renders it once, so the backfilled merge must too rather than leaving
+    // the streamed entry beside a second answer entry.
+    const merged = foldAsPrefix(events, 3);
+
+    expect(merged).toEqual(reduceEventBatch(initialCoreState(), events));
+    expect(merged.chatTranscripts['thread-a']?.map(entry => entry.content)).toEqual([
+      'complete answer',
+    ]);
+  });
+
   it('keeps a round started in the chunk and finished in the tail', () => {
     const events = [chunkEvent(1, 'work'), roundFinishedEvent(2)];
 
@@ -759,6 +779,17 @@ function chatEvent(sequence: number, threadId: string, answer: string, title?: s
     chat_thread_id: threadId,
     status: 'answered',
     data: {kind: 'chat', answer, ...(title === undefined ? {} : {thread_title: title})},
+  };
+}
+
+function chatChunkEvent(sequence: number, threadId: string, content: string): RunEvent {
+  return {
+    ...baseEvent(sequence, 'agent_output_chunk'),
+    agent_kind: 'chat',
+    round_label: 'experiment-chat',
+    chat_thread_id: threadId,
+    invocation_id: `${threadId}-turn`,
+    data: {kind: 'agent_output_chunk', channel: 'assistant', content},
   };
 }
 
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -153,11 +153,12 @@ describe('prefix merges across the chunk boundary', () => {
         model: 'opus',
       },
     ]);
-    // Consecutive chat answers carry no turn id, so the fold concatenates them
-    // exactly as it does inside one batch. The point is that it still does when
-    // the two answers land on opposite sides of the boundary.
+    // Each chat answer is a complete turn and stays a separate entry. The point
+    // is that the prefix fold agrees with the in-batch fold even when the two
+    // answers land on opposite sides of the boundary.
     expect(merged.chatTranscripts['thread-a']?.map(entry => entry.content)).toEqual([
-      'first answersecond answer',
+      'first answer',
+      'second answer',
     ]);
   });
 
--- a/clients/core-state/src/core-state.test.ts
+++ b/clients/core-state/src/core-state.test.ts
@@ -400,6 +400,49 @@ describe('core state projection', () => {
     expect(state.transcript).toEqual([]);
   });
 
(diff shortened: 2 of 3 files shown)


vic-lsh:
Review finding:

[P2] Reconcile terminal chat answers when history is merged across the backfill boundary (`clients/core-state/src/core-state.ts:735-741`, prefix merge at 399-405).

If a tail bootstrap contains the terminal `chat` event while that turn's streamed chunks are below the history floor, the tail stores the Answer separately. Later `reduceEventPrefix` folds the older chunks, but `mergeChatTranscriptsPrefix` combines already-folded entries through `foldTranscriptEntry` and never invokes `foldChatAnswer`. The streamed entry and terminal Answer have different IDs, so the duplicate returns. The new prefix test covers consecutive terminal answers, not a streamed-turn/final-answer boundary.

Please reconcile final answers during prefix merge, or preserve enough turn identity to deduplicate there, and add a boundary test with streamed chunks before the floor and the terminal chat event after it.

PR #528, opened 2026-09-01
Support optional task Dockerfiles

vic-lsh:
## Problem

Repository-native tasks cannot currently declare the execution environment they require. Authors must document a separate `docker build` step and pass `--docker-image`, which makes task bundles less self-contained and gives each task an ad hoc launch procedure.

## Solution

Recognize an optional `.vibesys/tasks/<task>/Dockerfile` by convention. Its presence automatically selects Docker for fresh runs. VibeSys builds it once per launch with the task directory as the complete build context and launches the run using the immutable image ID returned by Docker.

This adds no manifest fields or image naming/cache policy. Docker layer caching remains the cache. Explicit `--docker` and `--run-environment docker` are accepted, while `--docker-image` and non-Docker environments fail as conflicting inputs. Legacy root bundles do not auto-detect a root Dockerfile because it commonly belongs to the candidate application.

### Architecture

`vs-project` discovers and validates the conventional file. `InputBundle` carries its resolved path. The headless launch boundary chooses the environment and invokes a focused image builder. The builder uses direct argv, a task-only context, a bounded timeout, and Docker's `--iidfile`; the existing run-environment record persists the resulting image ID.

Resume behavior remains explicit: a run recorded as local remains local even if a Dockerfile is later added. A run recorded as Docker rebuilds the task image and the existing configuration comparison detects image drift.

This does not add evaluator isolation within Docker. The existing Docker run environment still uses the same writable container for agent and gate commands.

## Verification

The change was checked with focused layout, input loading, CLI/environment selection, launch wiring, and image-builder tests. Documentation link checking, Ruff, formatting, and whitespace checks pass.

### Correctness properties

- Only repository-native task Dockerfiles opt in.
- The Docker build context is exactly the selected task directory, so candidate source and Git metadata are unavailable to `COPY`.
- Symlinked and non-regular Dockerfiles are rejected.
- Successful builds must return a full immutable `sha256:` image ID.
- Validation does not build images, and each actual outer-loop launch builds at most once.
- Explicit conflicting image or environment selections fail instead of being silently overridden.
- Resume preserves the recorded environment and detects rebuilt-image drift through the existing run configuration contract.

### Testing

- `uv run pytest -q tests/entrypoints/test_headless.py` (109 passed)
- `uv run pytest -q libs/vs-project/tests/test_layout.py tests/vibesys/test_input_sources.py tests/vibesys/sandbox/test_task_image.py` (75 passed)
- `uv run ruff check libs/vs-project/src/vs_project/_layout.py libs/vs-project/tests/test_layout.py src/entrypoints/headless.py src/vibesys/input_manifest.py src/vibesys/sandbox/task_image.py tests/entrypoints/test_headless.py tests/vibesys/test_input_sources.py tests/vibesys/sandbox/test_task_image.py`
- `uv run ruff format --check libs/vs-project/src/vs_project/_layout.py libs/vs-project/tests/test_layout.py src/entrypoints/headless.py src/vibesys/input_manifest.py src/vibesys/sandbox/task_image.py tests/entrypoints/test_headless.py tests/vibesys/test_input_sources.py tests/vibesys/sandbox/test_task_image.py`
- `uv run python scripts/check_doc_links.py` (321 Markdown files checked)
- `git diff --check`

### DIFF
--- a/docs/cli-flags.md
+++ b/docs/cli-flags.md
@@ -330,6 +330,14 @@ starts with an actionable error.
 | `--modal` | Modal workflow. | Mutually exclusive with `--docker`. Intended for remote GPU dispatch. |
 | `--run-environment skypilot` | Local CPU editor with SkyPilot evaluators. | Requires portable task resources and an operator-owned cluster profile. See [Remote Slurm execution](remote-slurm-execution.md). |
 
+A repository-native task may provide
+`.vibesys/tasks/<task>/Dockerfile`. Its presence automatically selects the
+Docker environment, builds with the task directory as its complete context,
+and uses the resulting immutable image ID. No manifest field or pre-build step
+is required. Explicit alternative environments and `--docker-image` conflict
+with this convention; explicit `--docker` is harmless. Root Dockerfiles in
+legacy bundles remain candidate files and are not auto-detected.
+
 `--docker-image` overrides the backend's default container image when Docker or
 Modal is active, and the local CPU editor image for SkyPilot.
 
--- a/docs/contributing/coding-best-practices.md
+++ b/docs/contributing/coding-best-practices.md
@@ -16,8 +16,10 @@ path.
   code or prompt skeletons.
 - Keep reusable targets in their candidate repository under
   `.vibesys/tasks/<name>/`: `OBJECTIVE.md`, `vibesys.input.toml`, optional
-  `reference/`, task-owned checker/benchmark programs, and an optional
-  `README.md`.
+  `reference/`, task-owned checker/benchmark programs, an optional `Dockerfile`,
+  and an optional `README.md`. A task Dockerfile installs the execution
+  environment only. Its build context is the task directory, while VibeSys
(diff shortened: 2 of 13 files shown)


vic-lsh:
Additional end-to-end verification after rebasing #527: a clean-copy `uv run vibesys --outer-loop agent ... --task verus-mpmc-open --stub-agent` run, with no Docker flags, automatically built and selected the task image and completed one round. Repeated cached builds returned the same immutable image ID after disabling generated provenance attestations (`sha256:369be36b...a223` both times).

vic-lsh:
The first real Luna campaign exposed and validated two task-container activation fixes now included here: `054f2d4f` retains a unique local reference and records the stable runnable manifest ID, and `702b0d2b` installs pip based on the `pip3` capability and supports PEP 668 images before MCP bootstrap. Real validation produced the same runnable manifest ID twice and launched it successfully; the active campaign subsequently reached its first Codex model calls.

PR #529, opened 2026-09-01
TUI: One canonical slash-command registry shared by every input surface

### LINKED ISSUE #519: TUI: define slash commands in one canonical registry shared by every input surface
### Workstream

CLI/DX

### Problem

Slash commands live in `clients/tui/src/commands.ts` as two separate const tables (`SLASH_COMMANDS` at commands.ts:37-52, `CHAT_SLASH_COMMANDS` at commands.ts:101-105) with two hand-written parsers (`parseCommand` at commands.ts:154-197, `parseChatCommand` at commands.ts:142-149) that re-spell every command name as string literals in match arms, unchecked by the type system (no enum or union ties a table entry to its parser arm). The consequences are the concrete divergences confirmed live and by tests: chat autocomplete cannot suggest global commands (`suggestChatSlashCommands('/pa')` returns `[]` per commands.test.ts:181 even though `/pause` works in chat), `/clear` and `/model` fail in the command bar, `/resume` silently means two different things depending on focus and blocks run-resume from chat entirely, `/Pause` produces `Unknown command` in one surface and a help card in the other because one match is case-insensitive and the other is not, `/help` typed in chat shows global help while chat-specific help is only reachable via an unrecognized command, and the context filter that hides `/chat` when the chat is docked applies to suggestions but not parsing. Additional copies drift independently: the hand-maintained `Planned` block in `helpText` lists `/round` and `/invocation` which do not exist, footer hint strings in `ui/app.ts:37-48` embed `/todos`, `/prompt`, `/open-round --N` as literals, and `clients/tui/README.md:24-35` omits five real commands. A fix applied to one chatbox or one parser does not translate to the other, which is exactly how the `/pause`//`/steer` UX regressions in issue 2 survived.

### Desired outcome

One canonical command registry module defines every slash command exactly once: canonical name string, aliases, argument spec (none, optional, required with usage string), description, help section, which surfaces it is available on (command bar, chat, both), context availability, and the parsed action it produces. Both the command bar and the chat composers resolve input through the same matcher (case handling, unknown-command diagnosis, and usage errors identical everywhere), autocomplete and `/help` for every surface are generated from the registry filtered by surface, and name collisions such as run-`/resume` versus chat-thread-`/resume` are resolved explicitly in the registry (rename, alias, or surface-scoped variants) rather than by parser ordering. TypeScript enforces that a command cannot be registered without a handler and that parser match strings cannot drift from registry names.

### Acceptance criteria

- [ ] A single registry module is the only place a command name string literal appears; parsers and suggestion lists are derived from it (enforced by type, not convention).
- [ ] The same input typed in the command bar, docked chat, and modal chat yields the same parse result or an explicitly registry-declared surface difference.
- [ ] Chat autocomplete suggests every command available in chat, including forwarded global commands.
- [ ] Unknown-command and usage errors are identical across surfaces, including case-mismatch handling.
- [ ] The `/resume` collision is resolved explicitly and a paused run can be resumed from every surface.
- [ ] `/help` output per surface and the footer hints are generated or asserted against the registry; the stale `Planned` entries are removed or implemented.
- [ ] `commands.test.ts` divergence tests are replaced by tests asserting cross-surface parity.

### Scope and non-goals

In scope: 
`clients/tui/src/commands.ts`, the submit paths in `session-controller.ts` (submitCommand/submitChat), suggestion menus, help text generation, and footer hint sources. 

Not in scope: 
adding new commands, backend protocol changes (`command.pause`/`command.resume`/`command.steer` stay as-is), and the ack-rendering fixes tracked in issue 2.

Chat composers must still forward non-command text to the chat agent; the registry only owns inputs beginning with `/`. The modal chat z-order problem (acks behind the overlay) is an independent rendering fix but the registry should expose enough metadata (ack style per command) to route feedback to a surface-appropriate location.

### Verification approach

Type-level exhaustiveness plus unit tests that iterate the registry and assert parse/suggest/help parity across all three surfaces, and a live tmux pass exercising every command from every surface.

#284

AyanBinRafaih:
## Problem

Fixes #519.

Slash commands lived in `clients/tui/src/commands.ts` as two separate const tables (`SLASH_COMMANDS`, `CHAT_SLASH_COMMANDS`) with two hand-written parsers (`parseCommand`, `parseChatCommand`) that re-spelled every command name as string literals in match arms, unchecked by the type system. The command bar and the two chat composers diverged as a result:

- chat autocomplete could not suggest global commands (`suggestChatSlashCommands('/pa')` returned `[]` even though `/pause` worked in chat);
- `/Pause` produced `Unknown command` on one surface and a help card on the other, because one match was case-insensitive and the other was not;
- `/resume` meant two different things depending on focus, and blocked run-resume from the chat entirely;
- `/help` typed in chat showed the global help, while the chat-specific help was reachable only through an unrecognized command.

The copies drifted independently: a hand-maintained `Planned` block in `helpText` listed `/round` and `/invocation`, which do not exist; footer hint strings in `ui/app.ts` embedded `/todos`, `/prompt`, and `/open-round --N` as literals; and `clients/tui/README.md` omitted five real commands. A fix applied to one parser did not translate to the other.

## Solution

One registry, `COMMAND_REGISTRY` in `commands.ts`, defines every slash command exactly once: canonical name, aliases, argument arity, description, help section, the surfaces it is available on, its context availability, and a required `parse` handler that returns the command's action. Both the command bar and the chat composer resolve input through the same `parseCommand(text, {surface})` matcher, so case handling, unknown-command diagnosis, and usage errors are identical everywhere. Suggestions and per-surface help are generated from the registry filtered by surface, and footer hints reference it through an exported `COMMAND_NAMES` map instead of literals.

The matcher returns a discriminated `ParsedCommand` union. Both input surfaces resolve their own text once and hand the resulting action to one `#dispatchCommand` in `session-controller.ts`, which switches on `action.kind` with a `never` default: a command cannot be registered without a handler, and an unhandled action kind is a compile error.

The registry enforces each command's declared arity before dispatch, so a no-argument command rejects a trailing phrase (`/pause typo`) with its registered usage string rather than executing, and a required-argument command rejects an empty one. This restores what the pre-refactor exact-match parsers did.

The `/resume` collision is resolved explicitly: `/resume` resumes the paused run on every surface, and the chat's own thread switch is renamed `/switch` (chat only), so no command name appears twice and a paused run is now resumable from the chat as well as the command bar.

`/design` is an ordinary pane command in the registry, alongside `/perf`: it parses to `{kind: 'request', request: {type: 'query.design'}, paneView: 'design'}` and reaches the right pane through the same `PaneView` dispatch, with no special case.

### Architecture

`commands.ts` owns the registry and the matcher, and is the only place a command name string literal appears. The command bar (`ui/command-input.ts`) and the chat composer (`ui/chat-composer.ts`) each call `suggestSlashCommands` with their surface for autocomplete. `session-controller.ts` parses with the typing surface and dispatches the resulting action; the chat handles its three thread commands and its own help, and forwards every other resolved action to the shared dispatcher, so the chat and the command bar cannot diverge on a shared command or on its error text. The registry also exports `COMMAND_SPECS`, the contract half of each entry, which the contract tests and the README check enumerate. The backend protocol is unchanged: `command.pause`, `command.resume`, `command.steer`, `query.performance`, and `query.design` are the same requests as before.

```mermaid
flowchart TD
    CB[command bar input] --> P
    CH[chat composer input] --> P
    P["parseCommand(text, surface)"] --> R{"registry lookup:
name/alias, case-insensitive,
available on this surface?"}
    R -->|no| U["unknown: bar -> error,
chat -> chat help"]
    R -->|yes| C{"declared arity satisfied?"}
    C -->|no| E["error: registered usage string"]
    C -->|yes| A["entry.parse -> ParsedCommand action"]
    A --> S["#dispatchCommand: switch on action.kind
exhaustive, never default"]
    E --> S
    S --> Shared["shared: request / openChat / toggle /
theme / openRound / help"]
    S --> ChatOnly["chat only: chatClear / chatModel / chatSwitch"]
    REG[("COMMAND_REGISTRY:
one definition per command")] --> R
    REG --> Sug["suggestSlashCommands / helpText,
filtered by surface"]
    REG --> Names["COMMAND_NAMES -> app.ts footer hints"]
    REG --> Specs["COMMAND_SPECS -> arity and README contract tests"]
```

## Verification

Automated. The TUI test suite covers parsing parity across both surfaces exhaustively, so no manual verification of parsing is required. The one behavior that is stated in terminal geometry (chat autocomplete listing the forwarded globals) is covered through the OpenTUI test renderer in `ui/app.test.ts`.

### Correctness properties

- A command name string literal appears only in the registry. Parsers, suggestions, help, footer hints, and the README check are all derived from it, so the surfaces cannot drift.
- The same input typed on the command bar and in the chat yields the same parsed action, or an explicitly registry-declared surface difference (chat-only commands, and `/chat`, which the chat does not offer).
- Command names match case-insensitively on every surface, so `/Pause` and `/pause` resolve identically.
- Every command's declared arity is enforced before its parser runs, on every surface it is available on: a `none` command rejects any trailing text and a `required` command rejects an empty argument, both with the command's registered usage string.
- Each input is parsed exactly once, on the surface it was typed on, so a chat-only command's usage error is not re-diagnosed as an unknown command by the command bar.
- Chat autocomplete suggests every command available in the chat, including the forwarded global commands.
- A paused run resumes from `/resume` on every surface; the chat's thread switch is `/switch`, so the two never collide by name.
- A command cannot be registered without a `parse` handler, and an unhandled action kind is a compile error.
- `clients/tui/README.md`'s two command tables list exactly the registry's command-bar and chat-only commands, in registry order.

### Testing

- `commands.test.ts` is rewritten from divergence tests into cross-surface parity tests: shared commands parse identically on both surfaces, case handling is identical, `/steer` usage errors match, chat autocomplete includes forwarded globals (`suggestSlashCommands('/pa', {surface: 'chat'})` returns `/pause`), the chat-only `/clear`, `/model`, and `/switch` appear only in the chat, `/resume` resolves to `command.resume` on both surfaces, unknown input is an error on the command bar and chat help in the chat, and the generated help drops the stale `/round`, `/invocation`, and `Planned` entries.
- `commands.test.ts` argument-contract block: the reported cases (`/pause typo`, `/perf extra`, `/help ignored`, `/clear definitely-not`), plus a test that enumerates `COMMAND_SPECS` and asserts every command's declared arity on every surface it is available on, and a guard that all three arities occur so that loop cannot go vacuous. Removing the `checkArgument` call fails three of these tests.
- `commands.test.ts` README block: both `clients/tui/README.md` command tables are parsed and asserted against the registry, in registry order. Deleting a row fails the test.
- `session-controller.test.ts`: the help test asserts the `Planned` block is gone; the thread-list test drives `/switch`; a paused run resumes from the chat with the same `command.resume` request as the command bar; `/pause typo` reports `Usage: /pause` and sends no request; and `/clear definitely-not` typed in the chat reports `Usage: /clear` rather than an unknown-command error, and starts no thread.
- `ui/app.test.ts`: the fake controller's chat submit path and the chat-suggestion and thread-menu tests are updated to the registry API and the `/switch` rename, and assert the forwarded globals now appear in chat autocomplete.
- `clients/tui/README.md` lists every command the registry defines, including the previously undocumented `/todos`, `/prompt`, and the chat-only `/clear`, `/model`, and `/switch`, and drops the "planned controls" wording.

Commands run at the head of this branch:

- `cd clients/tui && bun test --max-concurrency 1 src`: 519 pass. (One local run also saw `launcher > returns the backend argument error for an explicitly empty runs directory` fail; that case spawns a real backend and was failing on a stale working directory in the sandbox, not on this change. It passes on a clean re-run and in CI.)
- `pnpm check:clients`: pass.
- `pnpm check:ts`: pass.
- `pnpm check:ts-architecture`: pass.
- No Python changed, so `pytest` was not run.

Closes #519.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -21,19 +21,33 @@ the Python `vibesys` package in the Python environment you want to use, or set
 Use Experiment chat for ordinary questions about the current run. The command
 input accepts these slash commands:
 
+Every command below is defined once in a shared registry, so the command bar and
+the chat resolve the same name to the same action, match case-insensitively, and
+report the same errors.
+
 | Command | Behavior |
 | --- | --- |
-| `/help` | Show commands and planned controls. |
-| `/chat` | Put the pane keys on the docked chat, or open it as a modal where it cannot dock; `/chat <question>` asks immediately. |
-| | Slash commands work inside the chat too, and do the same thing as in the main input. |
+| `/help` | Show the commands available on the current surface. |
+| `/chat` | Put the pane keys on the docked chat, or open it as a modal where it cannot dock; `/chat <question>` asks immediately. Command bar only, since the chat has nothing to open. |
+| | Every command below works in the chat too, and does the same thing as in the command bar. |
 | `/pause` | Pause after the current agent call finishes. |
-| `/resume` | Resume a paused run. |
+| `/resume` | Resume a paused run. Works from the command bar and the chat. |
 | `/steer <message>` | Queue an instruction that is appended to the next agent invocation's prompt. |
 | `/open-round` | Open the rounds behind the selected hypothesis. |
 | `/open-round --N` | Open round N, inside whichever hypothesis owns it. |
 | `/perf` | Plot the recorded performance metric by round, in the right pane. |
 | `/design` | Summarize what each round changed in the workspace, in the right pane. |
+| `/todos` | Expand or collapse the visible agent's todo list. |
+| `/prompt` | Expand or collapse the latest prompt in view. |
 | `/theme` | Pick a theme from a keyboard-navigable list; `/theme <name>` switches immediately. |
+
+The chat composer adds its own thread commands, which exist only in the chat:
+
+| Command | Behavior |
+| --- | --- |
+| `/clear` | Start a fresh thread with the current thread's agent and model. |
+| `/model` | Pick a harness and model, and start a thread on it. |
+| `/switch` | Switch to another chat thread. |
 
 ### Experiment log
 
@@ -188,8 +202,7 @@ not support OSC52, VibeSys keeps the selection and shows a status explaining
 that the terminal's native copy command is the fallback. Rounds and agents can
 also be clicked: a round chip
 selects its round, an agent node filters the transcript to that agent, and
-clicking the selected node clears the filter. Commands listed under "Planned" in
-`/help` are not accepted yet.
+clicking the selected node clears the filter.
 
 The rounds strip covers the whole run, including rounds it has not reached yet,
 and windows onto the part that fits. The selected round is always in view, and
--- a/clients/tui/src/commands.test.ts
+++ b/clients/tui/src/commands.test.ts
@@ -1,196 +1,147 @@
 import {describe, expect, it} from 'bun:test';
 import {
   availableCommands,
+  COMMAND_NAMES,
   chatHelpText,
   helpText,
-  parseChatCommand,
+  type ParsedCommand,
   parseCommand,
-  SLASH_COMMANDS,
   slashCommandRange,
-  suggestChatSlashCommands,
   suggestSlashCommands,
 } from './commands.js';
 
+const onCommand = (text: string): ParsedCommand => parseCommand(text, {surface: 'command'});
+const onChat = (text: string): ParsedCommand => parseCommand(text, {surface: 'chat'});
+const names = (commands: readonly {name: string}[]): string[] =>
+  commands.map(command => command.name);
+
+/** The commands both surfaces resolve identically, so parity is asserted once. */
+const SHARED = [
+  '/help',
+  '/pause',
+  '/resume',
+  '/steer look at the cache',
+  '/open-round',
+  '/perf',
+  '/design',
+  '/todos',
+  '/prompt',
+  '/theme',
+];
+
 describe('parseCommand', () => {
-  it('accepts the intentionally small slash-command surface', () => {
-    expect(parseCommand('/open-round')).toEqual({openRound: {}});
-    expect(parseCommand('/open-round --3')).toEqual({openRound: {round: 3}});
-    expect(parseCommand('/open-round 3')).toEqual({openRound: {round: 3}});
-    expect(parseCommand('/open-round latest').error).toContain('Unknown round: latest');
-    // Visualization commands opt into the right pane through one field.
-    expect(parseCommand('/perf')).toMatchObject({
-      request: {type: 'query.performance'},
-      paneView: 'perf',
+  it('parses the command surface into discriminated actions', () => {
+    expect(onCommand('/help')).toEqual({kind: 'help'});
+    expect(onCommand('/pause')).toEqual({kind: 'request', request: {type: 'command.pause'}});
+    expect(onCommand('/resume')).toEqual({kind: 'request', request: {type: 'command.resume'}});
+    expect(onCommand('/steer prioritize the KV cache path')).toEqual({
+      kind: 'request',
+      request: {type: 'command.steer', text: 'prioritize the KV cache path'},
     });
-    // Modal surfaces stay modal: no pane routing on any of them.
-    expect(parseCommand('/help').paneView).toBeUndefined();
-    expect(parseCommand('/theme').paneView).toBeUndefined();
-    expect(parseCommand('/chat').paneView).toBeUndefined();
-    expect(parseCommand('/nope').paneView).toBeUndefined();
-    expect(parseCommand('/perf')).toMatchObject({
+    expect(onCommand('/open-round')).toEqual({kind: 'openRound'});
+    expect(onCommand('/open-round --3')).toEqual({kind: 'openRound', round: 3});
+    expect(onCommand('/open-round 3')).toEqual({kind: 'openRound', round: 3});
+    expect(onCommand('/perf')).toMatchObject({
+      kind: 'request',
       request: {type: 'query.performance'},
       responseView: 'perf',
-    });
-    expect(parseCommand('/chat what changed in the latest round?')).toEqual({
-      localView: 'chat',
-      chatMessage: 'what changed in the latest round?',
+      paneView: 'perf',
     });
     // Every visualization goes through the one pane mechanism, so /design
     // reaches the right pane exactly the way /perf does.
-    expect(parseCommand('/design')).toEqual({
+    expect(onCommand('/design')).toEqual({
+      kind: 'request',
       request: {type: 'query.design'},
       paneView: 'design',
     });
-    expect(SLASH_COMMANDS.map(command => command.name)).toContain('/design');
-  });
-
-  it('parses run-control commands', () => {
-    expect(parseCommand('/pause')).toEqual({request: {type: 'command.pause'}});
-    expect(parseCommand('/resume')).toEqual({request: {type: 'command.resume'}});
-    expect(parseCommand('/steer prioritize the KV cache path')).toEqual({
-      request: {type: 'command.steer', text: 'prioritize the KV cache path'},
-    });
-  });
-
-  it('requires a message for /steer', () => {
-    expect(parseCommand('/steer').error).toContain('Usage: /steer');
-    expect(parseCommand('/steer   ').error).toContain('Usage: /steer');
-  });
-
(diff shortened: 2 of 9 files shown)


AyanBinRafaih:
This PR is stacked on #497, hence the merge conflicts. If we merge #497 first, then the issues should get resolved.

vic-lsh:
Review finding:

[P2] Enforce the registry's argument contract before dispatch (`clients/tui/src/commands.ts:304-317`).

`parseCommand` records the trailing argument and calls `command.parse(argument)` unconditionally, but commands declared with `args: 'none'` have parsers that ignore it. As a result, `/pause typo`, `/perf extra`, and `/help ignored` still execute; in chat, `/clear definitely-not` clears or starts a thread. Before this refactor, the exact-match parsers rejected these strings, so this is a behavior regression and can turn a slash-prefixed phrase into a state-changing command.

Please validate `none` and `required` centrally before dispatch, using the registered usage string for errors, or make each parser reject unexpected arguments. Add command-bar and chat cases for trailing text on no-argument commands.

vic-lsh:

vic-lsh:
Thanks @AyanBinRafaih — the registry is the right shape, and the exhaustive `ParsedCommand` switch is what makes the two surfaces provably identical rather than identical-for-now. I rebased the branch onto the merged #497 and pushed a maintainer pass on top of your two commits rather than asking for another round. Here is what changed and why.

**Rebase onto the merged #497.** The branch carried #497's commit, and #497 merged in a reworked shape, so I cherry-picked only this PR's own two commits onto `main`. Three files conflicted. The resolutions:

- `/design` is now an ordinary pane command in the registry: `parse: () => ({kind: 'request', request: {type: 'query.design'}, paneView: 'design'})`. The `openDesign` action kind and `#openDesignSummary` are gone, so `/design` reaches the right pane through the same `#requestPane` / `PaneView` dispatch as `/perf`, and adding a third visualization stays a compile error until `#renderPane` handles it.
- `clients/tui/README.md` keeps #497's `/design` row (right pane, not overlay) and your `/todos` and `/prompt` rows.
- `session-controller.ts` keeps `main`'s `PersistentEventStream` boot and design-log single-flight; only `submitCommand` / `submitChat` came from this branch.

**[P2] The argument contract is enforced, but the test named the commands by hand.** Your `checkArgument` is correct and runs before every dispatch, so the report's cases (`/pause typo`, `/perf extra`, `/help ignored`, `/clear definitely-not`) are all rejected. The test listed eight of the ten `args: 'none'` commands, though: `/todos` and `/prompt` were not covered, and the next command registered would not be either. It now enumerates `COMMAND_SPECS` and asserts every command's declared arity on every surface it is available on, plus a guard that all three arities are actually present so the loop cannot go vacuous. To make that possible the registry exports `COMMAND_SPECS`, the contract half of each entry with `parse` and `hiddenWhen` stripped. Verified the test fails at the pre-fix code: dropping the `checkArgument` call fails three cases, including the enumerated one.

**[P2] The chat threw away its own usage error.** `submitChat` forwarded the raw *text* to `submitCommand`, which re-parsed it with `surface: 'command'`. That was harmless while every forwarded action was a success, but the argument check made it visible: `/clear definitely-not` resolves on the chat surface to `Usage: /clear`, then the forward re-parses it on the command bar, which does not offer `/clear` at all, and the user gets `Unknown command: /clear definitely-not. Use /help.` for a command that exists. `submitChat` now hands the resolved action to a shared `#dispatchCommand`, so each input is parsed exactly once, on its own surface, and both surfaces execute the same action through the same code. Regression test in `session-controller.test.ts`.

**[P3] The README table was a hand-maintained copy.** It is the drift this PR exists to remove, and it was still the one copy nothing checked. `commands.test.ts` now parses both tables out of `clients/tui/README.md` and asserts they equal the registry's command-bar and chat-only commands, in registry order. Removing a row fails the test.

**Body.** Rewritten against the template with the "stacked on #497" note dropped and the `/design` claims corrected to the pane shape.

Checks on the pushed head: `bun test --max-concurrency 1 src` in `clients/tui` (518 pass, 1 fail: the pre-existing `launcher` case that spawns a real backend and fails identically on unmodified `main` in this environment), `pnpm check:clients`, `pnpm check:ts`, `pnpm check:ts-architecture`. No Python changed.

Not merged and not marked ready.

PR #532, opened 2026-09-01
Give each framework benchmark run its own result file and share the gate

### LINKED ISSUE #477: TUI: Evaluator-declared metric direction is resolved and then discarded, so protocol-v2 tasks can never resolve a hypothesis

### LINKED ISSUE #480: TUI: Framework benchmark result file is a shared `/tmp` path with no per-invocation identity

AyanBinRafaih:
> Stacked PR 1/6. Base branch: `main`. This is the bottom of the #495 split; the remaining five PRs (#533, #534, #535, #536, #537) stack on top in that order.

## Problem

Two independent bugs in the framework benchmark gate, both rooted in the same place.

**The result file had no per-invocation identity (#480).** The gate wrote to the fixed path `/tmp/vibesys-framework-benchmark-{round}-{retry}.json` and then read it back with `cat`. Nothing removed the file first, so any file already at that path satisfied the read: a leftover from a crashed earlier round, a concurrent run on the same host, or another user's file on a shared `/tmp`. A benchmark that wrote no result at all could therefore be reported as a successful measurement, with someone else's number attached to this candidate. On a shared `/tmp` a file owned by another user also makes the write fail outright with permission denied.

**The evaluator-declared metric direction was resolved and then discarded (#477).** `_read_protocol_benchmark` resolved `metric_direction` from the configured objective, falling back to the evaluator's `hello` declaration, and then the caller threw that away and recomputed the direction from `objectives.toml` alone (defaulting to `max` for the legacy scalar contract). Every shipped `result_protocol = 2` task configures bare metric names without directions, so the recomputed direction was always `None`, and every metric comparison that needed it came back undecided.

Benchmark execution and parsing also lived inline in the agent loop (`_read_protocol_benchmark`, `_run_framework_benchmark`, `_PROTOCOL_OUTPUT_FLAG`), so the evolve loop had no way to share the same trusted reader.

## Solution

Move benchmark execution and parsing out of the agent loop into `vibesys.loops.gates` so both optimization loops share one trusted implementation, make every benchmark invocation read only a file it created itself, and keep the direction the reader resolved.

- `run_benchmark_gate` writes to `/tmp/vibesys-framework-benchmark-{slug}-{12-hex-nonce}.json` and removes that path with `rm -f --` before executing. The result is recovered from stdout between explicit markers produced by the same invocation, and a missing or unparseable result is a hard gate failure, never a silent fallback to whatever is on disk. A `finally` block removes the transport artifact on every exit path so the gate does not leak one JSON per benchmark.
- `output_slug` is validated at the boundary: it is interpolated into the result path, so an empty slug or one containing `/` or `..` is rejected with an error naming the field.
- The metric direction the reader resolved (configured objective first, then the evaluator's `hello` declaration) now travels out on `FrameworkBenchmarkOutcome.metric_direction` and is what the loop records. The legacy scalar contract keeps its historical `max`; a protocol metric with no direction anywhere stays `None`, which `MetricSpace` reports as `INCOMPARABLE` rather than guessing an order.
- The relocated machinery is otherwise verbatim: `FrameworkBenchmarkOutcome`, `read_protocol_benchmark`, `_select_headline_metric`, and the marker transport move into `gates` unchanged, and the output flag and markers become the public `PROTOCOL_OUTPUT_FLAG` / `FRAMEWORK_BENCHMARK_*` constants.
- `BenchmarkGateResult` wraps the outcome with the command text and whether the gate reached a recordable conclusion. `passed` is a property derived from `outcome.feedback is None` rather than a second stored copy of the same fact.
- The agent loop's `_run_framework_benchmark` becomes a thin wrapper over `run_benchmark_gate`, and the dead private `_publish_subprocess_output` copy left behind in the agent loop is removed.

### Architecture

```mermaid
flowchart LR
    L[agent loop<br/>_run_framework_benchmark thin wrapper] --> G[gates.run_benchmark_gate]
    E[evolve loop] -. shares from #533 .-> G
    G --> P[nonce path<br/>rm -f -- before run<br/>rm -f -- in finally]
    G --> M[marker-delimited readback<br/>from same invocation]
    G --> O[FrameworkBenchmarkOutcome<br/>name, value, direction, row]
    O --> R[BenchmarkGateResult<br/>command, output, executed<br/>passed = feedback is None]
```

The gate layer (`vibesys.loops.gates`) now owns benchmark execution and parsing for every loop; callers receive a `BenchmarkGateResult` and never touch `/tmp` paths or output markers directly.

### Correctness properties

- Every benchmark gate invocation reads only a file it created itself. A benchmark that writes no result fails the gate instead of reporting a number it did not produce, and no two concurrent or sequential runs on one host can read or overwrite each other's result file.
- The transport artifact is removed on success, nonzero exit, and malformed output. Known limitation, documented in the code and covered by a test: `DockerSandbox.execute` and `ModalSandbox.execute` report a timeout as exit code `-1` rather than raising, so a timed-out benchmark may still write its result file after the cleanup ran. The nonce keeps that orphan from ever being read by a later invocation, which is the property the gate owns; the file itself can survive until the sandbox is torn down.
- `output_slug` is a single path segment; an empty slug or one containing `/` or `..` raises with the field named.
- A protocol metric's direction comes from the configured objective when there is one, else the evaluator's `hello` declaration. A metric with no direction from either source stays `None` so `MetricSpace` reports `INCOMPARABLE` instead of assuming an order.
- The benchmark outcome (metric name, value, full row) is otherwise identical to the pre-move agent-loop implementation; only the file path, the direction fallback, and the module boundary change.
- `BenchmarkGateResult` stores no field that another field determines.

### Testing

- `uv run pytest tests/vibesys/loops -q --no-cov`: all pass.
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 278 passed.
- `./scripts/check_format.sh`, `./scripts/check_lint.sh`, `./scripts/check_types.sh`: all clean.

New `tests/vibesys/loops/test_gates.py` drives the gate through a real `bash` judge backend so the transport file genuinely exists on disk:

- `test_benchmark_gate_fails_instead_of_reporting_a_stale_result` plants a file at the old fixed path and runs a benchmark that writes nothing. Verified to fail at the pre-fix path shape, reporting the planted `999.0` as the candidate's measurement.
- `test_a_timed_out_benchmarks_late_result_cannot_be_read_by_the_next_run` covers the documented timeout limitation. Also verified to fail at the pre-fix path shape.
- `test_benchmark_gate_removes_its_transport_artifact` covers success, malformed output, and a nonzero exit after the file was written.
- `test_benchmark_gate_keeps_the_evaluator_declared_direction` pins the #477 fix: a protocol-2 `hello` declaring `direction: "min"` with no configured objectives yields `min`. Verified to fail when the direction fallback is reverted to objectives-only.
- `test_benchmark_gate_rejects_an_output_slug_that_is_not_one_segment` and `test_configured_objective_overrides_the_declared_direction` cover the boundary and the precedence rule.

Closes #480. Closes #477.

### DIFF
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -9,11 +9,10 @@ from __future__ import annotations
 
 import hashlib
 import json
-import math
 import shlex
 import subprocess
-from collections.abc import Mapping, Sequence  # noqa: TC003  # tracked: #288
-from dataclasses import dataclass, replace
+from collections.abc import Sequence  # noqa: TC003  # tracked: #288
+from dataclasses import dataclass
 from pathlib import Path  # noqa: TC003  # tracked: #288
 from typing import Any, Literal
 
@@ -57,7 +56,10 @@ from vibesys.loops.gates import (
     GATE_FEEDBACK_TAIL_CHARS,
     GATE_LOG_TAIL_CHARS,
     GATE_RECORD_TAIL_CHARS,
+    PROTOCOL_OUTPUT_FLAG,
+    FrameworkBenchmarkOutcome,
     run_accuracy_gate,
+    run_benchmark_gate,
 )
 from vibesys.loops.metrics import (
     Measurement,
@@ -79,7 +81,6 @@ from vibesys.run.events import (
     ExperimentsChangedData,
     JudgeResultData,
     RoundFinishedData,
-    SubprocessOutputData,
 )
 from vibesys.sandbox.run_environment import (
     RunEnvironmentSpec,
@@ -106,13 +107,6 @@ from vibesys.skills import (
     ResolvedSkillSelection,
     build_skill_catalog,
     resolve_skill_selections,
-)
-from vs_evaluator_protocol import (
-    Hello,
-    ProtocolError,
-    check_objectives,
-    parse_records,
-    read_measurement,
 )
 from vs_loop_state.agent import RoundHistory, RoundRecord
 from vs_project import AgentRunConfiguration
@@ -1864,115 +1858,7 @@ def _run_framework_accuracy_gate(  # noqa: PLR0913  # tracked: #288
     return result.feedback
 
 
-_FRAMEWORK_BENCHMARK_MARKER = "__VIBESYS_FRAMEWORK_BENCHMARK_JSON__"
-_FRAMEWORK_BENCHMARK_END_MARKER = "__VIBESYS_FRAMEWORK_BENCHMARK_JSON_END__"
-
-# The flag every result-protocol evaluator registers for its output file; see
-# ``OutputFlag`` in the evaluator SDK (``sdk/vs-evaluator/vseval/schema.go``).
-_PROTOCOL_OUTPUT_FLAG = "--vs-output"
-
-
-@dataclass(frozen=True, slots=True)
-class FrameworkBenchmarkOutcome:
-    """What one framework benchmark run reported.
-
-    ``feedback`` is set exactly when the run failed and the round must retry.
-    On success ``metric_name`` and ``metric_value`` carry the headline scalar
-    both result contracts produce, and ``row`` carries the complete validated
-    metric row, which only the evaluator result protocol reports.
-    """
-
-    feedback: str | None = None
-    metric_name: str | None = None
-    metric_value: float | None = None
-    metric_direction: Literal["max", "min"] | None = None
-    row: Mapping[str, float] | None = None
-
-
-def _read_protocol_benchmark(
-    text: str, *, objectives: Sequence[Objective]
-) -> FrameworkBenchmarkOutcome:
-    """Turn a recovered evaluator record stream into a benchmark outcome.
-
-    Never raises for a bad stream: an invalid stream, a structured evaluator
-    failure, and an undecidable headline metric all become round feedback.
-    Objectives belong to the task rather than to the evaluator, so the check
-    that the evaluator declares every optimized metric happens here and not in
-    the protocol reader.
-    """
-    hello: Hello | None = None
-    try:
-        records = parse_records(text)
-        hello = next((record for record in records if isinstance(record, Hello)), None)
-        measurement = read_measurement(records)
-        if objectives and hello is not None:
-            check_objectives(hello, {objective.name for objective in objectives})
-    except ProtocolError as error:
-        return FrameworkBenchmarkOutcome(feedback=_protocol_feedback(error, hello))
-    if measurement.values is None:
-        return FrameworkBenchmarkOutcome(
-            feedback=f"benchmark evaluator reported a failure: {measurement.failure}"
-        )
-    outcome = _select_headline_metric(measurement.values, objectives)
-    if outcome.metric_name is not None:
-        configured = next(
-            (item.direction for item in objectives if item.name == outcome.metric_name),
-            None,
-        )
-        declared = (
-            hello.metrics[outcome.metric_name].direction
-            if hello is not None and outcome.metric_name in hello.metrics
-            else None
-        )
-        outcome = replace(outcome, metric_direction=configured or declared)
-    return outcome
-
-
-def _protocol_feedback(error: ProtocolError, hello: Hello | None) -> str:
-    """Render a rejected record stream, naming its reason code and metrics."""
-    declared = ", ".join(sorted(hello.metrics)) if hello is not None else "(none declared)"
-    return f"invalid benchmark result [{error.code}]: {error}; evaluator declares: {declared}"
-
-
-def _select_headline_metric(
-    values: Mapping[str, float], objectives: Sequence[Objective]
-) -> FrameworkBenchmarkOutcome:
-    """Select the back-compat headline scalar out of a complete metric row.
-
-    The first configured objective names it. With no objectives configured a
-    single-metric evaluator is unambiguous; anything else is a task
-    configuration error to report rather than a row to guess through.
-    """
-    if objectives:
-        name = objectives[0].name
-        return FrameworkBenchmarkOutcome(metric_name=name, metric_value=values[name], row=values)
-    if len(values) == 1:
-        name, value = next(iter(values.items()))
-        return FrameworkBenchmarkOutcome(metric_name=name, metric_value=value, row=values)
-    return FrameworkBenchmarkOutcome(
-        feedback=(
-            f"benchmark evaluator reported metrics {', '.join(sorted(values))} but the task "
-            "configures no objectives, so no headline metric is defined; declare the optimized "
-            "metrics in objectives.toml"
-        )
-    )
-
-
-def _metric_values(value: object, metric: str) -> list[object]:
-    if isinstance(value, dict):
-        matches = [item for key, item in value.items() if key == metric]
-        for item in value.values():
-            matches.extend(_metric_values(item, metric))
-        return matches
-    if isinstance(value, list):
-        matches: list[object] = []
-        for item in value:
-            matches.extend(_metric_values(item, metric))
-        return matches
-    return []
-
-
-def _run_framework_benchmark(  # noqa: C901, PLR0912, PLR0913, PLR0915  # tracked: #288
+def _run_framework_benchmark(  # noqa: PLR0913  # tracked: #288
     ctx: LoopContext,
     *,
     result_spec: BenchmarkResult | None,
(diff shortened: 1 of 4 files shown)


vic-lsh:
Review finding:

[P2] Clean up each nonce benchmark result file (`src/vibesys/loops/gates.py:299-310`).

The new path is collision-proof, but every invocation creates a unique `/tmp/vibesys-framework-benchmark-...-{nonce}.json`, pre-removes it, reads it, and never removes it afterward. The previous fixed-name leak was bounded; this becomes one persistent evaluator JSON file per benchmark invocation, including long-running agent/evolve campaigns on hosts where `/tmp` survives for weeks.

Please add deterministic cleanup covering success, nonzero exit, timeout, and malformed output, for example with a shell trap or a backend-level `finally`, and add a check that the transport artifact no longer exists after the gate completes.

vic-lsh:

vic-lsh:
I rebased and fixed this PR myself so the stack can start merging bottom-up. The change is @AyanBinRafaih's work; what follows is the rebase and the review findings I addressed on top of it.

## Rebase

Rebased onto `main` at `60c6763`, which is after #558 (gate tail constants), #576/#578 (`RunStatus`), #577 (per-attempt `JudgeOutcome`), #579 (`MetricSpace` owning `Objective` and comparison), and #583 (evolve on `MetricSpace`). The only conflicts were in `src/vibesys/loops/agent/loop.py`, entirely in the code this PR moves. #591 is still open, so its `[benchmark.result]` specs are not in this branch.

## Findings addressed

**[P1] The move silently changed `metric_direction`, and that change is the fix for #477.** Before the move, `_run_framework_benchmark` discarded the direction `_read_protocol_benchmark` had resolved and recomputed it from `objectives.toml` alone, falling back to `"max"` for the legacy scalar contract and `None` otherwise. The moved code keeps the evaluator-declared direction. That is the right behavior and it is exactly what #477 asks for, so I kept it and pinned it: `tests/vibesys/loops/test_gates.py::test_benchmark_gate_keeps_the_evaluator_declared_direction` runs a protocol-2 `hello` declaring `direction: "min"` with no configured objectives and asserts the outcome carries `min`. Verified it fails when the fallback is reverted to objectives-only. `Closes #477` moved from #533 to this PR, and the body now describes the direction fix.

**[P1] The regression evidence restated the fix's own constants instead of reproducing the symptom.** `test_framework_benchmark_extracts_declared_metric` and `test_protocol_benchmark_reads_complete_row` in `tests/vibesys/loops/agent/test_orchestrate.py` asserted the nonce regex and the `rm -f -- ` prefix on the executed command line: they pass by construction on any implementation that emits those strings, and say nothing about whether a stale file can still be read. Replaced with `tests/vibesys/loops/test_gates.py::test_benchmark_gate_fails_instead_of_reporting_a_stale_result`, which plants a file at the old fixed path, runs a benchmark that writes nothing through a real `bash` backend, and asserts the gate fails. Verified against the pre-fix path shape, where it reports the planted `999.0` as the candidate's measurement.

**[P2] `src/vibesys/loops/gates.py`: bare truncation lengths in the moved code.** `output[-1000:]`, `[-4000:]`, and `[-8000:]` are now `GATE_LOG_TAIL_CHARS`, `GATE_FEEDBACK_TAIL_CHARS`, and `GATE_RECORD_TAIL_CHARS` from #558, so the benchmark gate's windows match the accuracy gate's.

**[P2] `src/vibesys/loops/gates.py`: `Objective` imported from the evolve package.** After #579 the owner is `vibesys.loops.metrics`; importing it from `vibesys.loops.evolve.population` points the shared gate layer at a loop package.

**[P2] `BenchmarkGateResult` stored `passed` beside the outcome that determines it.** `passed` is exactly `outcome.feedback is None` at all three construction sites, so it was a second writer of one fact. It is now a property.

**[P2] `output_slug` was unvalidated.** It is interpolated into the result path that the gate creates and later removes, so an empty slug or one containing `/` or `..` is now rejected with an error naming the field. Covered by `test_benchmark_gate_rejects_an_output_slug_that_is_not_one_segment`.

**[P3] The cleanup does not cover the timeout path, and that limitation is now stated.** `DockerSandbox.execute` and `ModalSandbox.execute` report a timeout as exit code `-1` rather than raising, so the `finally` cleanup runs while the timed-out benchmark may still be alive in the sandbox and can write its result file afterwards. The file can outlive the gate; what it cannot do is be read by any later invocation, which is the property the nonce owns. Documented at the cleanup site and covered by `test_a_timed_out_benchmarks_late_result_cannot_be_read_by_the_next_run`, also verified to fail at the pre-fix path shape.

Gate tests now live in `tests/vibesys/loops/test_gates.py`, which is where the gate's own owner is; the moved-artifact cleanup test moved there from `test_orchestrate.py`.

## Provenance policy this stack implements

Recording it here because it is the decision the rest of the stack (#534, #535) is built on:

**Framework provenance is stamped only for a benchmark that the framework itself ran through a readable result contract.** A task that declares no `[benchmark.result]` spec and no `result_protocol` has no trusted number, so its rounds resolve `UNMEASURED` by design rather than being credited with the implementer's self-report. That is intended, not a gap to work around: the six latency examples that currently declare neither will keep resolving `UNMEASURED` until their `result_protocol` migration lands (#591 covers the first seven tasks).

## Checks

- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 278 passed

The title lost its `TUI:` prefix, which did not describe the change, and the body now matches the final diff.

PR #533, opened 2026-09-01
Run the framework benchmark gate in the evolve loop

AyanBinRafaih:
> Stacked PR 2/6. Base branch: `main` (#532 has merged). #534, #535, #536, and #537 stack on top in that order.

## Problem

The evolve loop never ran the trusted benchmark result contract. `_run_evolve` did not pass `benchmark_result`, `benchmark_result_protocol`, or the benchmark timeout into `run_evolve_loop`, so a task that declares a `[benchmark.result]` spec or `result_protocol = 2` got that contract honored under the agent loop and ignored under evolve. Candidate fitness came from the profiler agent's self-reported `perf_metric` even when the task declared a framework-owned benchmark that measures the same thing, which is the number the search then selects on for every generation.

Evolve's accuracy gate also passed the manifest timeout straight through, unlike the agent loop, which adds the run environment's `framework_setup_timeout_seconds`. On Modal and SkyPilot the environment's deployment and readiness happen inside that window, so a candidate could fail on a timeout it was never given the time to avoid.

The direction half of #477 is fixed in #532, which owns the gate.

## Solution

Thread the declared contract from the input bundle down to each candidate evaluation, and let the framework's measurement own fitness.

- `gates.BenchmarkContract` bundles at most one contract form plus the benchmark's own time budget. `declared` says whether either form is configured; `output_argument` gives the flag the benchmark takes its result path under (the scalar contract's own flag, or `PROTOCOL_OUTPUT_FLAG`). That expression previously existed in three places: the agent loop, the evolve loop, and inside `run_benchmark_gate`; all three now read it from the contract.
- `_run_evolve` passes `benchmark_result`, `benchmark_result_protocol`, and `benchmark.timeout_seconds` into `run_evolve_loop`, which builds the contract once and threads it through bootstrap, serial generations, and parallel generations.
- `_run_candidate_gates` runs the accuracy gate and then, when a contract is declared, the benchmark gate, for both the bootstrap seed and every generation child. A failed benchmark fails the candidate with the gate's feedback **before** the profiler runs, so a broken candidate does not pay for the diagnostic pass.
- The gates take the run's `MetricSpace` (from #583) rather than a separate objectives list, so the gate resolves the headline metric and its direction within the same space the frontier and parent selection use.
- `_candidate_fitness` merges the trusted row **over** `ProfilerSummary.metrics`: the contract owns the axes it measures, the profiler keeps the rest. The recorded unit is the evaluator's declared `unit` when the result protocol supplies one, else the profiler's; a metric name is not a unit.
- Evolve's accuracy gate now uses `gates.framework_command_timeout`, the same setup-aware budget the benchmark gate and the agent loop use. That helper moved from `agent/loop.py` into `gates.py` so both loops share it.
- `_run_framework_benchmark_gate` publishes the same `BENCHMARK_RESULT` event the agent loop publishes, so a client watching an evolve run sees the framework's measurement of each candidate.
- The run context is told the contract's `benchmark_output_argument`. This is the pass-through the sandbox uses to allowlist the framework result artifact; without it a remote evolve run's benchmark could not write its result at all.

**Cost.** With a contract declared, each candidate now pays for a benchmark run *and* a profiler run. The profiler is deliberately kept: its analysis, bottlenecks, and suggestions feed the next mutator, and it supplies the axes the contract does not measure. Tasks that declare no contract are unchanged.

### Architecture

```mermaid
flowchart LR
    M[vibesys.input.toml<br/>benchmark.result / result_protocol] --> RE[_run_evolve]
    RE --> BC[gates.BenchmarkContract<br/>declared, output_argument, timeout]
    BC --> CTX[run context<br/>benchmark_output_argument]
    BC --> CG[_run_candidate_gates<br/>accuracy then benchmark]
    SP[MetricSpace #583] --> CG
    CG --> G[gates.run_benchmark_gate]
    G --> O[FrameworkBenchmarkOutcome<br/>name, value, direction, unit, row]
    O --> EV[BENCHMARK_RESULT event]
    O --> CF[_candidate_fitness]
    PS[ProfilerSummary.metrics] --> CF
    CF --> IND[Individual.metrics<br/>trusted row merged over profiler's]
    IND --> FR[Population.frontier / select_parent]
```

### Correctness properties

- With a contract declared, the headline metric and the axes the contract measures come from the framework's own benchmark, never from the profiler agent's self-report. Without a contract, fitness is exactly what it was before.
- The trusted row overrides only the axes it carries. An individual keeps every other axis the profiler reported, so a one-metric contract on a two-axis task leaves the Pareto frontier populated.
- A failed benchmark gate fails the candidate before the profiler is invoked.
- The recorded `perf_unit` is a unit, not a metric name: the evaluator's declared unit when the result protocol supplies one, else the profiler's.
- Both framework gates in evolve extend their declared budget by the run environment's setup allowance, matching the agent loop.
- `BenchmarkContract.output_argument` is the single definition of the benchmark's result-path flag.

### Testing

- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 278 passed

New and updated coverage in `tests/vibesys/loops/evolve/test_evolutionary_loop.py`:

- `test_scalar_contract_keeps_the_profilers_other_axes_on_the_frontier`: a two-objective space with a scalar contract keeps both individuals complete and the frontier non-empty. Verified to fail when the trusted row replaces the profiler's, with the loop logging "Frontier is empty (no individual reported all objective metrics)".
- `test_protocol_contract_records_the_evaluator_declared_unit` and the corrected assertion in `test_benchmark_contract_owns_seed_and_child_fitness`: the recorded unit is the evaluator's declaration, or the profiler's when the scalar contract declares none. Both verified to fail when the metric name is recorded as the unit.
- `test_evolve_accuracy_gate_extends_timeout_by_environment_setup_allowance`, alongside the existing benchmark-gate timeout tests.
- `test_benchmark_contract_owns_seed_and_child_fitness`, `test_benchmark_contract_failure_fails_the_candidate_before_profiling`, `test_no_benchmark_contract_keeps_profiler_fitness`.

No closing keyword: this PR closes no issue. The metric-direction half of
#477 is fixed and closed by #532; this is the evolve wiring that the
original #495 split bundled with it.


vic-lsh:
Review finding:

[P2] Preserve the evaluator's benchmark budget when evolve performs environment setup (`src/vibesys/loops/evolve/loop.py:482-490`).

The evolve path forwards `contract.timeout_seconds` directly to `run_benchmark_gate`. The agent path explicitly adds `ctx.run_environment_view.framework_setup_timeout_seconds`, so environment-owned Modal/SkyPilot deployment and readiness time does not consume the benchmark command's declared runtime. Under evolve, a serving benchmark can time out during setup before receiving its configured execution budget.

Please use the same setup-aware timeout policy in evolve, preferably through shared gate wiring, and cover a run environment with nonzero setup allowance.

vic-lsh:

vic-lsh:
Rebased and fixed this one too, so the stack keeps moving. The change is @AyanBinRafaih's work; below is the rebase and the review findings I addressed on top of it.

## Rebase

Rebased onto #532's head, then onto `main` after #532 squash-merged, so this branch now carries only its own two author commits plus mine. `main` also picked up #591 (`[benchmark.result]` for seven example tasks) and #560; neither conflicted. The only conflict was the `vibesys.loops.gates` import block in `src/vibesys/loops/evolve/loop.py`.

The rebase also surfaced a break the auto-merge hid: `_evaluate_candidate` and `_bootstrap_seed` no longer take an `objectives` list (#583 replaced it with `space: MetricSpace`), so the new gate calls referenced an undefined name. Rather than reintroduce the list, the candidate gates now take the run's `MetricSpace` and read `space.objectives`, which is the same space the frontier and parent selection use.

## Findings addressed

**[P1] `_candidate_fitness` emptied the Pareto frontier on two-axis tasks.** It replaced the profiler's full metric row with the contract's row. The scalar result contract reports exactly one number, so on a task whose `objectives.toml` declares two axes every individual came out missing the second one. `Population.frontier` keeps only individuals carrying a value for every configured axis, so it returned nothing, and Pareto parent selection had no frontier to draw from. The trusted row is now merged **over** `summary.metrics`: the contract owns the axes it measures, the profiler keeps the rest. Regression test `test_scalar_contract_keeps_the_profilers_other_axes_on_the_frontier` uses a two-objective space with a scalar contract; verified to fail on the pre-fix code, where the loop logs "Frontier is empty (no individual reported all objective metrics)".

**[P2] `perf_unit` was being given a metric name.** `_candidate_fitness` returned `benchmark.metric_name` as the unit, and `test_benchmark_contract_owns_seed_and_child_fitness` pinned that as `{"total_ops_per_sec"}`. The result protocol's `hello` already declares a real `unit` per metric, so `FrameworkBenchmarkOutcome` now carries `metric_unit`, `read_protocol_benchmark` fills it from the declaration, and the profiler's unit is the fallback (the scalar contract names a metric and nothing else, so it declares no unit). The wrong assertion is corrected and `test_protocol_contract_records_the_evaluator_declared_unit` covers the declared case; both fail if the metric name is recorded as the unit.

**[P2] Evolve's accuracy gate did not extend its budget by the environment setup allowance.** The benchmark gate this PR adds does, and the agent loop does; the accuracy gate passed the manifest timeout straight through, so on Modal or SkyPilot the deployment and readiness time was charged to the candidate's accuracy budget. It now uses `gates.framework_command_timeout` like the others, covered by `test_evolve_accuracy_gate_extends_timeout_by_environment_setup_allowance`.

**[P3] The evolve gate wrapper published no `BENCHMARK_RESULT`.** The agent loop emits one for every trusted measurement; an evolve run gave a client nothing but the agent transcripts around the benchmark. `_run_framework_benchmark_gate` now emits the same event with a `gen-N-cand-K` round label.

**[P3] The benchmark output flag was written out three times.** The agent loop, the evolve loop, and `run_benchmark_gate` each restated "the scalar contract's `json_argument`, else `PROTOCOL_OUTPUT_FLAG`, else nothing". It is now `BenchmarkContract.output_argument`, and all three read it from there.

## Body and title

The title said `TUI:`, which this change is not, and the body's lead was the metric-direction narrative that belongs to #532 (whose merge closed #477). The body now describes what this PR actually does: the evolve wiring, the `benchmark_output_argument` pass-through that the sandbox needs to allowlist the result artifact and which the old body did not mention, and the cost, which is one benchmark run plus one profiler run per candidate once a contract is declared. `Closes #477` moved to #532; this PR says "Part of #477".

The profiler is deliberately still invoked when a contract is declared: its analysis feeds the next mutator, and it supplies the axes the contract does not measure. That is the same reason the row merge matters.

## Checks

- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 278 passed

PR #534, opened 2026-09-01
Record who produced each round's headline metric

AyanBinRafaih:
> Stacked PR 3/6. Base branch: `main` (#532 and #533 have merged). #535, #536, and #537 stack on top in that order.

## Problem

A round could write `official_evaluation=True` with a `perf_metric` the agent reported about its own work, and nothing on the record said which. `official_evaluation` means "the framework's own gates ran this round"; it does not say who produced the headline number, because the gates can run in a round whose metric is still the implementer's self-report. Every downstream consumer (hypothesis resolution, baselines, measurements, retention) had only that flag to go on, so an agent's own claim could seed a baseline or stand in as an official measurement. The gap starts at the record itself: there was no field for where a round's headline metric came from.

## Solution

Record provenance on every round record and stamp it wherever the agent loop produces a headline metric.

- `vs_loop_state` exports `PerfProvenance = Literal["framework", "implementer"]`, alongside `JudgeVerdict`, and `RoundRecord` gains `perf_provenance: PerfProvenance | None`. One exported definition, so the loop and the consumers that stack above this PR name the closed set rather than restating its literals.
- The agent loop stamps `"framework"` when the benchmark gate produced the number and `"implementer"` when the agent self-reported it, on both the single-agent and multi-agent paths. The stamped value rides on the `RoundRecord` written to the write-ahead journal.
- `None` marks a legacy record written before provenance was tracked. Those stay trusted, so reprojecting an old run does not rewrite its historical resolutions.
- `official_evaluation`'s comment is corrected: it says the framework's gates ran, and consumers deciding whether to trust a measurement read `perf_provenance` instead.

**Out of scope.** Nothing gates on the field yet; the consumers ship in #535. The operator-facing projection is also out of scope: `HypothesisRound` does not surface provenance, so the TUI cannot yet show an operator that a number was self-reported. That needs its own change.

### Architecture

```mermaid
flowchart LR
    G[framework benchmark gate] --> A[round record<br/>perf_provenance=framework]
    S[agent self-report] --> A2[round record<br/>perf_provenance=implementer]
    L[record written before #534] --> AN[perf_provenance=None<br/>stays trusted]
    A --> J[write-ahead journal / RoundRecord]
    A2 --> J
    AN --> J
    J -. consumed in #535 .-> R[resolution, retention, baselines]
```

The agent loop owns provenance stamping; `vs_loop_state` owns the closed set and the field; nothing gates on it in this PR.

### Correctness properties

- On every record the framework writes now, `perf_provenance` is non-None whenever `perf_metric` is non-None: `"framework"` from a benchmark gate, `"implementer"` from a self-report. `None` only ever identifies a legacy record.
- `official_evaluation` and `perf_provenance` are independent: a round can have the gates run and still carry an implementer-produced number.
- The field is additive loop state. Existing resolutions are unchanged at this layer and legacy records keep their historical values.
- The field is confined to persisted loop state. `perf_provenance` appears nowhere under `src/server/` or `clients/`: no event payload carries it and no client type references it, so nothing on the wire changes.

### Testing

- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 280 passed

New loop-level coverage in `tests/vibesys/loops/agent/test_orchestrate.py`, driven through the orchestrate harness rather than asserting the field in isolation:

- `test_round_records_stamp_who_produced_the_headline_metric`: an implementer-reported number is stamped `implementer` even though `official_evaluation` is true.
- `test_framework_measured_round_is_stamped_framework`: the framework benchmark's number overrides the agent's and is stamped `framework`.
- `test_a_round_with_no_headline_metric_carries_no_provenance`: the invariant's other half.

The first two were verified to fail with the stamp removed from the record write.

Part of #479. The issue also asks that an operator be able to see which numbers are self-reported; that projection is not in this PR.

### DIFF
--- a/libs/vs-loop-state/src/vs_loop_state/__init__.py
+++ b/libs/vs-loop-state/src/vs_loop_state/__init__.py
@@ -7,6 +7,7 @@ all three loops. Project layout and filesystem persistence belong to
 
 from vs_loop_state.agent import (
     JudgeVerdict,
+    PerfProvenance,
     RoundHistory,
     RoundRecord,
     parse_round_record,
@@ -33,6 +34,7 @@ __all__ = [
     "IndividualRecord",
     "JudgeVerdict",
     "MetricComparison",
+    "PerfProvenance",
     "PlainLoopCursor",
     "PlainPerformanceRecord",
     "PlainPerformanceSnapshot",
--- a/libs/vs-loop-state/src/vs_loop_state/agent.py
+++ b/libs/vs-loop-state/src/vs_loop_state/agent.py
@@ -49,6 +49,12 @@ if TYPE_CHECKING:
 #: is the only value compatible with an unreviewed round.
 JudgeVerdict = Literal["pass", "fail", "deferred"]
 
+#: Who produced a round's headline ``perf_metric``. ``framework`` means a
+#: framework-owned benchmark result contract measured it; ``implementer`` means
+#: the agent reported it about its own work. Only the framework may write
+#: ``framework``, so this is the trust boundary every consumer branches on.
+PerfProvenance = Literal["framework", "implementer"]
+
 
 @dataclass(config=ConfigDict(extra="forbid", populate_by_name=True, serialize_by_alias=True))
 class RoundRecord:
@@ -92,9 +98,13 @@ class RoundRecord:
     hypothesis_parent_commit: str | None = None
     metrics: dict[str, float] = Field(default_factory=dict)
     evaluation_artifact: str | None = None
-    # Only framework-owned gates can set this. A judge-approved hypothesis may
-    # be a useful provisional working checkpoint without becoming the latest
-    # officially verified checkpoint.
+    # Whether the framework's own gates ran this round. Only framework-owned
+    # gates can set it: a judge-approved hypothesis may be a useful provisional
+    # working checkpoint without becoming the latest officially verified
+    # checkpoint. It says nothing about who produced ``perf_metric`` -- the
+    # gates can run in a round whose headline number is still the agent's own
+    # report -- so consumers deciding whether to trust a measurement read
+    # ``perf_provenance``, not this flag.
     official_evaluation: bool = False
     official_evaluation_reason: str | None = None
     # Candidate evidence is deliberately separate from official tracking. It
@@ -125,6 +135,13 @@ class RoundRecord:
     # framework stored the comparison; readers then re-derive it from the
     # run's persisted metric space.
     perf_comparison: MetricComparison | None = None
+
+    # Who produced ``perf_metric``. ``None`` identifies a legacy record written
+    # before provenance was tracked; consumers treat those as trusted so
+    # reprojection does not rewrite their historical resolutions. On every
+    # record the framework writes now, ``perf_provenance`` is non-None whenever
+    # ``perf_metric`` is non-None.
+    perf_provenance: PerfProvenance | None = None
 
     # Implementer attribution: which driver/provider/model produced this round
     # attempt. The provider session ID itself is machine-local and never lives
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -108,7 +108,7 @@ from vibesys.skills import (
     build_skill_catalog,
     resolve_skill_selections,
 )
-from vs_loop_state.agent import RoundHistory, RoundRecord
+from vs_loop_state.agent import PerfProvenance, RoundHistory, RoundRecord
 from vs_project import AgentRunConfiguration
 
 # Candidate process boundaries selected by ``--interface``. Language, tooling,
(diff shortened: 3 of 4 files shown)


vic-lsh:

vic-lsh:
Rebased and fixed this one as well. The change is @AyanBinRafaih's work; below is the rebase and the review findings I addressed on top of it.

## Rebase

Rebased onto #533's head, then onto `main` once #533 merged as `cc10504`. Every conflict was keep-both, in the two places `RoundRecord` grows fields:

- `libs/vs-loop-state/src/vs_loop_state/agent.py`: the new field landed in the same block as #579's `perf_comparison` and, after the second rebase, #523's `implementer_driver`/`implementer_provider`/`implementer_model`.
- `src/vibesys/loops/agent/loop.py`: the same three fields in the `RoundRecord(...)` construction.
- `tests/vibesys/loops/agent/test_orchestrate.py`: #523's new test and this PR's new tests were appended at the same point in the file.

## Findings addressed

**[P1] The field and its writer had no test.** The PR added `perf_provenance`, stamped it in two code paths, and asserted nothing about either. Three loop-level cases now go through the orchestrate harness in `tests/vibesys/loops/agent/test_orchestrate.py`:

- `test_round_records_stamp_who_produced_the_headline_metric`: a round whose headline number is the implementer's self-report is stamped `implementer`, while `official_evaluation` is true. That pairing is the whole point of the field.
- `test_framework_measured_round_is_stamped_framework`: a round the framework benchmark measured is stamped `framework`.
- `test_a_round_with_no_headline_metric_carries_no_provenance`: the other half of the invariant.

The first two were verified to fail with the stamp removed from the record write.

**[P2] The closed set was an inline `Literal` restated at each site.** `vs_loop_state` now exports `PerfProvenance = Literal["framework", "implementer"]` next to `JudgeVerdict`, and `RoundRecord` and the agent loop both use it. #535's trust predicate will use the same alias rather than a fourth copy of the literals.

**[P2] The `official_evaluation` comment contradicted the new field.** It read "Only framework-owned gates can set this", which is true but reads as a claim about the metric's source, and the new field's comment then said the opposite. `official_evaluation` now documents what it actually means, that the framework's gates ran this round, and points at `perf_provenance` for the trust question.

**[P3] The invariant was implied but not written down.** The field's comment now states it: on every record the framework writes now, `perf_provenance` is non-None whenever `perf_metric` is non-None. I checked every branch of the round-record write and it holds; the third test covers the None case.

## Body and title

Title lost its `TUI:` prefix. The body now says explicitly that the operator-facing projection is out of scope: `HypothesisRound` does not surface provenance, so the TUI still cannot show an operator which numbers were self-reported. #479 asks for that too, so the PR says "Part of #479" rather than closing it. If you would rather the issue close here, the projection needs its own change first.

## Checks

- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 280 passed

vic-lsh:
CI caught a real gap in my own coverage of this change, not a test failure: `diff-cover` reported 66% patch coverage (9 new lines, 3 missing) against a threshold of 80. The missing lines were `src/vibesys/loops/agent/loop.py` 2945, 2967, and 2970, which are the single-agent path's stamping branches. My three loop-level tests all drove the multi-agent path, so the `--inner-loop=single-agent` half of the change was asserted nowhere.

That is worth fixing rather than waiving, because the single-agent ablation is exactly where the trust question bites hardest: one agent implements, self-judges, and profiles, so its `perf_metric` is a self-report by construction, and nothing about that round makes it more trustworthy than the multi-agent implementer's number.

Added, mirroring the three multi-agent cases through the same orchestrate harness:

- `test_single_agent_self_reported_metric_is_stamped_implementer`
- `test_single_agent_framework_benchmark_overrides_and_stamps_framework`
- `test_single_agent_round_without_a_metric_carries_no_provenance`

The first two were verified to fail with the two single-agent stamps removed.

The harness could not answer a single-agent round at all before this: `_make_orchestrate_runner` dispatched only on `kind`, and the loop asks for `SingleAgentRoundResponse` under `kind="implementer"`. It now takes a `single_agent_perf_metrics` queue and returns a combined response through a `_single_agent_round` helper, so the ablation is drivable from the harness for future tests too. `_make_orchestrate_runner` picks up a `C901` noqa, matching how the evolve loop's `_make_runner` harness is already annotated.

Reproduced the check locally after the change: `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80` now reports 9 lines, 0 missing, 100%, with both changed files at 100%.

Also rebased onto current `main` (`2a3486c`, after #497), which had no conflicts with this diff.

Checks: `./scripts/check_format.sh`, `./scripts/check_lint.sh`, `./scripts/check_types.sh` clean; `uv run pytest tests/vibesys/loops -q --no-cov` 611 passed and `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov` 299 passed.

PR #535, opened 2026-09-01
Resolve a hypothesis unmeasured when the framework measured nothing

### LINKED ISSUE #475: TUI: Hypothesis resolves PROVEN with no measurement, and renders as "Accepted"

AyanBinRafaih:
> Stacked PR 4/6. Base branch: `main` (#532, #533, and #534 have merged). #536 and #537 stack on top.

## Problem

A hypothesis could resolve `PROVEN` on the agent's word alone. `resolve_hypothesis_outcome` treated the absence of a measurement as permission to accept a supportive declaration, so a "proven" hypothesis could carry no empirical weight, and the TUI rendered it as "Accepted".

#534 gave round records `perf_provenance`, but nothing consumed it. An agent self-reported number could still prove a hypothesis, seed a baseline, count as an official measurement, or dominate a later candidate in the Pareto archive.

## Solution

A hypothesis is proven by a measurement the framework took, or it is not proven.

- `HypothesisResolution.UNMEASURED` is a new resolution: the review passed but the framework measured nothing, so the empirical claim is neither proven nor failed. It is distinct from `INCONCLUSIVE`, which reports a trusted measurement that did not decide the claim.
- The trust boundary is applied **where the comparison is recorded**, not at the resolution call. #579 made the round store the `MetricComparison` the framework made; an implementer-reported number is now never ordered at all, so `perf_comparison` stays `None` and resolution has nothing to work from. `ResolutionEvidence` reduces to "no comparison means `UNMEASURED`, otherwise the comparison decides", and `benchmark_expected` is deleted (`main` fed it from two different facts: `record.official_evaluation` in one caller and `framework_benchmark_configured` in the other).
- `round_comparison` gains a provenance guard **before** its re-derivation fallback. An implementer round stores no comparison, which on disk is indistinguishable from a record written before #579; without the guard, `--resume` and the server read path would re-derive one and disagree with what the loop recorded.
- One exported predicate, `trusted_perf_provenance`, over #534's `PerfProvenance` closed set, is the only place the rule lives. It gates the five loop sites (retention, prior readings, the recorded delta, trusted Pareto-parent selection, and the provisional-candidate cadence) plus baseline selection and measurement projection in `hypotheses.py`.
- `_detect_plateau` stops counting implementer numbers. Telling the orchestrator it has plateaued on the strength of its own reports is a feedback loop: the agent would be reacting to numbers it chose. **This has a cost worth naming:** a task that declares no benchmark result contract now gets no plateau warning at all, because the framework has measured nothing to detect a plateau in. That is the same trade the rest of this PR makes, applied consistently, but it does remove a signal those runs used to get.
- `UNMEASURED` is not a failure: it does not feed strategy pruning or hypothesis retirement, and it consumes provisional-candidate cadence budget like `PROVEN`, because it is exactly the checkpoint the next official evaluation must measure.
- Legacy records carry no provenance and stay trusted, so reprojecting a run recorded before #534 does not rewrite its historical resolutions.

### Architecture

```mermaid
flowchart LR
    P[perf_provenance from #534] --> TP[trusted_perf_provenance<br/>the one rule]
    TP --> PC[perf_comparison recorded<br/>None when self-reported]
    PC --> RE[ResolutionEvidence.comparison]
    RE --> R[resolve_hypothesis_outcome<br/>None then UNMEASURED<br/>else the comparison decides]
    TP --> RC[round_comparison guard<br/>before re-derivation]
    RC --> RP[resume and server read path<br/>agree with the record]
    TP --> L[retention, delta, prior readings,<br/>Pareto parents, cadence, plateau]
```

### Correctness properties

- A hypothesis never resolves `PROVEN` without a comparison the framework made. Without one a supportive declaration resolves `UNMEASURED`.
- An implementer-reported number is never ordered, never seeds a baseline, never enters a retention or dominance decision, and never contributes to plateau detection.
- Reprojection agrees with the round record. `--resume` and the server read path resolve a self-reported round `UNMEASURED`, the same value the loop wrote.
- Legacy records (`perf_provenance` is `None`) stay trusted and keep their historical resolutions.
- `UNMEASURED` is not a failed outcome for strategy pruning or retirement, and counts like `PROVEN` for the provisional-candidate cadence.

**Not a property**, despite the earlier draft of this body claiming it: "`INCONCLUSIVE` implies a trusted measurement". An agent can declare `HypothesisOutcome.INCONCLUSIVE`, and that declaration maps straight to `HypothesisResolution.INCONCLUSIVE` without consulting any measurement. `INCONCLUSIVE` therefore means either "a trusted measurement did not decide the claim" or "the agent said it was inconclusive".

## Policy this implements

Strict, by design. Framework provenance is stamped only for a benchmark the framework itself ran through a readable result contract. A task that declares neither a `[benchmark.result]` spec nor `result_protocol` has no trusted number, so its rounds resolve `UNMEASURED` rather than being credited with the implementer's self-report. That is the intended outcome, not a gap: #591 declared contracts for seven example tasks, and the six latency examples whose headline metric is a nested percentile with a `min` direction remain `UNMEASURED` until their `result_protocol` migration.

## Protocol change

`unmeasured` is a new member of `HypothesisResolution`, which crosses the backend boundary. `clients/backend-client/src/generated/protocol.{schema.json,generated.ts}` are regenerated through `pnpm run generate:protocol`; the diff is that one additive enum member in each. The TUI's two consumers, `outcomeColor` and `outcomeLabel`, already fall through correctly, so an unmeasured hypothesis renders as body-text "Unmeasured" rather than being forced into a verdict color. `outcomeColor`'s comment, which listed the outcomes with no verdict reading, is updated to name it.

The Pareto-archive prompt is also corrected: it told the orchestrator that a row outside the trusted frontier was "awaiting independent review", which after this change may instead be because the row's numbers are the implementer's own.

## Testing

- `uv run pytest tests/vibesys/loops -q --no-cov`: 617 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 299 passed
- `cd clients/tui && bun test --max-concurrency 1 src`: 513 passed
- `pnpm check:clients`: clean
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 14 lines, 0 missing, 100%

Both directions of the rule are covered, because a strict rule is only correct if the gate still opens:

- `test_framework_measured_improvement_resolves_proven` drives two rounds through the loop with a framework-stamped benchmark (100 then 200) and asserts the second resolves `proven`. Every other end-to-end expectation in that file reads `unmeasured`, so without this the change could have made `proven` unreachable rather than honest.
- `test_reprojection_of_a_self_reported_round_stays_unmeasured` pins the `round_comparison` guard through `reproject_run_evidence`. Verified to fail without the guard, resolving `INCONCLUSIVE` where the loop recorded `UNMEASURED`.
- `test_a_self_reported_improvement_never_resolves_proven` and `test_a_self_reported_round_is_not_a_baseline_for_a_later_trusted_one` cover the live path and baseline selection.
- `test_implementer_report_cannot_seed_archive_or_dominate_candidates` covers trusted Pareto-parent selection and dominance.

Closes #475.

### DIFF
--- a/clients/backend-client/src/generated/protocol.generated.ts
+++ b/clients/backend-client/src/generated/protocol.generated.ts
@@ -395,7 +395,7 @@ export type HypothesisOutcome1 =
  * Framework-owned resolution after all available evidence is known.
  */
 export type HypothesisResolution =
-  "proven" | "disproven" | "inconclusive" | "implementation_failed" | "blocked" | "rejected";
+  "proven" | "disproven" | "inconclusive" | "implementation_failed" | "blocked" | "rejected" | "unmeasured";
 export type JudgeVerdict1 = ("pass" | "fail" | "deferred") | null;
 export type PerfMetric2 = number | null;
 export type PerfUnit2 = string | null;
--- a/clients/backend-client/src/generated/protocol.schema.json
+++ b/clients/backend-client/src/generated/protocol.schema.json
@@ -1597,7 +1597,8 @@
         "inconclusive",
         "implementation_failed",
         "blocked",
-        "rejected"
+        "rejected",
+        "unmeasured"
       ],
       "title": "HypothesisResolution",
       "type": "string"
--- a/clients/tui/src/ui/experiment-log.ts
+++ b/clients/tui/src/ui/experiment-log.ts
@@ -645,8 +645,10 @@ export function entryRow(entry: HypothesisEntry, columns: Columns): string {
 
 /**
  * Green for a hypothesis that held, red for one that did not, and the active
- * accent while it is still open. Outcomes with no such reading (``continue``,
- * ``inconclusive``) stay in body text rather than being forced into a verdict.
+ * accent while it is still open. Outcomes with no such reading stay in body
+ * text rather than being forced into a verdict: `inconclusive`, where a
+ * trusted measurement did not decide the claim, and `unmeasured`, where the
+ * framework measured nothing to decide it with.
  */
 export function outcomeColor(theme: Theme, entry: HypothesisEntry): string {
   const outcome = entry.resolved_outcome ?? null;
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -339,14 +339,22 @@ def round_comparison(
 ) -> MetricComparison | None:
     """Return how one round's headline reading compared with its baseline.
 
-    ``None`` means the round recorded no official metric, so there was nothing
-    to order. A record that carries ``perf_comparison`` answers for itself: the
-    framework decided it once, when the round was written, and re-deriving it
-    later from a possibly re-configured space would let a resumed run disagree
-    with what it recorded. Older records have the comparison derived from
-    *space* instead.
+    ``None`` means there is nothing to order: the round recorded no official
+    metric, or the number it recorded is the implementer's own report. A record
+    that carries ``perf_comparison`` answers for itself: the framework decided
+    it once, when the round was written, and re-deriving it later from a
+    possibly re-configured space would let a resumed run disagree with what it
+    recorded. Older records have the comparison derived from *space* instead.
+
+    The provenance guard has to come before that re-derivation. A round the
+    implementer self-reported stores no comparison, which is indistinguishable
+    from a pre-#579 record; without this guard the fallback would re-derive one
+    from the space and resume, and the server read path, would resolve the
+    hypothesis INCONCLUSIVE where the loop resolved it UNMEASURED.
     """
     if not record.official_evaluation or record.perf_metric is None:
+        return None
+    if not trusted_perf_provenance(record.perf_provenance):
         return None
     if record.perf_comparison is not None:
         return record.perf_comparison
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -23,7 +23,19 @@ from vibesys.schemas import (
 if TYPE_CHECKING:
     from collections.abc import Sequence
 
-    from vs_loop_state import RoundRecord
+    from vs_loop_state import PerfProvenance, RoundRecord
+
+
+def trusted_perf_provenance(provenance: PerfProvenance | None) -> bool:
+    """Whether a round's headline metric may drive a framework decision.
+
+    The one trust rule for :data:`~vs_loop_state.PerfProvenance`, shared by
+    hypothesis resolution, scalar and Pareto retention, the recorded delta, and
+    trusted Pareto-parent selection. Legacy records carry no provenance and stay
+    trusted, so reprojecting an old run does not rewrite its historical
+    resolutions; only an explicit agent self-report is untrusted.
+    """
+    return provenance != "implementer"
 
 
 @dataclass(frozen=True)
@@ -31,23 +43,30 @@ class ResolutionEvidence:
     """Inputs the framework needs to finalize one hypothesis declaration.
 
     ``comparison`` is the round's headline reading ordered against its causal
-    baseline. ``None`` means no official metric was recorded, which is a
-    different fact from ``MetricComparison.INCOMPARABLE``: one was recorded but
-    could not be ordered. Resolution consumes the comparison and never the
-    readings, the axis direction, or the measurement tolerance behind it.
+    baseline, and it is set only for a reading the framework measured itself.
+    ``None`` therefore covers both "no official metric was recorded" and "the
+    number on the record is the implementer's own report", which is a
+    different fact from ``MetricComparison.INCOMPARABLE``: there a trusted
+    reading exists but could not be ordered. Resolution consumes the
+    comparison and never the readings, the axis direction, or the measurement
+    tolerance behind it.
     """
 
     declared: HypothesisOutcome | None
     passed: bool
     reviewed: bool
     comparison: MetricComparison | None
-    benchmark_expected: bool = False
 
 
 def resolve_hypothesis_outcome(
     evidence: ResolutionEvidence,
 ) -> HypothesisResolution | None:
-    """Resolve one declaration only after review and trusted evidence."""
+    """Resolve one declaration only after review and trusted evidence.
+
+    A supportive declaration (``SUPPORTED``/``NOMINATED``) never resolves
+    ``PROVEN`` on the agent's word alone: without a trusted measurement it
+    resolves ``UNMEASURED``, and with one the measurement decides.
+    """
     if not evidence.reviewed:
         resolution = None
     elif not evidence.passed:
@@ -64,14 +83,10 @@ def resolve_hypothesis_outcome(
         if evidence.declared is HypothesisOutcome.CONTINUE:
             resolution = None
         elif resolution is None:
-            if not evidence.benchmark_expected:
-                resolution = HypothesisResolution.PROVEN
-            elif evidence.comparison is not None:
-                resolution = _resolve_metric_evidence(evidence.comparison)
-            elif evidence.declared is HypothesisOutcome.SUPPORTED:
-                resolution = HypothesisResolution.PROVEN
+            if evidence.comparison is None:
+                resolution = HypothesisResolution.UNMEASURED
             else:
-                resolution = HypothesisResolution.INCONCLUSIVE
+                resolution = _resolve_metric_evidence(evidence.comparison)
     return resolution
 
 
@@ -117,7 +132,10 @@ def metric_baseline(
     comparable = [
         item
         for item in rounds
-        if item.official_evaluation and item.perf_metric is not None and item.perf_unit == metric
+        if item.official_evaluation
+        and item.perf_metric is not None
+        and trusted_perf_provenance(item.perf_provenance)
+        and item.perf_unit == metric
     ]
     if parent_commit is not None:
         return next(
@@ -296,7 +314,6 @@ def project_round_evidence(
                 reviewed=updated.review
                 not in {HypothesisReview.PENDING, HypothesisReview.DEFERRED},
                 comparison=comparison,
-                benchmark_expected=record.official_evaluation,
             )
         )
     else:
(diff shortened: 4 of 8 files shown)


vic-lsh:
Review finding:

[P1] Apply provenance filtering to retention, deltas, and trusted Pareto selection, not only hypothesis resolution (`src/vibesys/loops/agent/loop.py:3157-3203`).

The PR excludes `perf_provenance == 'implementer'` only when building `ResolutionEvidence`. The same untrusted `official_metric` and `accepted_metrics` still drive `candidate_retained` and `perf_delta_pct`. This is reachable in objective-based tasks with an accuracy command but no benchmark result contract: an implementer self-report is stamped with implementer provenance, yet an undominated row persists `candidate_retained=True`. `_trusted_candidate_records` then accepts it as a trusted Pareto parent because it checks retention but not provenance.

Please gate scalar/Pareto retention, delta computation, and trusted metric-row selection on framework provenance. Add a regression test showing that a reviewed, accuracy-passing implementer report cannot seed the archive or influence later dominance decisions.

vic-lsh:

vic-lsh:
Rebased and fixed this one too. The change is @AyanBinRafaih's work; below is the rebase and the review findings I addressed on top of it.

## Rebase

This one needed more than a mechanical rebase. #579 landed between this PR being written and now, and it changed the shape this PR builds on: a round record stores the `MetricComparison` the framework made, and resolution consumes that comparison rather than a value, a direction, and a tolerance. The PR's `ResolutionEvidence(official_metric, baseline_metric, direction, noise_fraction)` no longer exists on `main`, so I ported the change onto the new shape rather than resolving the conflicts line by line.

## Findings addressed

**[P1] Ported onto #579's stored comparison, and moved the trust gate to where the comparison is recorded.** The PR gated provenance at the `resolve_hypothesis_outcome` call, passing `official_metric=None` for an implementer round. On the current shape the natural place is one level up, in `loop.py` where `perf_comparison` is computed: an implementer-reported number is never ordered at all, so the stored comparison is `None`. `ResolutionEvidence` then reduces to "no comparison means `UNMEASURED`, otherwise the comparison decides", which is the whole rule in two lines. `benchmark_expected` is deleted; `main` fed it from two different facts, `record.official_evaluation` in `hypotheses.py` and `framework_benchmark_configured` in `loop.py`.

**[P1] Required guard: `round_comparison` re-derived a comparison for implementer rounds.** When a record's stored `perf_comparison` is `None`, `round_comparison` falls back to deriving one from the run's metric space. An implementer round stores `None` for a different reason, and on disk the two are indistinguishable, so `--resume` and the server read path would have disagreed with what the loop recorded. Added an early `return None` for implementer provenance, ahead of the fallback. `test_reprojection_of_a_self_reported_round_stays_unmeasured` pins it through `reproject_run_evidence`; verified to fail without the guard, resolving `INCONCLUSIVE` where the loop recorded `UNMEASURED`.

**[P2] Three copies of the trust rule became one.** The PR added `_is_trusted_provenance` in `loop.py` and `_trusted_measurement` in `hypotheses.py`, both spelling out `!= "implementer"`. Both are gone; `trusted_perf_provenance` in `hypotheses.py`, over #534's exported `PerfProvenance`, is the only definition, and every site calls it.

**[P2] The positive case was untested.** Every end-to-end expectation in `test_orchestrate.py` reads `unmeasured`, so nothing in the suite would have caught it if the change had made `proven` unreachable rather than honest. `test_framework_measured_improvement_resolves_proven` drives two rounds with a framework-stamped benchmark (100 then 200) and asserts the second resolves `proven` off a `better` comparison.

**[P2] `_detect_plateau` counted implementer numbers.** Telling the orchestrator it has plateaued on the strength of its own self-reports is a feedback loop: the agent would be reacting to numbers it chose. It now only counts framework-measured rounds.

Worth flagging explicitly, because it is a behavior loss and not just a correctness fix: a task that declares no benchmark result contract now gets **no plateau warning at all**. `test_loop_threads_plateau_warning_into_prompt` was driving the plateau path with `implementer_perf_metrics`, so it caught this immediately; I converted it to drive the same flat trajectory through the benchmark gate, and added `test_self_reported_numbers_do_not_raise_a_plateau_warning` to pin the new behavior with the reasoning in its docstring. If you would rather keep the warning for unmeasured runs (with different wording, say), that is a one-line change to the filter and I am happy to make it instead.

**[P3] The Pareto-archive prompt mislabelled why rows are excluded.** `_pareto_archive_summary` told the orchestrator that a row outside the trusted frontier was "awaiting independent review". After this change a row can land there because its numbers are the implementer's own, which is a different thing to tell an agent. Reworded to name both reasons; `test_pareto_archive_distinguishes_trusted_and_pending_candidates` had pinned the old phrase and now asserts both.

**[P3] The TUI's `outcomeColor` comment** listed the outcomes with no verdict reading and did not know about `unmeasured`. Updated.

## One thing the review missed: this is a protocol change

`unmeasured` is a new member of `HypothesisResolution`, which crosses the backend boundary, so `tests/server/test_protocol_schema.py` fails without regenerating the committed schema. The PR did not, so CI would have failed on it regardless of the rebase. Regenerated through `pnpm run generate:protocol`; the diff is exactly one additive enum member in `protocol.schema.json` and in the generated TypeScript literal union.

That makes the client change more than comment-only, so I ran the client suites: `pnpm check:clients` clean and `bun test --max-concurrency 1 src` in `clients/tui` 513 passing. The two consumers, `outcomeColor` and `outcomeLabel`, already fall through correctly, so an unmeasured hypothesis renders as body-text "Unmeasured" rather than being forced into a verdict color. No behavioral client change was needed.

## Body

Rewritten around the ported design, and it now states the policy plainly: strict by design, tasks without a readable result contract resolve `UNMEASURED`, #591 covered seven examples, and the six latency examples remain `UNMEASURED` until their `result_protocol` migration.

I also removed a stated correctness property that is not true: "`INCONCLUSIVE` always implies a trusted measurement existed". An agent can declare `HypothesisOutcome.INCONCLUSIVE`, and that maps straight to `HypothesisResolution.INCONCLUSIVE` without consulting any measurement. The body now says so.

## Checks

- `uv run pytest tests/vibesys/loops -q --no-cov`: 617 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 299 passed
- `cd clients/tui && bun test --max-concurrency 1 src`: 513 passed; `pnpm check:clients` clean
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 14 lines, 0 missing, 100%

PR #536, opened 2026-09-01
Resolve a baseline when the causal parent is a provisional round

AyanBinRafaih:
> Stacked PR 5/6. Base branch: `main` (#532 through #535 have merged). #537 stacks on top.

## Problem

A performance baseline never resolved on an official round at any `official_eval_every` above 1.

`metric_baseline` demanded an exact parent-commit match among official rounds. `parent_commit` is round N-1's commit, and for any cadence >= 2 a round being official implies round N-1 was provisional, so the exact match could never hit an official record. Replaying eight rounds at the default cadence of 3, `perf_delta_pct` is `None` in every round, and a baseline resolves only on rounds that have no official metric to compare it against.

The comparable filter also required `perf_unit == metric`, so a record whose headline unit was later renamed was dropped even when its `metrics` row still carried the value for that axis.

## Solution

Keep preferring the exact causal baseline; fall back to the newest trusted official measurement that does not postdate the parent round.

- `metric_baseline` prefers the exact parent-commit match. When the parent is provisional, it falls back to the newest trusted official measurement with `round_number <= parent_round`, so a comparison never inverts cause and effect by measuring against a later round.
- A named parent commit that matches nothing, **with no parent round to bound the search**, now returns nothing rather than falling through to the newest measurement. That fall-through could hand back a round recorded after the parent, which is exactly the inversion the bound exists to prevent. Failing closed there means the hypothesis resolves `UNMEASURED` (#535) rather than against an arbitrary comparison.
- The comparable filter accepts a record when `perf_unit` matches the metric **or** the metric is present in the record's `metrics` row, so a renamed headline unit still answers for the axis it measured. Trust is unchanged: `trusted_perf_provenance` still means an implementer-reported number never seeds a baseline.
- `baseline_commit` reports the commit of the record the baseline value actually came from, symmetric with `baseline_round`. It previously reported `hypothesis_parent_commit`, which is correct only when the exact parent match wins; whenever the fallback supplies the baseline (the case this PR exists for) it named a commit whose measurement was never used.
- Retention and baseline selection now admit the same rounds. `_retained` matched on the headline unit alone while `metric_baseline` also accepts the objective row, so with a renamed unit the two disagreed about how much history existed: resolution compared against the renamed round while retention behaved as if it were the first reading ever taken, which it always retains.
- One reader, `record_metric_value`, serves both. It replaces `_baseline_metric_value` here and `_metric_value` in the agent loop, which were two spellings of the same lookup.

### Architecture

```mermaid
flowchart LR
    P[parent round / commit] --> EX{exact parent-commit<br/>trusted match?}
    EX -- yes --> BX[use exact baseline]
    EX -- no --> RB{parent round known?}
    RB -- yes --> FB[newest trusted official<br/>round_number &lt;= parent_round]
    RB -- no --> NC[fail closed: no baseline]
    BX --> RV[record_metric_value<br/>objective row, else headline]
    FB --> RV
    RV --> D[baseline value, commit, delta]
    RV --> RET[retention history<br/>same admissibility rule]
```

### Correctness properties

- A baseline never postdates the round compared against it: the exact parent commit first, else the newest trusted official measurement at or before the parent round, else nothing.
- A provisional-parent chain (any `official_eval_every` above 1) resolves a baseline through the fallback instead of producing none.
- `baseline_commit` names the record the baseline value came from, not the hypothesis's parent.
- A renamed headline unit still resolves a baseline through the objective row, and retention walks the same history that resolution does.
- An implementer-reported measurement is never eligible as a baseline (unchanged from #535).

### Testing

- `uv run pytest tests/vibesys/loops -q --no-cov`: 621 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 299 passed
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 15 lines, 0 missing, 100%

Each of the three fixes above its own regression test, all verified to fail at the merge base:

- `test_provisional_parent_baseline_now_resolves_via_fallback` asserts the reported `baseline_commit` is the official round's, and explicitly not the provisional parent's. Pre-fix it reported the provisional commit.
- `test_an_unfindable_parent_commit_with_no_round_bound_fails_closed`, contrasted against no parent linkage at all, which legitimately takes the newest measurement.
- `test_retention_sees_the_same_history_the_baseline_does` drives a renamed headline unit through the derived-retention path and asserts a worse reading is not retained. Pre-fix, retention saw no history and retained it.

Part of #478. This PR fixes the baseline-lookup half of the issue. The issue's other half, telling an unresolved baseline apart from a measured non-improvement in the operator-facing projection, is only partly closed elsewhere:

- `perf_baseline_value`, `perf_metric_name`, and `perf_direction` **are** projected by `src/server/api/experiments.py::_entry`, and the TUI renders `Baseline <value> <unit>` in the hypothesis drill-down (`measurementMetadata` in `clients/tui/src/ui/experiment-log.ts`). That landed with #465. An earlier revision of this body said otherwise, quoting #478's grounding at commit `3a2fd8b` without re-checking it; that was wrong and is corrected here.
- `baseline_round` and `baseline_commit` are still not projected, and `formatMeasured` still falls back to a bare absolute number when `delta_pct` is null, so in the Measured column an unresolved baseline still looks like a round legitimately showing an absolute value.

#535 addressed the resolution half of the ambiguity by separating `UNMEASURED` from `INCONCLUSIVE`. What remains of the projection gap is tracked separately.

### DIFF
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -128,26 +128,37 @@ def metric_baseline(
     metric: str | None,
     rounds: Sequence[RoundRecord],
 ) -> RoundRecord | None:
-    """Find the exact causal baseline for one metric when provenance permits."""
+    """Find the baseline for one metric, preferring the exact causal parent.
+
+    The parent of an official round is usually a provisional round whenever
+    ``official_eval_every`` exceeds 1, so an exact parent match rarely exists.
+    Fall back to the newest trusted official measurement that does not
+    postdate the parent round; comparing against a later measurement would
+    invert cause and effect.
+    """
     comparable = [
         item
         for item in rounds
         if item.official_evaluation
         and item.perf_metric is not None
         and trusted_perf_provenance(item.perf_provenance)
-        and item.perf_unit == metric
+        and (item.perf_unit == metric or (metric is not None and metric in item.metrics))
     ]
+    if not comparable:
+        return None
     if parent_commit is not None:
-        return next(
+        exact = next(
             (item for item in reversed(comparable) if item.commit == parent_commit),
             None,
         )
+        if exact is not None:
+            return exact
     if parent_round is not None:
         return next(
-            (item for item in reversed(comparable) if item.round_number == parent_round),
+            (item for item in reversed(comparable) if item.round_number <= parent_round),
             None,
         )
-    return comparable[-1] if comparable else None
+    return comparable[-1]
 
 
 def start_hypothesis(
@@ -525,7 +536,7 @@ def _measurement(
     baseline = _baseline(record, prior_rounds)
     baseline_value = record.perf_baseline_metric
     if baseline_value is None and baseline is not None:
-        baseline_value = baseline.perf_metric
+        baseline_value = _baseline_metric_value(baseline, record.perf_unit)
     delta = record.perf_delta_pct
     if delta is None and baseline_value not in {None, 0}:
         assert baseline_value is not None  # noqa: S101  # narrowed above
@@ -556,6 +567,13 @@ def _baseline(record: RoundRecord, prior_rounds: Sequence[RoundRecord]) -> Round
         metric=record.perf_unit,
         rounds=prior_rounds,
     )
+
+
+def _baseline_metric_value(baseline: RoundRecord, metric: str | None) -> float | None:
+    """Read a baseline's value for *metric*, tolerating a renamed headline unit."""
+    if metric is not None and metric in baseline.metrics:
+        return baseline.metrics[metric]
+    return baseline.perf_metric
 
 
 def _retained(
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -121,6 +121,23 @@ def scalar_candidate_retained(comparison: MetricComparison) -> bool | None:
             return None
 
 
+def record_metric_value(record: RoundRecord, metric: str | None) -> float | None:
+    """Read one official metric off *record* without guessing across axes.
+
+    The objective row wins when it names *metric*, which is what lets a record
+    whose headline unit was later renamed still answer for the axis it
+    measured. The headline scalar answers only when it is the axis asked for,
+    or when the record carries no row to disagree with.
+    """
+    if metric is not None and metric in record.metrics:
+        return record.metrics[metric]
+    if record.perf_metric is not None and (
+        metric is None or record.perf_unit == metric or not record.metrics
+    ):
+        return record.perf_metric
+    return None
+
+
 def metric_baseline(
     *,
     parent_round: int | None,
@@ -158,6 +175,13 @@ def metric_baseline(
             (item for item in reversed(comparable) if item.round_number <= parent_round),
             None,
         )
+    if parent_commit is not None:
+        # The caller named a causal parent, no trusted official round carries
+        # that commit, and there is no round number to bound the fallback by.
+        # Falling through to the newest measurement here would let a round that
+        # came *after* the parent serve as its baseline, inverting cause and
+        # effect. Fail closed instead.
+        return None
     return comparable[-1]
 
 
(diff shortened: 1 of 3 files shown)


vic-lsh:

vic-lsh:
Rebased and fixed this one too. The change is @AyanBinRafaih's work; below is the rebase and the review findings I addressed on top of it.

## Rebase

Rebased onto #535's head, then onto `main` after #535 merged as `d8060ed`. The conflict was in `metric_baseline`'s comparable filter, where this PR's renamed-metric clause and #535's provenance predicate landed on the same lines; both apply. #529 and #539 were TUI-only and did not conflict.

## Findings addressed

**[P2] `baseline_commit` named the wrong commit exactly when the fallback fired.** `_measurement` reported `record.perf_baseline_commit or record.hypothesis_parent_commit`. The parent commit is right only when the exact parent match wins. Whenever the fallback supplies the baseline, which is the entire reason this PR exists, it named the provisional parent's commit while the value came from an earlier official round. It is now the baseline record's own commit, symmetric with how `baseline_round` was already derived. `test_provisional_parent_baseline_now_resolves_via_fallback` asserts both that it equals the official round's commit and that it differs from the provisional parent's; verified to report the provisional commit pre-fix.

**[P3] The fallback did not fail closed when it had nothing to bound itself with.** With `parent_commit` set to something no trusted official round carries and `parent_round` absent, the function fell through to `comparable[-1]`, the newest measurement, which can be a round recorded *after* the parent. That is precisely the cause-and-effect inversion the `round_number <= parent_round` bound exists to prevent, just without a bound to enforce it. It now returns `None`, so the hypothesis resolves `UNMEASURED` rather than against an arbitrary comparison. Covered by `test_an_unfindable_parent_commit_with_no_round_bound_fails_closed`, contrasted against the genuinely unlinked case, which still legitimately takes the newest measurement.

**[P3] Retention and baseline selection disagreed about admissible history.** This PR widened `metric_baseline` to accept a record carrying the axis in its `metrics` row, but `_retained` still matched on `perf_unit` alone. With a renamed headline unit, resolution compared against the renamed round while retention saw an empty history, and an empty history is always retained. Both now use the same rule. `test_retention_sees_the_same_history_the_baseline_does` drives this through the derived-retention path; pre-fix it retained a strictly worse reading.

**[P3] Two spellings of the same lookup.** `_baseline_metric_value` here and `_metric_value` in the agent loop both read "the objective row if it names this axis, else the headline scalar". They are now one exported `record_metric_value` in `hypotheses.py`; the loop imports it and its private copy is gone. `hypotheses.py` is the right owner because the loop already depends on it and it already owns `RoundRecord` interpretation, whereas `metrics.py` deliberately knows nothing about loop-state records.

## Body and title

Title lost the `TUI:` prefix. The body now leads with the observed symptom from the issue (no delta at any cadence above 1) and describes the final diff.

`Closes #478` became `Part of #478`. The issue asks for two things and this PR delivers one. The other, "an unresolved baseline must be distinguishable from a measured non-improvement", is still open in the operator-facing projection: `baseline_round`, `baseline_commit`, and `baseline_value` never reach the wire, and `formatMeasured` still falls back to the raw absolute metric when `delta_pct` is null, which reads identically to a deliberate absolute display. #535 addressed the resolution half of that ambiguity by separating `UNMEASURED` from `INCONCLUSIVE`; the projection needs its own change. Happy to file that as a follow-up issue if you want it tracked separately.

## Checks

- `uv run pytest tests/vibesys/loops -q --no-cov`: 621 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 299 passed
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 15 lines, 0 missing, 100%

vic-lsh:
Correction to my earlier comment on this PR, and to the body I wrote for it.

I claimed that "`baseline_round`, `baseline_commit`, and `baseline_value` never reach the wire". That was taken from #478's own grounding, which cites commit `3a2fd8b`, and I did not re-verify it against current `main` before writing it. Two thirds of it is now wrong:

- `perf_baseline_value` **is** projected. `src/server/api/experiments.py::_entry` sets `perf_baseline_value=measurement.baseline_value`, alongside `perf_metric_name` and `perf_direction`.
- The TUI **does** surface it. `measurementMetadata` in `clients/tui/src/ui/experiment-log.ts` renders `Baseline <value> <unit>` in the hypothesis drill-down.

That work landed with #465, which closed after #478 was filed. Only `baseline_round` and `baseline_commit` remain unprojected, and the `formatMeasured` observation still holds: the Measured column falls back to a bare absolute number when `delta_pct` is null, so in that column a round whose baseline could not be resolved still looks like a round legitimately showing an absolute value.

None of this changes the diff that merged here, which is about baseline *lookup*, not projection. It changes the size of the remaining gap, and the follow-up issue I file will be scoped to what is actually left rather than to the stale description. Apologies for the sloppy citation.

PR #537, opened 2026-09-01
Reprompt once when a plan reuses a hypothesis_id

### LINKED ISSUE #476: TUI: `hypothesis_id` schema fails runs for re-used IDs, asserting invalid input

AyanBinRafaih:
> Stacked PR 6/6, the top of the #495 split. Base branch: `main` (#532 through #536 have merged).

## Problem

An orchestrator plan that reused a `hypothesis_id` raised out of `_validate_orchestrator_plan_state` and terminated the whole run. That is a recoverable agent mistake being treated as a framework invariant violation: the run had produced valid work for every prior round, and one malformed plan threw it away.

The schema made the mistake likely. `hypothesis_id` was documented as "Reuse it across rounds while the same implementer investigation remains active", but the plan role runs only when no hypothesis is active (`loop.py`: "The designer runs only when selecting a new causal claim"), and validation rejects any identifier already in state. The described case cannot reach this code, and following the description as written produces exactly the collision that killed the run.

## Solution

Reprompt once, and tell the orchestrator the rule the framework actually enforces.

- A plan failing lifecycle validation gets one corrective reprompt. A second failure raises as before, so an agent that ignores the correction still fails closed rather than looping.
- The `hypothesis_id` description states the rule unconditionally: never reuse an identifier used earlier in this run. The corrective feedback says the same thing and additionally names the rejected plan's own `hypothesis_id` and `hypothesis_updates` ids, so the orchestrator does not have to infer which identifier collided.
- The retry is labelled `round-N-retry-K-plan`. The first attempt keeps `round-N-plan`.
- `planningStage` in `clients/tui/src/session-model.ts` accepts the retry shape, so a reprompted round still shows as planning instead of the operator watching planning silently stop.

### Why the label shape matters

Two consumers parse round labels, and the shape has to satisfy both:

| Consumer | Pattern | `round-2-plan-retry` | `round-2-retry-1-plan` |
| --- | --- | --- | --- |
| `planningStage` (`clients/tui/src/session-model.ts`) | `^round-\d+(?:-retry-\d+)?-plan$` | no match: round drops out of the planning activity | matches |
| `_attempt_from_label` (`src/vibesys/context.py`) | `retry-(\d+)` | no match: attempt reported as `None` | attempt 1 |

The client regex is widened in this PR; without that change the second column's first row would still apply.

### Correctness properties

- A rejected plan costs one reprompt, never the run. A second rejection raises, so the loop cannot re-ask indefinitely.
- Every identifier the plan role proposes must be new. The framework enforces it, the schema says it, and the corrective feedback repeats it.
- A rejected attempt writes no plan artifact and no progress note, and leaves durable hypothesis state untouched: `_validate_orchestrator_plan_state` applies strategy updates to a clone it discards.
- It is **not** a full rollback, and the code now says so. The orchestrator's one allowlisted write, the roadmap index, is preserved by `_invoke_read_only_role` rather than reverted, so roadmap edits from a rejected attempt survive into the retry. That is deliberate: the roadmap is the orchestrator's own long-lived planning document, and the thinking recorded there is not invalidated by the plan JSON being rejected for an identifier collision.
- Both the retry label's consumers parse it.

### Testing

- `uv run pytest tests/vibesys/loops -q --no-cov`: 624 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 312 passed
- `cd clients/tui && bun test --max-concurrency 1 src`: 549 passed
- `pnpm check:clients`: clean
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 21 lines, 0 missing, 100%

Coverage:

- `test_reused_hypothesis_id_is_reprompted_once_and_recovers`: the recovery path, asserting the corrected plan is the one written.
- `test_reused_hypothesis_id_after_failed_reprompt_is_not_written`: asserts exactly three plan calls, so a second failure fails closed rather than reprompting again.
- `test_reprompted_plan_is_labelled_as_a_retry_of_the_same_round`: pins the exact label sequence and that the corrective prompt names the rejected identifier.
- `still presents a reprompted plan as planning` in `session-model.test.ts`, plus a negative case for an unrelated retry label. Verified to fail against the old regex.

### Prompt budget

No budget change. The earlier revision of this PR raised the orchestrator's fallback budget from 13,200 to 13,300 bytes to fit a longer description; the reworded contract is shorter than that and fits inside the original budget, which is restored. `tests/vibesys/loops/agent/test_prompt_snapshots.py` is unchanged by this PR.

### Prompt wording needs a maintainer decision

`src/vibesys/schemas.py` is not under CODEOWNERS, but its `Field(description=...)` strings are agent-visible product behavior: they are rendered into the fallback schema hint that every orchestrator turn sees. Please review the `hypothesis_id` wording specifically, both in the schema and in the corrective feedback in `_run_orchestrator_plan`, and say if you want it phrased differently. The behavioral claim I am making is that the unconditional rule is the only correct one because the plan role never runs while a hypothesis is active; if that assumption is wrong, the wording should change with it.

Closes #476.

### DIFF
--- a/clients/tui/src/session-model.test.ts
+++ b/clients/tui/src/session-model.test.ts
@@ -146,6 +146,20 @@ describe('hypothesis planning activity', () => {
     });
   });
 
+  it('still presents a reprompted plan as planning', () => {
+    // The framework reprompts the orchestrator once when a plan reuses a
+    // hypothesis_id, and labels that attempt `round-N-retry-K-plan`. It is the
+    // same stage of the same round, so the operator must not see planning
+    // silently stop just because the first attempt was rejected.
+    expect(
+      hypothesisPlanningActivity(stateFor('orchestrator', 'round-3-retry-1-plan')),
+    ).toMatchObject({stage: 'plan', roundNumber: 3});
+  });
+
+  it('does not treat an unrelated retry label as planning', () => {
+    expect(hypothesisPlanningActivity(stateFor('orchestrator', 'round-3-retry-1-judge'))).toBeNull();
+  });
+
   it('does not present a phase as planning after its round has a hypothesis', () => {
     const entries = [
       {
--- a/clients/tui/src/session-model.test.ts
+++ b/clients/tui/src/session-model.test.ts
@@ -157,7 +157,9 @@ describe('hypothesis planning activity', () => {
   });
 
   it('does not treat an unrelated retry label as planning', () => {
-    expect(hypothesisPlanningActivity(stateFor('orchestrator', 'round-3-retry-1-judge'))).toBeNull();
+    expect(
+      hypothesisPlanningActivity(stateFor('orchestrator', 'round-3-retry-1-judge')),
+    ).toBeNull();
   });
 
   it('does not present a phase as planning after its round has a hypothesis', () => {
--- a/clients/tui/src/session-model.ts
+++ b/clients/tui/src/session-model.ts
@@ -1161,7 +1161,13 @@ function planningStage(
   if (roundLabel === null) return null;
   if (agentKind === 'orchestrator' && /^round-\d+-pre$/.test(roundLabel)) return 'pre';
   if (agentKind === 'profiler' && /^round-\d+-profiler$/.test(roundLabel)) return 'profile';
-  if (agentKind === 'orchestrator' && /^round-\d+-plan$/.test(roundLabel)) return 'plan';
+  // A plan the framework reprompted carries `round-N-retry-K-plan`. It is the
+  // same planning stage, produced by the same role for the same round, so it
+  // must not fall out of the planning activity just because the first attempt
+  // was rejected.
+  if (agentKind === 'orchestrator' && /^round-\d+(?:-retry-\d+)?-plan$/.test(roundLabel)) {
+    return 'plan';
+  }
   return null;
 }
 
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -1013,16 +1013,32 @@ def _run_orchestrator_plan(  # noqa: PLR0913  # tracked: #288
         official_eval_cadence_due=official_eval_cadence_due,
     )
     # One corrective reprompt: a plan that fails lifecycle validation (for
-    # example a hypothesis_id reused after that investigation resolved) is a
-    # recoverable agent mistake, not a framework invariant violation. No
-    # artifact is written for a rejected attempt, so state stays untouched.
+    # example a hypothesis_id already used in this run) is a recoverable agent
+    # mistake, not a framework invariant violation.
+    #
+    # A rejected attempt writes no plan artifact and no progress note -- both
+    # happen after validation -- and leaves durable hypothesis state untouched,
+    # because `_validate_orchestrator_plan_state` applies strategy updates to a
+    # clone it discards. It is not a full rollback, though: the orchestrator's
+    # one allowlisted write, the roadmap index, is preserved by
+    # `_invoke_read_only_role` rather than reverted, so roadmap edits the
+    # rejected attempt made survive into the retry. That is deliberate. The
+    # roadmap is the orchestrator's own long-lived planning document, and the
+    # thinking it recorded there is not invalidated by the plan JSON being
+    # rejected for an identifier collision.
     corrective_feedback: str | None = None
+    attempt = 0
     while True:
-        retry_suffix = "" if corrective_feedback is None else "-retry"
+        attempt += 1
+        # `round-N-plan`, then `round-N-retry-1-plan`. Both the client's
+        # planning-stage matcher and `_attempt_from_label` parse this shape, so
+        # a reprompted plan still appears as a planning activity and still
+        # reports which attempt produced it.
+        label = f"round-{round_number}" + (f"-retry-{attempt - 1}" if attempt > 1 else "") + "-plan"
         plan = _invoke_read_only_role(
             ctx,
             role="orchestrator",
-            checkpoint_label=f"round-{round_number}-plan-input{retry_suffix}",
+            checkpoint_label=f"{label}-input",
             allowed_workspace_paths=(
                 f"{roadmap_location.rstrip('/')}/index.md"
                 if roadmap_location.endswith("/")
@@ -1039,7 +1055,7 @@ def _run_orchestrator_plan(  # noqa: PLR0913  # tracked: #288
                 pass_criteria="/health returns 200.",  # noqa: S106  # tracked: #288
                 reasoning="fallback: orchestrator produced no structured response",
             ),
-            round_label=f"round-{round_number}-plan{retry_suffix}",
+            round_label=label,
             reuse_session=False,
         )
         plan.hypothesis_id = plan.hypothesis_id.strip() or f"hypothesis-{round_number:04d}"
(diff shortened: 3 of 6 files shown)


vic-lsh:

vic-lsh:
Last one. Rebased and fixed; the change is @AyanBinRafaih's work, and below is what I addressed on top of it.

## Rebase

Rebased onto each preceding PR in turn, then onto `main` after #536 merged as `9e50219f`. No conflicts.

## Findings addressed

**[P1] The retry label parsed under neither consumer.** The PR labelled the reprompted attempt `round-N-plan-retry`. `planningStage` in `clients/tui/src/session-model.ts` matches `^round-\d+-plan$`, so the retry dropped out of the planning activity and an operator watching a reprompted round would see planning simply stop. `_attempt_from_label` in `src/vibesys/context.py` reads `retry-(\d+)`, which finds no number in `-retry`, so the execution reported no attempt. The label is now `round-N-retry-K-plan`, which `_attempt_from_label` already parses, and the client regex is widened to `^round-\d+(?:-retry-\d+)?-plan$`. `test_reprompted_plan_is_labelled_as_a_retry_of_the_same_round` pins the exact label sequence; the `session-model.test.ts` case was verified to fail against the old regex.

**[P1] The schema description described a case that cannot occur, and invited the collision.** It said to repeat an identifier "while the same implementer investigation is still active". But the plan role runs only when `active_hypothesis is None`, and validation rejects any id already in state, so there is no reachable situation in which repeating an id is correct. An orchestrator following the description as written produces exactly the failure this PR exists to survive. Both the schema description and the corrective feedback now state the rule the framework enforces: never reuse an identifier used earlier in this run.

**[P2] The corrective feedback did not say what was rejected.** It described the rule but not the plan, leaving the orchestrator to infer which identifier collided. It now names the rejected `hypothesis_id` and the `hypothesis_updates` ids.

**[P2] The prompt budget was raised rather than kept.** The earlier revision bumped the orchestrator fallback budget from 13,200 to 13,300 bytes to fit the longer description. The reworded contract is shorter and fits inside the original budget, measured at 13,171 bytes against the 13,200 limit, so the budget is restored and `test_prompt_snapshots.py` is untouched by this PR.

**[P2] The "no state written" claim was wrong.** The comment said a rejected attempt leaves state untouched. Three of the four things it could mean are true: no plan artifact, no progress note, and no durable hypothesis change, since `_validate_orchestrator_plan_state` applies strategy updates to a clone it discards. The fourth is not: `_invoke_read_only_role` **preserves** the orchestrator's one allowlisted write, the roadmap index, rather than reverting it, so roadmap edits from a rejected attempt survive into the retry. I corrected the claim rather than adding a revert, and said why in the comment: the roadmap is the orchestrator's own long-lived planning document, and the thinking it recorded there is not invalidated by the plan JSON being rejected for an identifier collision. Say the word if you would rather revert it.

**[P3] The reuse-twice test did not assert the reprompt count.** `test_reused_hypothesis_id_after_failed_reprompt_is_not_written` checked only that the run raised. It now asserts `orch_plan == 3`: round one's plan, round two's rejected plan, and its single corrective retry. Without that, a regression that reprompted repeatedly before failing would still pass.

## Prompt wording needs your sign-off

`src/vibesys/schemas.py` is not under CODEOWNERS, but its `Field(description=...)` strings are agent-visible product behavior, rendered into the fallback schema hint on every orchestrator turn. Please review the `hypothesis_id` wording in the schema and the corrective feedback in `_run_orchestrator_plan`. The behavioral claim behind the rewrite is that the unconditional rule is the only correct one because the plan role never runs while a hypothesis is active. I verified that against `loop.py` (`if active_hypothesis is None:` guards the call, and the comment above it says the designer runs only when selecting a new causal claim), but it is your prompt surface, so it is your call.

## Checks

- `uv run pytest tests/vibesys/loops -q --no-cov`: 624 passed
- `uv run pytest libs/vs-loop-state/tests tests/server -q --no-cov`: 312 passed
- `cd clients/tui && bun test --max-concurrency 1 src`: 549 passed; `pnpm check:clients` clean
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=80`: 21 lines, 0 missing, 100%

vic-lsh:
Pushed a follow-up commit: `pnpm check:ts` (Biome) failed on the new `session-model.test.ts` case for line width. That is a separate gate from `./scripts/check_format.sh` and `check_lint.sh`, which are Python-only, and I had run only the latter. Fixed with `pnpm format:ts`; `pnpm check:ts` and `pnpm check:ts-architecture` are now clean, and the 81 session-model tests still pass.

**Unrelated failure worth your attention.** The four "Build and publish release wheels" jobs fail on this PR, but they also failed on #535, which merged anyway, and the cause is on `main` rather than in this stack:

```
File ".../vibesys/loops/agent/loop.py", line 3278, in run_agent_loop
    implementer_driver=ctx.agent_client.driver_name,
AttributeError: 'StubAgentClient' object has no attribute 'driver_name'
```

#523 added `implementer_driver` / `implementer_provider` / `implementer_model` to the `RoundRecord` construction, reading them off `ctx.agent_client`, but `StubAgentClient` in `src/vibesys/agents/stub_runner.py` never gained them. I confirmed it against `origin/main` at `9e50219f`: `stub_runner.py` has zero occurrences of `driver_name`, and `loop.py:3278` reads it unconditionally.

The unit suite misses it because the orchestrate harness uses a `MagicMock` agent client, which auto-creates any attribute. Only the installed-wheel smoke test (`vibesys --headless --stub-agent ...`) drives the real `StubAgentClient`, so `main` currently cannot pass its own release verification. Not something I should fix inside this PR's scope, but it wants its own change.

PR #539, opened 2026-09-01
TUI: vertical rounds rail with left-to-right drill-down

### LINKED ISSUE #518: TUI: vertical rounds strip with left-to-right drill-down hierarchy
### Workstream

CLI/DX

### Problem

Rounds are presented as a one-line horizontal chip strip (`[ r1 16m 20s ]  r2` in `RoundStripView`, `clients/tui/src/ui/round-strip.ts`, fixed `height: 3`, `flexDirection: 'row'`) pinned above the round view. The horizontal form wastes the dimension terminals have the least of (width is already contested by the Agents pane, transcript, and right pane), scales badly (long runs collapse into `‹ 12` / `4 ›` overflow markers, and the marker counts go stale because `#repaint` at round-strip.ts:170-199 only diffs chips, not markers), and carries almost no information per round (only `rN` and a ticking elapsed for the active round; no outcome, no hypothesis, no measured delta). Navigation is also inconsistent with the drawn axes: `[`/`]` move rounds horizontally, arrows move panes and agents, and the hierarchy run > hypothesis > round > agents/transcript has no single visual direction.

### Desired outcome

The rounds presentation becomes a vertical rail: rounds listed top-to-bottom on the left edge of the round view, one row per round showing at minimum the round number, status glyph and word, elapsed or duration, and outcome or measured delta when resolved, with the selected round highlighted and Up/Down moving between rounds. The overall hierarchy then reads left to right: rounds rail, then the agents graph and transcript for the selected round to its right, so drilling deeper always moves rightward. The vertical rail scrolls naturally for long runs instead of collapsing into overflow markers. At narrow widths the rail may collapse to the current compact strip as a fallback.

### Acceptance criteria

- [ ] Rounds render as a vertical list in the round view at wide widths, with number, status word plus glyph, elapsed/duration, and outcome or measured value per row.
- [ ] Up/Down (or an equivalent documented binding) moves the round selection; the footer hint is updated accordingly.
- [ ] Long runs (20 or more rounds) scroll within the rail with no stale overflow counts.
- [ ] The agents graph and transcript remain to the right of the rail and update on selection change, preserving existing `[`/`]` behavior as an alias or with a documented replacement.
- [ ] Narrow-width fallback is defined and tested at the existing responsive breakpoints.
- [ ] Layout and interaction covered by tests alongside the existing round-strip tests.

### Scope and non-goals

In scope: 
`round-strip.ts` (or a successor component), round-view layout composition in `ui/app.ts`, keybindings, and footer hints.

Not in scope: 
the agent graph internals (issue 1 covers its correctness; its left-to-right stage layout already matches the desired hierarchy), the experiments log table, and any protocol change.

The Agents pane and transcript already compete for width (`STACKED_WIDTH`, `TRANSCRIPT_MIN`, `GRAPH_SHARE` in agent-map.ts:33-37); the rail must fit the existing width budget or take over the stacked-list fallback slot. Per-round outcome data is already available in the round records that feed the experiments table, so no new backend queries should be required.

### Verification approach

tmux-driven frame captures at 200x50, 120x40, and 92x40 across a real multi-round run, plus unit tests for rail rendering, selection, and scrolling with synthetic round sets of 1, 3, and 30 rounds.

#284

AyanBinRafaih:
## Problem

Fixes #518.

Rounds were presented as a one-line horizontal chip strip (`RoundStripView` in `clients/tui/src/ui/round-strip.ts`, fixed `height: 3`, `flexDirection: 'row'`) pinned above the round view. The horizontal form spent the dimension a terminal has least of: width was already contested by the agents pane, the transcript, and the right pane. It scaled badly, since a long run collapsed into `‹ 12` / `4 ›` overflow markers whose counts went stale because the strip's repaint diffed chips but not the markers. It also carried almost no information per round, only `rN` and a ticking elapsed for the active round, with no outcome, no status word, and no measured delta. Navigation contradicted the drawn axes: `[` and `]` moved rounds horizontally while the arrows moved panes and agents, so the hierarchy run to hypothesis to round to agents and transcript had no single visual direction.

## Solution

Rounds become a vertical rail down the left edge of the round view, one round per row, so the hierarchy reads left to right: the rounds rail, then the agents graph and transcript for the selected round to its right. Drilling deeper always moves rightward. `RoundStripView` is replaced by `RoundRailView` in a new `clients/tui/src/ui/round-rail.ts`; the old component and its test are deleted.

Each full row shows the round number, a status glyph and word (`⟳ run`, `✓ done`, `✗ fail`, `· plan`), and a metric: the live agent-active elapsed while a round runs, the measured `perf_delta_pct` once it resolves, or its wall duration when no delta was recorded. The delta is read from the experiment log's `HypothesisRound` for that round, joined by round number, which is the same record the experiments table reads, so the rail and the table cannot disagree about a round. The selected round is marked with `▸` and drawn on the accent surface, so it never depends on colour alone. A run taller than the rail becomes a window that follows the selection and slides by one round per step, with `↑ n` and `↓ n` indicators recomputed every render, so the overflow counts can never go stale. Rounds are placed before indicators and never exceed the rows on hand, so a one- or two-row rail emits nothing it cannot show.

Focus becomes a left-to-right order. The round view gains a `rounds` focus target (a new `RoundFocus` value), and `←` and `→` step focus across the rail, the agents graph, and the transcript. `↑` and `↓` act within the focused surface, so on the rail they move the round selection. `[` and `]` still step rounds from any focus. The rail is a selector rather than a zoomable pane, so `focusedPane` reports it as the agents side and `F4` zoom is unaffected. Because a resize can hide the rail while `roundFocus` still reads `rounds`, both the arrow keys and the agents pane's focus border treat that stale focus as the agents pane, so the keys always drive a visible surface and a pane always says it holds them.

Width follows the terminal, and the breakpoints are derived from the floors rather than pinned: the rail may appear only where its column still leaves the agents pane and the transcript their minimums. At or above 100 columns (`RAIL_FULL_WIDTH + STACKED_WIDTH + TRANSCRIPT_MIN`) it shows the full rows; from 85 (`RAIL_COMPACT_WIDTH + …`) it falls back to a compact number-and-glyph column; narrower than that it collapses. It also yields while the experiment log is the landing view, while a pane is zoomed, while a right-pane split takes the row, and before the run has any rounds.

### Architecture

This is a TUI-layer change only. `round-rail.ts` owns the rail's width function (`roundRailWidth`), its visibility predicate (`roundRailVisible`), the windowing of rounds onto the available rows (`railWindow`), and the `RoundRailView` widget. `app.ts` composes the round view, mounting the rail as the first child of the content row and handing it its width and row budget each frame, and passes the rail's column width to the agents pane, which sizes itself against what is left. The rail's row budget comes from `todoStripHeight(state)`, the same function the todo strip sets its own box height from, so the two are one value rather than a box height read back a paint late. `keybindings.ts` owns the `←`/`→` focus order and the `↑`/`↓` within-pane behavior. `session-model.ts` owns the `RoundFocus` type, the `focusedPane` coercion, and the `hypothesisRoundFor` selector. No protocol, `core-state`, or backend change is needed: the per-round facts already cross the protocol on `HypothesisRound`.

```mermaid
flowchart LR
    subgraph state[session state]
      R["stripRounds: run rounds"]
      D["hypothesisRoundFor: experiment log record"]
    end
    subgraph tui[tui round view]
      APP["app.ts render: railWidth, railRows from todoStripHeight"]
      RAIL["RoundRailView: roundRailWidth / roundRailVisible / railWindow"]
      AG["AgentMapView sized against terminal minus railWidth, transcript floor kept"]
      TR[Transcript]
      KEYS["keybindings: left/right focus order rounds to agents to transcript, up/down within"]
    end
    R --> RAIL
    D --> RAIL
    APP --> RAIL
    APP --> AG
    APP --> TR
    KEYS -->|focusRound| RAIL
    KEYS -->|selectRound| RAIL
    RAIL -->|selection| AG
    RAIL -->|selection| TR
```

## Verification

Automated. Layout behavior stated in columns or rows is verified through `createTestRenderer`, which runs the real renderable tree and Yoga layout; pure functions (windowing, breakpoints, visibility) are covered by unit tests.

### Correctness properties

- The hierarchy reads left to right: rail, then agents graph and transcript for the selected round, and drilling deeper always moves rightward.
- Every full row carries the round number, a status glyph and word, and a metric: live elapsed while active, the experiment log's `perf_delta_pct` once resolved, or wall duration when no delta was recorded; planned rounds show no metric.
- The rail's measured delta and the experiments table's come from the same `HypothesisRound`, joined by round number, so the two cannot describe a round differently.
- The selected round is marked with `▸` and the accent surface, so selection is legible without colour.
- A run taller than the rail is a window that always keeps the selection in view, slides by at most one round per step, and reports `↑ n` / `↓ n` counts recomputed every render. With too few rows for both, rounds are drawn and indicators dropped, so the rail never emits more children than rows.
- `←` and `→` step focus across rail, agents, and transcript; `↑` and `↓` act within the focused surface; `[` and `]` still step rounds from any focus.
- The rail joins the focus order only when it is on screen. When it is hidden, a `roundFocus` of `rounds` drives the agents pane and the agents pane carries the focus border, so the arrows never move an invisible selection and the view always says where its keys are.
- The rail appears only at widths where its column still leaves the agents pane and the transcript their floors; the transcript is never squeezed below `TRANSCRIPT_MIN` beside it.
- The rail's row budget and the todo strip's height are the same value, so expanding or collapsing todos sizes the rail correctly in the same paint.

### Testing

- `clients/tui/src/ui/round-rail.test.ts` (new): `railWindow` (all-fit, selection always visible, per-side hidden counts, two rows reserved for indicators on overflow, never exceeds the rows given, fills at either end, slides by one per step, stable order, empty and one-round runs, zero rows, very short rail, and the one- and two-row cases); `roundRailWidth` (the 100 and 85 breakpoints and the collapse below them); `roundRailVisible` (live run true; rounds-empty, log-visible, zoomed, split, and below-collapse-width false); and `RoundRailView` row-budget cases through `createTestRenderer` for a zero-row box and for dropping indicators before rounds.
- `clients/tui/src/ui/agent-map.test.ts` (new): the transcript floor holds for every terminal width from 72 to 140 across round shapes, at the exact widths the review measured (72, 80, 84), and the compact rail appears only where both floors still fit beside it.
- `clients/tui/src/ui/app.test.ts` (updated): the round-view render tests assert the vertical rows (`▸r2 ⟳ run 1m …`, `r1 ✓ done`, `r3 ✗ fail`, `▸r9 · plan`) and the `↓ n` overflow indicator instead of the old horizontal chips; the no-rounds cases assert the rail is absent; new cases cover the transcript floor at 80 and 100 columns measured off the laid-out box, the rail's delta coming from the experiment log, a stale `rounds` focus driving the agents pane and giving it the focus border at 70 columns, the rail taking `↑`/`↓` while it is on screen, and the rail resizing from the todo strip's current height across an expand and a collapse.
- `clients/tui/README.md` (updated): the rounds presentation is documented as the vertical rail (rows, metric, windowing, and the full/compact/hidden widths and why those thresholds), and the round-view navigation is restated as `←`/`→` across the rail, agents, and transcript with `↑`/`↓` within and `[`/`]` from anywhere.

Commands run at the head of this branch:

- `cd clients/tui && bun test --max-concurrency 1 src`: 542 pass, 0 fail.
- `pnpm check:clients`: pass.
- `pnpm check:ts`: pass.
- `pnpm check:ts-architecture`: pass (no dependency violations, 81 modules).
- `pnpm --dir clients/backend-client generate:protocol`: the working tree is clean afterward, so the generated protocol is unchanged by this branch.
- No Python changed, so `pytest` was not run.

Closes #518.


vic-lsh:
Review findings:

[P2] Preserve the transcript width floor while the compact rail is visible (`clients/tui/src/ui/agent-map.ts:113-116`, wiring at `ui/app.ts:358-362`).

At terminal widths 72-84, the 13-column rail leaves too little room for the 30-column agent floor plus the 42-column transcript floor. `agentPaneWidth(...)` correctly returns null, but line 116 turns null back into a 30-column pane. Direct checks produced transcript widths 29, 37, and 41 at widths 72, 80, and 84. Please hide/change the rail or collapse the agent layout until both floors fit, and test the breakpoint range.

[P2] Do not navigate an invisible rail after its focus becomes stale (`clients/tui/src/ui/keybindings.ts:258-263`).

Resize below 72, zoom, split, or other visibility changes can hide the rail without normalizing `roundFocus`. Left/Right consult visibility, but Up/Down branches only on `roundFocus === 'rounds'`, so no visible pane carries focus and the keys change round selection. Gate this branch on current rail visibility or normalize focus whenever visibility changes, with a focus-then-hide test.

[P2] Compute rail rows from the current todo-strip height (`clients/tui/src/ui/app.ts:345-370`).

`railRows` reads `todoStrip.output.height` before `todoStrip.render(...)` updates it for the new state. Expanding or collapsing todos therefore sizes the rail with the previous height until another render, which can clip the selected late round or overflow indicator. Render/measure todos first, or derive the height from current state, and test both transitions.

[P3] Keep indicators and rounds within very short row budgets (`clients/tui/src/ui/round-rail.ts:89-114`, rendering at 195-201).

With only one or two content rows and overflow, `railWindow` reserves one round but rendering can add two indicators, producing more children than available rows. With zero rows it can still report and render a hidden-after indicator. Hide the rail/indicators when space is insufficient or budget indicators dynamically, and add 0-2-row cases.

vic-lsh:

vic-lsh:
Thanks @AyanBinRafaih — the vertical rail is the right call, and making the focus order match the drawn axes (rounds → agents → transcript, with `↑`/`↓` inside whichever holds the keys) is what turns it from a rotated strip into a coherent hierarchy. I rebased onto the merged #497 and the reworked #529 and pushed a maintainer pass on top of your three commits.

**Dependency: this must merge after #529.** The round-view footer hint is generated from #529's `COMMAND_NAMES`, so this branch is built on #529's head. GitHub cannot express a cross-repository stack, so the base stays `main`; please merge #529 first. Once it lands, this branch's diff against `main` collapses to the rail files alone.

**Rebase.** Only this PR's three commits were cherry-picked onto #529's reworked head. One conflict, in `keybindings.ts`: `main` had gained an `actions.inputIsEmpty()` guard on `←`/`→`, so pane focus yields to a typed command. The resolution keeps both that guard and the new three-way focus order.

**[P2] The rail read `perf_delta_pct` off the design record.** #497 moved every per-round stage fact onto `HypothesisRound` and left `DesignRound` with `round`, `commit`, `files`, so `designRoundFor(...)?.perf_delta_pct` no longer exists. The rail now joins the experiment log by round number through a new `hypothesisRoundFor` selector, which is the same record the experiments table reads, so the two cannot report different numbers for the same round. Renderer test: an experiment log with `-3.5` and `12.4` on rounds 1 and 2 puts `-3.5%` and `+12%` on the rail rows; against the design-log source those rows fall back to a duration.

**[P2] Transcript width floor beside the compact rail.** Your `RAIL_MIN = RAIL_COMPACT_WIDTH + STACKED_WIDTH + TRANSCRIPT_MIN` (85) and `RAIL_WIDE_MIN` (100) are the right fix, and deriving both from the floors rather than pinning two magic numbers is better than what the finding asked for. The evidence was a pure-function test in `agent-map.test.ts` that recomputes app.ts's width arithmetic; the repo requires a symptom stated in columns to be reproduced through `createTestRenderer`, and a test that computes the expected width itself cannot catch a layout regression. Added `app.test.ts` cases at 80 and 100 columns that build the real round view and measure the laid-out transcript box against `TRANSCRIPT_MIN`, and assert rail + agents + transcript still fit the terminal. At the old 72-column collapse point the 80-column case reports exactly the 37 columns the review measured. Your unit test is kept: it covers the whole 72-140 range across round shapes, which the renderer test cannot.

**[P2] Focus staleness, and [P2] rail rows from the todo strip.** Both fixes are correct and covered. Two follow-ons:

- `todoStripHeight` restated `render`'s height arithmetic, so the derived height and the box height were two copies that could drift. `render` now sets `output.height` from `todoStripHeight`, making it one value.
- With the rail hidden and `roundFocus` still `rounds`, the arrows correctly drive the agents pane but no pane drew a focus border, so the view did not say where its keys went. `AgentMapView.render` now takes the rail's column width (0 when off screen), which says both how much room is left and whether the rail is a surface the keys can be on, and carries the border in that case. Asserted in your stale-focus test.

**[P3] Short row budgets.** Placing rounds first and drawing an indicator only while a row is still free is the right shape; nothing to add.

**Protocol.** These three commits touch only `clients/tui`. `pnpm --dir clients/backend-client generate:protocol` leaves `git status` clean, so the generated schema on this branch is what `main` already has.

Checks on the pushed head: `bun test --max-concurrency 1 src` in `clients/tui` (542 pass, 0 fail), `pnpm check:clients`, `pnpm check:ts`, `pnpm check:ts-architecture`. No Python changed.

Not merged and not marked ready.

vic-lsh:
#529 is merged as 99af45c4, so I rebased this branch onto `main`. It now carries only its own four commits, and the `COMMAND_NAMES` footer hint it needs is in `main` already. Clean replay, no conflicts: the squash of #529 is content-identical to the base this branch was on.

Dropped the merge-order note from the body; there is nothing left to sequence.

Re-run at the new head: `bun test --max-concurrency 1 src` in `clients/tui` (542 pass, 0 fail), `pnpm check:clients`, `pnpm check:ts`, `pnpm check:ts-architecture` all pass.

Still not merged and not marked ready.

PR #540, opened 2026-09-02
fix(core-state): reset terminal flag on run_started for resumed runs

### LINKED ISSUE #505: TUI: Resumed runs leave the TUI terminal flag stuck true, suppressing disconnect reporting and live activity
### Observed behavior

When a run is resumed, the durable log still contains the prior process's `run_failed`/`run_interrupted` event followed by the new `run_started`. Both types are in `_BOOTSTRAP_SPINE_TYPES` (`src/vibesys/server/supervisor.py:81-92`), so the TUI's subscription replay always folds the old terminal event: it calls `terminate()` (`clients/core-state/src/core-state.ts:549-551`), setting `terminal: true`. The subsequent `run_started` handler only sets `status: 'running'` and `maxRounds` (`core-state.ts:525-528`) and never resets `terminal`. The boot snapshot could repair it, but it is requested exactly once, concurrently, and the replay batch is the longest boot round trip (`session-controller.ts:239-247`), so it folds last and re-poisons the flag; the rebootstrap path re-poisons deterministically.

Confirmed by executing the built reducer: folding `[run_failed seq 1, run_started seq 2]` through `reduceEventBatch` yields `status='running'` with `terminal=true`. The client even documents the replay pattern (`session-model.ts:1341-1348`) and tests it (`session-model.test.ts:54-79`) without asserting on `terminal`.

Three user-visible consequences of the stuck flag:

1. `session-controller.ts:284`: when the backend socket later drops mid-run, the disconnect banner and `markEventStreamUnavailable` are both skipped, so stale active executions keep rendering as live (`session-model.ts:1308-1316`) and the TUI silently freezes on stale state.
2. `session-model.ts:1062`: the hypothesis planning-activity indicator never appears for the whole resumed run.
3. `conversation.ts:423`: live output renders as non-streaming.

### Expected behavior

`run_started`, a strictly newer event than the replayed failure, resets `terminal: false`, restoring `terminate()`'s invariant symmetrically. Setting it in the `run_started` branch suffices; active executions are already reconciled per batch via `reconcileActiveExecutions` (`core-state.ts:257-259`).

### Reproduction

1. Start a run and stop the backend with SIGTERM or Ctrl-C (not SIGKILL; `runtime.py:70-88` must get to write `RUN_INTERRUPTED`).
2. Resume the run with `--resume` and attach the TUI.
3. Observe `state.core.terminal === true` while `status === 'running'`: no planning indicator all run; kill the backend socket again and no error banner appears while active-execution spinners keep running.

Unit-level: add `expect(state.core.terminal).toBe(false)` to the existing test at `clients/tui/src/session-model.test.ts:54-79`; it fails today.

### Affected subsystem

CLI and packaging

### Environment

Commit: 32ca79d (main)
OS: any
Hardware: any
Backend: any resumed run (sequence continuity across resume: server/events.py:463)
Profiler: any

### Impact

Blocks normal use

### Relevant logs

```shell
reduceEventBatch([run_failed seq 1, run_started seq 2]) => { status: 'running', terminal: true }
```

#284

### Pre-submission checks

zatchbell1311-wq:
Fixes #505

## Background
VibeSys persists run events to a durable log. When a run is resumed 
with `--resume`, the TUI replays the full event history from that log 
to reconstruct its state. This replay always includes the terminal 
event from the previous session (`run_failed` or `run_interrupted`) 
followed by the new `run_started`.

## The Bug
The `run_started` handler in `core-state.ts` only sets 
`status: 'running'` and `maxRounds` — it never resets `terminal`. 
So after replay, the session ends up with `status: 'running'` and 
`terminal: true` simultaneously, which is a self-contradictory state 
the rest of the TUI was never designed to handle.

## User-Visible Impact
Three things break silently for the entire resumed session:
- **No disconnect banner** — if the backend socket drops, the error 
  banner and `markEventStreamUnavailable` are both skipped
- **No planning indicator** — the hypothesis planning-activity 
  spinner never appears
- **Non-streaming output** — live agent output renders as static 
  instead of streaming

## Fix
Add `next.terminal = false` to the `run_started` handler in 
`clients/core-state/src/core-state.ts`. This mirrors the symmetry 
that `terminate()` sets `terminal: true` — `run_started` should 
always undo it.

## Tests
Added `expect(state.core.terminal).toBe(false)` to the existing 
resumed-run reducer test in `clients/tui/src/session-model.test.ts`. 
The assertion failed before this fix and passes now. 
All 77 tests pass, 0 fail.

### DIFF
--- a/clients/core-state/src/core-state.ts
+++ b/clients/core-state/src/core-state.ts
@@ -524,6 +524,7 @@ function foldEvent(state: CoreState, event: RunEvent, folder: TranscriptFolder |
 
   if (event.type === 'run_started') {
     next.status = 'running';
+    next.terminal = false;
     if (data?.kind === 'run_started') next.maxRounds = data.max_rounds;
   }
   if (event.type === 'configuration_failed') return terminate(next, 'failed');
--- a/clients/tui/src/session-model.test.ts
+++ b/clients/tui/src/session-model.test.ts
@@ -74,6 +74,7 @@ describe('event batch projection', () => {
     ]);
 
     expect(state.core.status).toBe('running');
+    expect(state.core.terminal).toBe(false);
     expect(state.core.diagnostics).toHaveLength(1);
     expect(state.errorBanner).toEqual(before.errorBanner);
   });


vic-lsh:
FYI @zatchbell1311-wq: #576 follows up on this fix. It removes the separate `terminal` flag from `CoreState` and derives whether a run has ended from a typed run status enum, so the inconsistent `status: running` plus `terminal: true` state that this PR fixed can no longer be represented. No action needed on your side; thanks for the fix and the clear write-up.

PR #541, opened 2026-09-02
fix(hypotheses): thread noise_fraction into evidence projection

### LINKED ISSUE #507: TUI: Evidence projection drops the configured noise tolerance, flipping within-noise resolutions to proven or disproven
### Observed behavior

Two call sites resolve the same round with different evidence. The loop resolves with the task's configured noise floor: `resolve_hypothesis_outcome(ResolutionEvidence(..., noise_fraction=pareto_relative_noise, ...))` (`src/vibesys/loops/agent/loop.py:3344-3355`, also passed to `scalar_candidate_retained` at `loop.py:3375-3380`), so a within-noise official delta is recorded in the round as `hypothesis_outcome="inconclusive"`. But `project_round_evidence` (`hypotheses.py:286-299`), which owns the authoritative persisted `Hypothesis.resolution`, rebuilds `ResolutionEvidence` without `noise_fraction`; the dataclass default is 0.0 (`hypotheses.py:38`), so `_resolve_metric_evidence` uses `tolerance_pct=0` (`hypotheses.py:84`) and the identical delta resolves PROVEN (or DISPROVEN for a negative delta). The recompute runs for every new record because `loop.py:3404` always sets `judge_verdict`. `reproject_run_evidence` (`hypotheses.py:316-358`) has no noise parameter at all, and its callers (`state.py:82` on resume, `server/service.py:248`) cannot supply one, so resume rebuilds the same wrong values, and the TUI displays them (`server/experiments.py:43-45`, `resolved_outcome`).

Trigger precisely: a reviewed pass round with `official_evaluation=True`, declared supported/nominated, direction and baseline present, and 0 < |delta| < relative_noise * 100 percent. `relative_noise` is a shipped config (`main.py:2195-2218`; `examples/model-serving/Llama-3-8B/objectives.toml` sets 0.03), though the default is 0.0 (`loop.py:238`), so only opt-in configs are affected. `_retained` (`hypotheses.py:491-495`) omits `noise_fraction` too, affecting legacy records only.

Scope note (checked 2026-08-31): open PR #495 rewrites this exact projection site. It removes `benchmark_expected` from `ResolutionEvidence`, tightens `official_metric` to trusted-provenance measurements, and adds an UNMEASURED resolution, which fixes the two other input divergences this finding originally bundled. It does NOT thread `noise_fraction` into `project_round_evidence` or `reproject_run_evidence`: the PR's projection still builds `ResolutionEvidence` without it while the loop call site keeps `noise_fraction=pareto_relative_noise` unchanged. 

### Expected behavior

The persisted `Hypothesis.resolution` matches the round record's resolution: `noise_fraction` (and the other resolution inputs) are threaded into `project_round_evidence`/`reproject_run_evidence`, or the loop's resolved outcome is treated as authoritative for its own round rather than recomputed with different parameters.

### Reproduction

No run needed:

1. Construct a `RoundRecord` with `judge_verdict="pass"`, `passed=True`, `official_evaluation=True`, declared outcome supported, `perf_metric=1.02`, `perf_baseline_metric=1.0`, `perf_direction="max"` (after #495, also framework provenance so the measurement counts as trusted).
2. `append_round` yields `hypothesis.resolution=PROVEN`, while `resolve_hypothesis_outcome` on the same evidence with `noise_fraction=0.03` returns INCONCLUSIVE.

Live: `objectives.toml` with `[pareto] relative_noise = 0.05`, a round whose official benchmark improves the headline metric by 1 percent: round record says inconclusive, persisted `state.json` hypothesis says proven; `--resume` reproduces the contradiction.

### Affected subsystem

Optimization loop and agents

### Environment

Commit: 32ca79d (main)
OS: any
Hardware: any
Backend: any; task with [pareto] relative_noise > 0 in objectives.toml
Profiler: any official benchmark path

### Impact

Blocks normal use

### Relevant logs

```shell
round record:         hypothesis_outcome="inconclusive"   (loop, tolerance 3%)
persisted hypothesis: resolution="proven"                 (projection, tolerance 0)
```

#284

### Pre-submission checks

zatchbell1311-wq:
Fixes #507

## Problem
Evidence projection rebuilt ResolutionEvidence with noise_fraction=0.0
instead of the task's configured pareto_relative_noise. This caused
within-noise benchmark deltas to resolve as PROVEN or DISPROVEN in the
persisted Hypothesis, contradicting the round record's INCONCLUSIVE outcome.
On --resume, reproject_run_evidence reproduced the same wrong values,
making the contradiction permanent.

## Solution
Thread noise_fraction through the full projection call chain so the
persisted resolution always matches what the loop computed.

### Architecture
No structural changes. noise_fraction is added as a keyword parameter
(default 0.0) to project_round_evidence and reproject_run_evidence in
hypotheses.py, then forwarded through migrate_legacy and
_migrate_legacy_state in state.py. The loop call site in loop.py passes
pareto_relative_noise to migrate_legacy.

### Correctness properties
- Persisted Hypothesis.resolution matches the round record's
  hypothesis_outcome for within-noise deltas.
- reproject_run_evidence on --resume produces the same resolution
  as the original loop run.
- Default noise_fraction=0.0 preserves existing behavior for configs
  without pareto relative_noise set.

### Testing
uv run pytest tests/vibesys/loops/agent/ -x -q
314 passed, 0 failed

### DIFF
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -271,6 +271,7 @@ def project_round_evidence(
     *,
     prior_rounds: Sequence[RoundRecord],
     legacy_directions: Mapping[str, Literal["max", "min"]] | None = None,
+    noise_fraction: float = 0.0,
 ) -> Hypothesis:
     """Return a hypothesis updated with one completed round's evidence."""
     if record.hypothesis_id != hypothesis.hypothesis_id:
@@ -295,6 +296,7 @@ def project_round_evidence(
                     measurement.direction if measurement is not None else record.perf_direction
                 ),
                 benchmark_expected=record.official_evaluation,
+                noise_fraction=noise_fraction,
             )
         )
     else:
@@ -317,6 +319,7 @@ def reproject_run_evidence(
     state: AgentRunState,
     *,
     legacy_directions: Mapping[str, Literal["max", "min"]] | None = None,
+    noise_fraction: float = 0.0,
 ) -> AgentRunState:
     """Rebuild hypothesis summaries from their authoritative round evidence."""
     updated = state.clone()
@@ -353,6 +356,7 @@ def reproject_run_evidence(
             record,
             prior_rounds=prior_rounds,
             legacy_directions=legacy_directions,
+            noise_fraction=noise_fraction,
         )
         prior_rounds.append(record)
     return _validated_state(updated)
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -2498,6 +2498,7 @@ def run_agent_loop(  # noqa: C901, PLR0912, PLR0913, PLR0915  # tracked: #288
         rounds=legacy_records,
         local_namespace=local_agent_state,
         legacy_directions=legacy_metric_directions,
+        noise_fraction=pareto_relative_noise,
     )
     active_hypothesis = agent_run_state.active_hypothesis
     if active_hypothesis is not None and _backfill_revert_commit(
--- a/src/vibesys/loops/agent/state.py
+++ b/src/vibesys/loops/agent/state.py
@@ -75,6 +75,7 @@ class AgentRunStateStore:
         rounds: Sequence[RoundRecord],
         local_namespace: StateNamespace,
         legacy_directions: Mapping[str, Literal["max", "min"]] | None = None,
+        noise_fraction: float = 0.0,
     ) -> AgentRunState:
         """Build unified state from legacy files without modifying either format."""
         existing = self.load_optional()
@@ -82,6 +83,7 @@ class AgentRunStateStore:
             return reproject_run_evidence(
                 existing,
                 legacy_directions=legacy_directions,
+                noise_fraction=noise_fraction,
             )
         ledger = self._namespace.load_optional(
             self._LEGACY_LEDGER_FILE,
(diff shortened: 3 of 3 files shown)


vic-lsh:
FYI @zatchbell1311-wq: I opened #579, which fixes #507 by making the noise tolerance a persisted single-source value rather than a threaded parameter. Reproducing the bug at this PR's head, two call sites still compare with zero tolerance: `append_round` (which the loop calls after every completed round) never passes a fraction to `project_round_evidence`, and the server read path in `src/server/api/service.py` has no way to obtain one because the value is not persisted anywhere it can read.

Because each new parameter defaults to `0.0`, forgetting one stays silent, so #579 instead stores a `NoiseModel` on `AgentRunState`, has `run_agent_loop` write it once at startup, and has `append_round`, `reproject_run_evidence`, the legacy migration, retention, and the Pareto code all read it back from the state. It also collapses the three different tolerance formulas into one comparison so resolution, retention, and dominance cannot drift apart again.

Not closing this PR: your diff identified the right defect and the right call sites, and it is the maintainer's call which approach to take.

vic-lsh:
Closing in favor of #579, which merged and fixes #507 at every reader: the tolerance now lives in one persisted MetricSpace that append_round, resume reprojection, and the server read path all consult, so the two call sites this PR left on zero tolerance are covered structurally. Thanks for tracking down the original contradiction; the reproduction in your PR description is what made the wider fix obvious.

PR #543, opened 2026-09-02
feat(tui): add a dev-only replay harness driven by recorded runs

### LINKED ISSUE #542: TUI: drive rendering work from recorded runs instead of a live agent run
### Workstream

CLI/DX

### Problem

Every TUI rendering change costs a real optimization run to look at.

The TUI is a separate process that requires a live Python backend: `clients/tui/src/index.ts:13-14` refuses to start without `VIBESYS_CONTROL_SOCKET`, and the only thing that sets it today is `clients/tui/src/launcher.ts:126`, which spawns the backend alongside the client. There is no other way to get a populated frame on screen.

So checking a layout, theme, wrapping, or keybinding change means launching a real run: an agent invocation, provider tokens, and minutes of wall clock before the first interesting event arrives. States that matter for rendering are worse than merely slow, because reaching them is not under the developer's control:

- Terminal states (`run_finished`, `run_interrupted`, `run_failed`, and the error banner) require waiting out or killing a whole run.
- `tool_result` events carrying a typed `payload` require a run that actually used the tools in question, which is what the agent decides, not the developer.
- Schema variation across older and newer runs cannot be produced on demand at all.

The unit tests cover state transitions, but they render nothing, so there is no cheap way to answer "what does this actually look like at 200x50, in this theme, at this moment of a run". The cost lands on exactly the work the TUI roadmap (#284) asks for, and on reproducing the rendering bugs already filed under it (#516, #465, #464, #514).

### Desired outcome

A developer can put the real TUI on screen, populated with a realistic run, in seconds, with no Python backend, no agent, and no tokens.

Concretely: pick a recorded run, watch it render at controllable speed, stop on a frame and inspect it, and capture frames non-interactively so rendering behavior can be checked from a script rather than by eye.

### Acceptance criteria

- [ ] The unmodified shipped TUI binary can be driven against a recorded run with no Python backend, no agent process, and no provider tokens.
- [ ] Replay speed is controllable, including delivering an entire run at once, so a terminal state is reachable in seconds.
- [ ] The replay can be held on a frame and resumed, from inside the TUI.
- [ ] Frames can be captured non-interactively at a fixed terminal size, as plain text and with escape sequences, so layout and color are both scriptable.
- [ ] A developer can replay one of their own `run-events.jsonl` files, not only bundled captures.
- [ ] Bundled captures cover a full `run_started` through `run_finished` lifecycle, and a run whose `tool_result` events carry typed payloads.
- [ ] Bundled captures contain no usernames, home paths, credentials, or other local identifying content.
- [ ] Nothing added here is compiled into `dist` or included in the published `@vibesys/tui` package, and no file on the shipping path imports it.

### Scope and non-goals

In scope:

- Development-only tooling under `clients/tui`, plus recorded fixtures and its own documentation.

Not in scope:

- Any change to shipped TUI or backend code. If the seam is not already sufficient, that is a separate issue.
- Synthesizing or hand-authoring events. Fixtures are recordings of real runs.
- Interactive request paths that a recording cannot answer (starting a run, resuming, live chat against an agent). Replay is one-directional by nature.
- A replacement for the unit tests, or a screenshot/golden-frame regression suite in CI. This is a development loop, not a gate.

- The TUI already connects to whatever socket `VIBESYS_CONTROL_SOCKET` names, so the server side of the control protocol is the natural seam and no product code should have to change to accommodate this.
- `run-events.jsonl` lines are already the on-wire `RunEvent`, so a recording should be usable as a fixture without a translation step that could drift from the schema.
- `RunEvent` tolerates older records that omit newer fields. Keeping a pre-schema-change capture is a useful check that renderers still degrade to raw content.
- Fixtures are about a megabyte of mostly-repeated shell output per run, so raw captures should not go into the tree uncompressed.
- `clients/tui/tsconfig.json` sets `rootDir: "src"` and `package.json` publishes only `dist`, so tooling placed outside `src/` is already excluded from the shipping path. Note that it is then also outside `tsconfig.check.json`, so CI will not typecheck it.

### Verification approach

- Replay each bundled fixture against the built TUI and confirm the expected end state renders: a completed lifecycle for one, the error banner for the other.
- Capture frames at a fixed size through the scripted path and confirm the panes, wrapping, and status line are populated rather than empty.
- Confirm nothing ships: `dist` contains no file from the tooling directory, `pnpm --dir clients/tui pack` excludes it, and no `src/` file imports it.
- Scan the fixtures for usernames, home paths, emails, and credential-shaped strings.
- Run the existing TypeScript checks (`pnpm check:ts`, `pnpm check:ts-architecture`, `pnpm check:clients`, `pnpm test:clients`) to confirm the addition does not disturb them.

Parent: #284. Directly reduces the cost of the open TUI rendering work under it, including #516, #514, #465, and #464.

Nano-AI:
Closes #542.

## Problem

Every TUI rendering change costs a real optimization run to look at. The TUI will not start without a live Python backend (`src/index.ts:13-14`), so checking a layout or theme change means an agent invocation, tokens, and minutes before the first interesting event. Terminal states, typed `tool_result` payloads, and older-schema records cannot be produced on demand at all.

## Solution

`dev/mock-server.ts` speaks the server side of the control protocol and replays a recorded `run-events.jsonl`. The binary under test is the unmodified `dist/index.js`, so no product code changes. Journal lines are already the on-wire `RunEvent`, so replay is re-enveloping, not translation.

Two real captures, scrubbed and committed plain. `bad-cpp-round1` is the only one with a full lifecycle; `queue-rs-payloads` carries a typed payload on all 105 `tool_result` events and ends in `run_failed`, so it exercises the error banner.

```bash
pnpm --dir clients/tui build
clients/tui/dev/mock-ui.sh --speed 0        # interactive
clients/tui/dev/mock-ui.sh --tmux 150x40    # scripted capture
```

## Verification

`tests/server/test_tui_dev_harness.py` (17 tests) declares each fixture `CURRENT` or `LEGACY`: every line must validate as a `RunEvent`, and the current-schema one must round-trip byte for byte. The mock's eleven response bodies live in `dev/mock-responses.json` so the test validates the bytes actually sent, which caught three out of contract.

Nothing ships: `dev/` is outside `rootDir`, absent from `dist` and from `pnpm pack`'s 66-entry tarball, and a `src/` to `dev/` import is rejected by a dependency-cruiser rule.

**Known limit:** CI installs neither node nor bun, so the mock check validates the file, not a live exchange. Verified manually over the socket.

| Check | Result |
| --- | --- |
| `pytest tests/server/test_tui_dev_harness.py` | 17 passed |
| `pnpm test:clients` | 485 pass, 0 fail |
| `check:ts` / `check:ts-architecture` / `check:clients` | clean |
| `check_format` / `check_lint` / `check_doc_links` | clean |


vic-lsh:
@Nano-AI Nice work! Here's some things to address before merging:

## 1. Add a CI check that the fixtures match the schema

Right now nothing ties the fixtures to `RunEvent`. The TS client never validates events, so after a rename or removal the harness keeps rendering via the raw-`content` fallback and nobody notices. The "schema tolerance is exercised" property is a manual observation today, not a check.

Bidirectional checks:

- Every fixture line must pass `RunEvent.model_validate` (catches breaking changes: renamed/removed fields, event types, payload kinds).
- The fixture designated "current schema" must round-trip exactly: `model_dump(mode="json") == raw line` (catches additive drift too, and names the key).

I ran this locally against main:

| Fixture | validates | round-trips | drifted keys |
| --- | --- | --- | --- |
| queue-rs-payloads | yes | yes | none |
| bad-cpp-round1 | yes | no | `execution_id`, `data` |

The mock's hand-written `snapshot`/`tui_defaults`/`chat_options`/`ack` bodies have the same problem at lower stakes. Validating them against `protocol.schema.json` in the same test would be nice.

## 2. Commit the fixtures as plain JSONL, not gz

Git zlib-compresses blobs, so the repo cost is identical:

| Fixture | raw | gzip | zlib (git) |
| --- | --- | --- | --- |
| bad-cpp-round1 | 731 KB | 151 KB | 151 KB |
| queue-rs-payloads | 978 KB | 116 KB | 116 KB |

Plain files are greppable, the scrub is auditable in the PR view (I had to gunzip to check it), and a hand-edit after a rename is a readable diff. The mock already reads both forms. If diff noise on re-record is a concern, mark them `-diff` in `.gitattributes`.

## Smaller

- `dev/` is not typechecked in CI. Add a `tsconfig.check.json` override of `rootDir`/`include` covering `dev/**`.
- Add a one-line pointer to `dev/README.md` from the Validation section of `docs/contributing/tui-architecture.md`.

Nano-AI:
All four are in, now `ac9be23`. The schema test declares each fixture CURRENT or LEGACY rather than hardcoding a skip, and caught a real one immediately: `markdown.jsonl` on the branch above had `judge_verdict: "accepted"` and a string in the numeric `perf_metric`. I also moved the mock's eleven response bodies into `dev/mock-responses.json` so the test validates the bytes actually sent rather than a copy, which found three more out of contract. Two deliberate departures: I did not mark the fixtures `-diff`, since that renders them as `Bin` and hides the scrub plain files were meant to expose; and widening `tsconfig.check.json` let a `src/` → `dev/` import typecheck clean, so that rule now lives in `.dependency-cruiser.cjs` rather than relying on the build to catch it. One caveat worth stating: CI installs neither node nor bun, so the mock check validates the file, not a live exchange.

Nano-AI:
Correction to point 3 above, found after posting.

Widening `tsconfig.check.json`'s root to cover `dev/**` has a cost I did not account for: with the harness in the check program, a `src/` file importing `dev/` now typechecks cleanly, where before it failed. I verified it with a probe import:

| | before | after |
| --- | --- | --- |
| `pnpm check:clients` | fails | **passes** |
| `pnpm build:clients` | fails (TS6059) | fails (TS6059) |

So the "nothing ships" property was still enforced, but only by the build, and only as a side effect of `rootDir`. The check named for catching type errors had started accepting the one import that breaks the invariant.

Fixed by stating the rule where the other layering rules live, rather than leaving it to a later step:

```js
name: 'shipping-path-does-not-depend-on-dev-harness',
severity: 'error',
from: {path: '^clients/tui/src/'},
to: {path: '^clients/tui/dev/'},
```

Confirmed it fires on the probe (`clients/tui/src/__leak_probe.ts → clients/tui/dev/mock-server.ts`) and that a clean tree still passes. `dev/README.md` records the tradeoff.

Amended into the same commit, now `ac9be23`.

Nano-AI:
Superseded by #569, which is part of native stack #573. This PR could not join a stack because its head was on a fork, and a PR's head cannot be changed after creation. Same branch, same commits, same reviewers.

Issue #433, opened 2026-08-25
Indicate focus consistently on every TUI pane


vic-lsh:
## Observed behavior

Focus indication is inconsistent on the hypothesis landing view. Chat and performance panes change their border/title when focused, but the Experiments pane always uses the accent border and the static title `Experiments`, regardless of whether it owns keyboard input. The primary command input also retains its success-colored border when another pane is focused.

Clicking visible pane content also does not consistently select that pane. Click handling existed only on some controls, and OpenTUI routes a mouse event to the innermost handled surface, so clicks on pane backgrounds, scroll surfaces, transcript cards, experiment rows, agent nodes, and performance content could leave keyboard focus on a different pane.

This makes pane-specific actions appear broken. For example, Escape closes the performance pane only while that pane is focused; when focus is visually ambiguous or a click does not transfer focus, users cannot tell whether Escape will dismiss Performance or act on another surface.

## Expected behavior

Every visible pane should use one consistent, unmistakable focused treatment. Unfocused panes should use a neutral border so keyboard and scrolling targets are clear. Clicking anywhere within a pane should transfer semantic focus to it without suppressing the clicked control's selection action.

## Reproduction

1. Start the TUI wide enough to dock Experiment chat.
2. Run `/perf` to add the performance pane.
3. Press `Ctrl+W` to cycle through chat, Experiments, and Performance.
4. Observe that chat and performance advertise focus, but the Experiments pane does not change when it becomes selected.
5. Click the body or inner content of another visible pane.
6. Press an arrow key, Tab, or Escape and observe that the action can still be routed to the previously focused pane.

## Affected subsystem

CLI and packaging

## Environment

- Commit: `063cffe54c4bbdc40ba213c28296a32fe34bfffe`
- OS: macOS
- Frontend: wide TUI layout, high-contrast-dark theme

## Impact

Workaround exists. Keyboard navigation and performance-pane dismissal work, but users must infer which pane owns input, scrolling, or Escape and use keyboard focus cycling instead of direct pointer selection.

## Relevant logs

```text
Ctrl+W changes layout.focus to `left`, but the Experiments title and border remain unchanged.
A mouse-up inside nested pane content does not consistently update layout.focus or roundFocus.
```

## Related issues or pull requests

#283 required focus distinguishability for split panes but does not cover this hypothesis-index gap.

## Acceptance criteria

- [ ] Every focusable pane consumes the same authoritative focus state.
[...]

PR #544, opened 2026-09-02
fix(tui): make pane focus unambiguous and legible in every theme

### LINKED ISSUE #561: TUI: pane focus is hard to distinguish in five built-in themes
### Observed behavior

Pane focus changes the border colour and inserts `▸` before the title. The
focused and resting border colours are less than 1.5:1 apart in five of the
eight built-in themes. In `high-contrast-light`, the focused border is less
prominent than the resting one.

Inserting the marker also moves the title two cells whenever focus changes.

### Expected behavior

Focused panes differ from resting panes without relying on colour. The title
stays in the same position when focus moves.

### Reproduction

Run:

```sh
clients/tui/dev/mock-ui.sh --tmux 150x40 --speed 0 --theme high-contrast-light
```

Press `Ctrl+W`, then capture the pane with `tmux capture-pane -e -p`.

### Impact

The active pane is difficult to identify in most themes. This is especially
noticeable in the high-contrast themes.

### Acceptance criteria

- [ ] Focus has a visible non-colour treatment in every built-in theme.
- [ ] A focused border never becomes less prominent than its resting border.
- [ ] Pane titles do not move when focus changes.
- [ ] Keyboard and mouse focus changes update the treatment together.

Roadmap #284. #433 and PR #434 established pane-focus semantics. Blocked by PR
#544, which fixes the remaining cases where the wrong surface is marked. This
issue covers the legibility of that mark.

Nano-AI:
Closes #561. Regression of #433, which PR #434 recoloured rather than resolved.

## Problem

Two problems with one cause: nothing owned what focus means.

**Which surface is focused had no single authority.** `experiment-log` and the transcript asked `focusedPane(state)`, `agent-map` asked `state.roundFocus`, and `command-input` was driven by a pane-focus expression despite the command box being shared by every pane in its column. Two boxes lit at once.

**How focus is shown was carried by colour alone.** Across the eight built-in themes, focused-versus-resting border contrast is 4.33 at best and under 1.5 in five. `high-contrast-light` was inverted: focus at 11.22 against the canvas was *quieter* than the 12.63 resting border, so focusing a pane made it recede.

before and after

## Solution

Every surface compares its own `PaneId` against `focusedPane`, documented as the sole authority. `focus.ts` owns three channels, colour last:

1. Marker in a **reserved** title gutter, so focus swaps a glyph rather than inserting one and the label never shifts.
2. `high-contrast-light` focus becomes navy `#000080` (16.01), removing the inversion.

A heavier frame was tried as a second non-colour channel and dropped. Unicode has no heavy rounded corner: the arcs at U+256D-2570 are light-only and the heavier weights are square, so the swap traded rounded corners for square ones. That reads as the pane becoming a different kind of object rather than a louder one. Focus never changes a pane's shape.

No other palette changes: every resting border already sits below the 3:1 non-text floor, so there is no headroom to quiet one further.

The command box calls the same helpers but only ever in the resting form. It takes the title so its label aligns with the panes above, never the marker or the heavy frame.

The high-contrast palettes are deliberately small, so no retune fixes this. Colour cannot be the primary channel: WCAG 1.4.1 and 2.4.7.

## Verification

`focus.test.ts` pins the invariant **for every built-in theme**: a focused pane differs from a resting one by at least one non-colour channel, and no theme makes a focused border quieter than a resting one.

| Check | Result |
| --- | --- |
| `bun test clients/tui/src/` | 524 pass, 0 fail |

Reverting the source while keeping the tests fails exactly the new assertions.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -93,12 +93,15 @@ list rises out of the command input rather than across the chat.
 
 The cursor starts in the command input, and `Ctrl+W` moves both it and the pane
 keys to the chat and back; the chat's instruction line says so (`Ctrl+W to type
-here`) until it holds them, and the focused input carries the focus border, so
-where a keystroke lands is never a guess. Only the chat composer accepts
-ordinary questions; the command input accepts slash commands. Page Up and Page Down scroll the focused pane, and
-Escape hands the keys back to the table. Arrows, Enter, and the rest of the
-table's keys are unaffected by the dock. A slash command typed into the chat
-input runs through the same command path as the command box.
+here`) until it holds them, and the chat pane carries the focus border while it
+does, so where a keystroke lands is never a guess. The command box is shared by
+every pane in its column rather than being one of them, so it keeps the neutral
+border in every theme and never competes with the pane that holds the keys.
+Only the chat composer accepts ordinary questions; the command input accepts
+slash commands. Page Up and Page Down scroll the focused pane, and Escape hands
+the keys back to the table. Arrows, Enter, and the rest of the table's keys are
+unaffected by the dock. A slash command typed into the chat input runs through
+the same command path as the command box.
 
 `/chat` leaves the command surface on this view, since the chat is already
 beside the table: it is absent from `/help` and from the completions, though it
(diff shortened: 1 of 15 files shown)


Nano-AI:
Superseded by #572, which is part of native stack #573. This PR could not join a stack because its head was on a fork, and a PR's head cannot be changed after creation. Same branch, same commits, same reviewers.

PR #545, opened 2026-09-02
fix(tui): render transcript tables and fenced code blocks readably

### LINKED ISSUE #562: TUI: render transcript tables and fenced code blocks readably
### Workstream

CLI/DX

### Problem

Markdown tables use the renderer defaults. Short cells stretch across the pane,
wide tables overflow their card, and the fixed grey border is faint on light
themes.

Fenced code blocks have no distinct foreground or background, so they blend
into the surrounding prose. During streaming, a block can also remain blank
when its declared language has no available grammar.

### Desired outcome

Tables fit their card and follow the active theme. Fenced blocks remain readable
with or without a language grammar and keep the renderer's normal spacing and
streaming behaviour.

### Acceptance criteria

- [ ] Tables size to content, wrap when needed, and never overflow their card.
- [ ] Table borders use the active theme.
- [ ] Fenced blocks use a distinct surface in every built-in theme.
- [ ] Unknown and unlabelled languages render as readable plain text.
- [ ] Other markdown rendering is unchanged.

### Verification approach

Use a synthetic replay fixture containing narrow and wide tables plus labelled,
unlabelled, and unknown-language fences. Check narrow and wide terminals in one
light and one dark theme.

Parent: #284. Implemented by PR #545. Syntax highlighting is a separate
follow-up.

Nano-AI:
Closes #562.

> Stacked on #543. Base is `adi/tui-mock-harness`, so the diff is this change alone. GitHub retargets to `main` when #543 merges.

## Problem

Tables and fenced code blocks are the two markdown constructs the transcript claims to support and does not render usably.

Tables take the renderer's defaults, so columns stretch to the pane width, a six-column table overflows its card, and the border is a fixed `#888888` that ignores the theme. Fenced blocks arrive with no foreground and no background, visually identical to the prose around them.

before and after

## Solution

Tables size to content, wrap on word boundaries, fit inside the card, and take the theme's border colour. Fenced blocks draw on the theme's code surface.

Two things worth a close look:

**The override is marked `codeBlockOnly`.** Without it a prose-only message with 100 paragraphs creates 100 native renderables instead of one. Note `createMarkdownCodeBlockRenderer` could not be the dispatcher, only the source of that marker: it is a per-language table that bails on any fence without a recognized info string, which is most of a transcript.

**The block is styled, not replaced.** Mutating the `CodeRenderable` that `defaultRender()` returns is what keeps `canUpdateInPlace` true, which restores the margin after a fence and preserves streaming, concealment and the tree-sitter client.

## Verification

| Check | Result |
| --- | --- |
| `bun test clients/tui/src/` | 502 pass, 0 fail |

Tests build a real `MarkdownRenderable` wired like `ConversationView`, with a variant omitting the override as the control. Neutering the marker fails 2 of them; restoring it passes all 17.

The `markdown.jsonl` fixture is synthetic because no recorded run contains a table or a fenced block. It includes a table split mid-row across three chunks to exercise streaming.

**Follow-up, not fixed here:** `applyCodeBlockRenderable` resets a code block's colours on in-place update. Unreachable today because the card is rebuilt on every change, but live if the transcript ever updates markdown in place.


Nano-AI:
Superseded by #570, which is part of native stack #573. This PR could not join a stack because its head was on a fork, and a PR's head cannot be changed after creation. Same branch, same commits, same reviewers.

PR #546, opened 2026-09-02
fix(tui): give the header real content and its own housing

### LINKED ISSUE #517: TUI: replace the raw status/metadata line above the rounds strip with a curated header
### Workstream

CLI/DX

### Problem

The single header line above the rounds strip leaks internal identifiers that mean nothing to an operator and change constantly. Observed live: `VibeSys · running · implementer · round-1-retry-1-implementer · 118k/400k tokens · Preallocated lock-free SPSC ring · r1`. The third and fourth segments are the raw backend `agent_kind` and `round_label` strings (`statusText` in `clients/tui/src/session-model.ts:1765-1773` interpolates `state.core.roundLabel` verbatim), so users see loop-internal labels like `round-1-retry-1-implementer`, `round-1-plan`, `round-1-pre`, and in other loops `gen-2-cand-1-mutator` or `impl issue #7 att2`, duplicating information the rounds strip and Agents pane already present properly. The token meter is also misleading: it renders cumulative run tokens over a single context window, so it read `472k/400k tokens` (118 percent) during the round-1 judge, which looks like an overflow error. The same header also appends the hypothesis scope label and selected agent kind, producing lines like `VibeSys · running · implementer · round-1-retry-2-implementer · 223k/400k tokens · Preallocated lock-free SPSC ring · r1` that crowd out the information that matters (what is running and for how long). This is the "random metadata at the top of the round strip" complaint.

### Desired outcome

The header communicates run state in operator terms: run status (running, pausing, paused, failed, done), the current phase in plain words (for example `implementer · attempt 3`), the active hypothesis title, and a token or cost meter with an honest denominator (cumulative tokens without a fake limit, or per-call context usage labeled as such). Raw `round_label` strings and other internal identifiers no longer appear anywhere in the header; the rounds strip and Agents pane remain the detailed views. Anything dropped from the header that has diagnostic value stays reachable (for example in `/help` or a detail view

### Acceptance criteria

- [ ] No raw backend `round_label` string (`round-N-retry-M-<kind>`, `gen-*`, `impl issue #*`) is rendered in the header in any loop mode.
- [ ] The token meter never displays a value over its denominator; cumulative and per-context numbers are not mixed.
- [ ] Paused/pausing state is visible in the header (shared with the /pause feedback fix).
- [ ] Header contents fit 200-column and 100-column widths without truncating the hypothesis title ahead of less useful segments.
- [ ] Existing header tests in `clients/tui/src` updated; a snapshot test pins the new format.

### Scope and non-goals

In scope: 
`statusText` and header assembly (`session-model.ts:1765-1773`, `ui/app.ts:315-319`) and the phase labeling helper needed to humanize `agent_kind`/attempt. 

Not in scope: 
the rounds strip chips themselves (they already render clean `r1`/`r2` labels), the Agents pane heading, or changing the backend labels (they are load-bearing for round attribution in `run-map.ts`).

`roundNumberFromLabel` regex parsing in `clients/core-state/src/run-map.ts:56-60` depends on the raw labels internally; this change is presentation-only and must not alter the projection. Header width budgeting interacts with the responsive breakpoints, so the segment drop order must be explicit.

### Verification approach

Unit test on the header formatter across loop modes and statuses, plus live frame captures at 200x50 and 100x40 during a real run confirming no internal labels and a sane meter.

#284

Nano-AI:
Closes #517.

> Stacked on #545. Base is `adi/tui-markdown-render`, so the diff is this change alone. GitHub retargets to `main` when #545 merges.

## Problem

The header interpolated backend identifiers verbatim, so a live run read:

```
VibeSys · running · implementer · round-1-retry-1-implementer · 2/200k tokens · Preallocated lock-free SPSC ring · r1
```

`round-1-retry-1-implementer` is a loop-internal key that duplicated both the agent kind beside it and the rounds strip below. The token meter rendered 118 percent during a real run, because `input_tokens` is the last call's context pressure while `context_window` is a static lookup. At 100 columns the line clipped mid-title, losing the one segment saying what the run was trying.

It also rendered into a bare one-row line while every other region sits in a bordered box, so it read as a caption above the UI rather than part of it.

before and after

## Solution

**Content.** `phase-label.ts` parses every loop's labels into operator-facing words (`implementing · attempt 2`); an unrecognized label degrades to the agent kind rather than printing an identifier. Segments drop whole in priority order rather than clipping, and the hypothesis title outranks the meter, the selection note and the hint. Width is measured in terminal cells via `text-width.ts`, not UTF-16 code units, so CJK and emoji budget correctly.

Pause is a tri-state (`paused | resumed | null`) rather than a boolean, because a boolean cannot distinguish "explicitly resumed" from "issued neither" — that is why a paused boot snapshot stayed paused after `/resume`. A terminal status always outranks a local override.

**Housing.** The header is now a pane: three rows, bezel to bezel, no fill, no blank separator, so its bottom bezel meets the next pane's top one. An earlier attempt used a surface band and a rule, which was worse: a partial-width fill is exactly what reads as floating in a terminal.

**Cost:** two extra rows. On a short terminal that is two rows of transcript.

## Verification

| Check | Result |
| --- | --- |
| `bun test clients/tui/src/` | 541 pass, 0 fail |

Tests cover the pause tri-state, CJK and emoji budgeting, grapheme-safe truncation, and the width ladder at 200/100/60/34/22/16 columns. Backend labels are unchanged on the wire, so round attribution in `run-map.ts` is unaffected.

### DIFF
--- a/clients/tui/dev/fixtures/markdown.experiments.json
+++ b/clients/tui/dev/fixtures/markdown.experiments.json
@@ -1,0 +1,21 @@
+[
+  {
+    "hypothesis_id": "h1",
+    "title": "Preallocated lock-free SPSC ring replaces the Mutex<VecDeque> hot path",
+    "claim": "Removing the mutex from the producer/consumer path removes the dominant cost.",
+    "action": "Replace the queue with a bounded SPSC ring and keep the C ABI identical.",
+    "first_round": 1,
+    "last_round": 1,
+    "rounds": [{"round": 1, "passed": true, "reviewed": true, "hypothesis_outcome": "proven"}],
+    "resolved_outcome": "proven",
+    "judge_verdict": "accepted",
+    "perf_metric": 301.4,
+    "perf_unit": "ms",
+    "perf_metric_name": "total_ms",
+    "perf_direction": "min",
+    "perf_baseline_value": 412.7,
+    "perf_delta_pct": -27.0,
+    "kept": true,
+    "active": false
+  }
+]
--- a/clients/tui/dev/mock-server.ts
+++ b/clients/tui/dev/mock-server.ts
@@ -152,6 +152,24 @@ function resolveFixture(name: string): string {
   throw new Error(`no fixture found for ${name}`);
 }
 
+/**
+ * Hypotheses for the experiment log, from `<fixture>.experiments.json` beside
+ * the fixture when one exists.
+ *
+ * Recorded journals hold events, not the projected experiment table, and the
+ * header and log render the hypothesis title rather than anything in the event
+ * stream. A sidecar keeps that title pinnable without inventing event types the
+ * backend never emits.
+ */
+function loadExperiments(fixturePath: string): unknown[] {
+  const sidecar = `${fixturePath.replace(/\.gz$/, '').replace(/\.jsonl$/, '')}.experiments.json`;
+  try {
+    return JSON.parse(readFileSync(sidecar, 'utf8')) as unknown[];
+  } catch {
+    return [];
+  }
+}
+
 function loadFixture(path: string): RunEventRecord[] {
   const raw = readFileSync(path);
   const text = path.endsWith('.gz') ? gunzipSync(raw).toString('utf8') : raw.toString('utf8');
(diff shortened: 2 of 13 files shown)


Nano-AI:
Superseded by #571, which is part of native stack #573. This PR could not join a stack because its head was on a fork, and a PR's head cannot be changed after creation. Same branch, same commits, same reviewers.

PR #556, opened 2026-09-02
fix(tui): line up the bottom input boxes and stop the chat modal overlapping itself

### LINKED ISSUE #563: TUI: chat composer layout shifts and overlaps at common terminal sizes
### Observed behavior

Three small layout defects affect the chat composer:

- The Message box ends two rows above the Command box at 150x40.
- The modal transcript overlaps the composer, or leaves a gap above it, at
  terminal heights where its percentage-based edges round differently.
- The chat suggestion menu leaves the hint row between the menu and Message
  box; the command menu sits flush on its box.

### Expected behavior

The Message and Command boxes share a bottom row. The modal transcript ends
where its composer begins. Both suggestion menus sit flush on their input box.

### Reproduction

Run `mock-ui.sh --tmux 150x40 --speed 0` for alignment and type `/` in each
input for the menus. Run at 70x24 and enter `/chat` for the modal overlap. Pin
detached captures with `tmux resize-window` before reading row positions.

### Acceptance criteria

- [ ] Message and Command share a bottom row while chat is docked.
- [ ] Modal transcript and composer never overlap or leave a gap after resize.
- [ ] Both suggestion menus touch the box they complete.

Roadmap #284. Implemented as three independent commits in PR #556 because all
three changes are confined to the composer layout.

Nano-AI:
Closes #563.

## Problem

Three defects in the bottom of the screen, where you type.

**The Message and Command boxes did not line up.** Two separate one-row asymmetries: the two columns ordered hint and box differently, and the docked chat pane is bordered so its bottom border owns the terminal's last row while the command column has no border. Measured at 150x40: Message bottom on row 38, Command bottom on row 40.

The bottom five rows: the Message and Command boxes sharing a top and a bottom edge, each hint on the row above

**The chat modal overlapped its own transcript.** Its edges were fractions of a row (`top: 10%, height: 76%`), and Yoga rounds a child's offset independently of its size:

```
transcript end = round(0.86H - 5)
composer start = round(0.1H) + round(0.76H - 5)
```

These disagree at **25 of the 111 heights** from 10 to 120. The sign varies: some overlap by a row, others leave a gap and push the input onto the border.

**The slash-command menu floated a row above its box** instead of sitting flush on it, unlike the command bar's.

## Solution

Both columns read hint-then-box, and `app.ts` adds a one-row inset when the chat is docked. The modal computes its four edges in whole cells, which reproduces the rectangle the layout already painted at every size, and hands the composer the width it just assigned rather than one read back out of the layout. The menu anchors on the box height rather than the whole composer.

**Cost:** when the chat is docked and the landing table can spare the row, the terminal's last row carries only the chat pane's closing border. A terminal too short to spare it keeps the row and the two boxes sit one row apart. Undocked, the layout is byte-identical to before.

## Verification

The modal test resizes across heights 18 to 48 and pins the invariant **per height** rather than at one size, since the earlier one-size sweep is what hid this. Each test was confirmed to fail without its fix. Rebased onto `adi/focus-contrast` at bfd2607 (itself on `main` at 8e09827), which moved the nested `Message` composer to the resting frame; no conflicts and no test asserted the old chrome.

| Check | Result |
| --- | --- |
| `pnpm --filter @vibesys/tui test` | 679 pass, 0 fail, 28 files |
| `pnpm --filter @vibesys/tui check` | clean |
| `pnpm exec biome check clients/tui/src clients/tui/dev` | clean, 68 files |
| `pnpm check:ts-architecture` | clean |

### DIFF
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -2183,6 +2183,45 @@ describe('OpenTUI presentation', () => {
     expect(controller.state.chatOpen).toBe(false);
   });
 
+  it('stacks the modal chat without a seam at any terminal height', async () => {
+    // The modal takes a share of the terminal, and a share of a row is not a
+    // row: the layout rounds a child's offset from its parent separately from
+    // that child's size, so at a fractional offset the two disagree and the
+    // transcript either runs a row into the composer or leaves a row of the
+    // modal's floor blank. Which heights round badly follows from the shares
+    // rather than from any one size, so the whole range is checked.
+    const testRenderer = await createTestRenderer({width: 100, height: 24});
+    const controller = new FakeController({...initialSessionState(), chatOpen: true});
+    const app = createOpenTuiApp(testRenderer.renderer, controller);
+    registerCleanup(testRenderer.renderer, app);
+
+    for (let height = 18; height <= 48; height += 1) {
+      testRenderer.renderer.resize(100, height);
+      await frameAfter(testRenderer);
+      const modal = testRenderer.renderer.root.findDescendantById('chat-overlay');
+      const transcript = testRenderer.renderer.root.findDescendantById('chat-transcript');
+      const composer = testRenderer.renderer.root.findDescendantById('chat-modal-composer');
+      if (modal === undefined || transcript === undefined || composer === undefined)
+        throw new Error('modal chat geometry was missing');
+      // The height rides along in the comparison so a failure names the
+      // terminal it happened on, not only the rows that disagreed.
+      expect({
+        height,
+        top: transcript.y,
+        end: transcript.y + transcript.height,
+        floor: composer.y + composer.height,
+      }).toEqual({
+        height,
+        // The transcript starts under the top border and ends exactly where
+        // the composer starts, and the composer's last row is the one above
+        // the bottom border.
+        top: modal.y + 1,
+        end: composer.y,
+        floor: modal.y + modal.height - 1,
+      });
+    }
+  });
+
   it('accepts another chat message while an agent turn is pending', async () => {
     const testRenderer = await createTestRenderer({width: 100, height: 24});
     const controller = new FakeController({
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -64,6 +64,7 @@ import {
 } from '../session-model.js';
 import {TRANSCRIPT_MIN} from './agent-map.js';
 import {createOpenTuiApp, type OpenTuiApp} from './app.js';
+import {MIN_DOCK_WIDTH} from './chat-pane.js';
 import type {ClipboardCopyResult, SelectionClipboard} from './clipboard.js';
 import {renderDesignSummary} from './design-log.js';
 import {paneTitle} from './focus.js';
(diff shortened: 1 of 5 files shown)


vic-lsh:

AyanBinRafaih:
@Nano-AI Here are a few comments:

1. The new command inset adds a row whenever docked chat is visible:

      const commandInset = showChatPane ? 1 : 0;
      bottom.paddingBottom = commandInset;

That row comes out of the experiment landing area. The diff also changes the existing kickoff test from 100x16 to 100x18, even though the three-row header cited by the new comment is already present in this PR's base commit. In other words, the test adjustment hides a layout regression introduced by this change rather than accommodating an unrelated upstream change. The alignment fix should not silently stop supporting a terminal size that passed on the base. At short heights, the layout should degrade explicitly, for example by dropping or combining a hint row, or only applying the inset when enough vertical space remains.

Please restore coverage at 100x16 and assert that the kickoff content and command surface remain visible without overlap or clipping. A separate 100x18 test can still cover the full presentation.

2. `setPaneBounds(null)` now assigns a new numeric modal width immediately before `render()` calls:

      this.#composer.activate(Math.max(1, this.output.width - 4), ...)

In OpenTUI 0.4.3, the `Renderable.width` getter returns `_widthValue`, which is the last computed Yoga layout. Assigning `output.width` only changes the pending style and requests another render. `_widthValue` is not updated until `updateFromLayout()` reads `yogaNode.getComputedLayout()` during the later layout pass. Consequently, after resizing an open modal, `this.output.width - 4` still describes the previous modal width when `ChatComposerView` calculates `wrappedRows()`. A draft near a wrapping boundary can retain the wrong editor height, and the suggestion menu is then anchored using that stale composer height. A resize does not guarantee a subsequent controller state update that would correct it. Please pass the modal content width calculated by `#applyModalGeometry()` directly into `activate()`, or defer composer sizing until after the updated layout has been computed. Add a resize regression test with an open modal and a draft that changes from one row to multiple rows at the new width.

3. The comment says that rounding the four edges keeps the modal rectangle unchanged, but the old and new calculations are not equivalent. The previous layout used `left: 10%` and `width: 80%`, so Yoga rounded the offset and width independently. The new implementation calculates width as:

      round(terminalWidth * 0.9) - round(terminalWidth * 0.1)

 At common widths this changes the rendered rectangle. For example:

  - Width 84: the old modal width is 67 cells, while the new calculation produces 68.
  - Width 86: the old modal width is 69 cells, while the new calculation produces 68.
  - Width 87: the old modal width is 70 cells, while the new calculation produces 69.

This changes transcript wrapping, composer wrapping, and the amount of visible content during ordinary terminal resizes. It is also inconsistent with the PR's stated property that only fractional-row disagreement is being removed while the existing rectangle remains unchanged. Please preserve the old width calculation with `Math.round(terminalWidth * 0.8)`, unless changing the horizontal geometry is intentional and documented. Add a width sweep alongside the new height sweep so these one-column oscillations are covered.

Nano-AI:
@AyanBinRafaih Thanks. Findings 1 and 2 were correct and are fixed.

The command inset is now conditional. At 100x16 the kickoff content remains visible and the boxes intentionally sit one row apart; in the tested kickoff state they align at 100x17. The 100x16 regression test is restored.

The modal now passes its newly calculated content width directly to the composer, and a resize test covers rewrapping an open draft.

I could not reproduce finding 3. OpenTUI rounds the absolute left and right edges, so the previous percentage layout matches `round(0.9W) - round(0.1W)`, including widths 84, 86, and 87. A width sweep now pins that behavior. One additional cleanup stops hidden rows from contributing to split-modal bounds.

PR #557, opened 2026-09-02
fix(tui): persist pause state in header and fix overlay z-order

### LINKED ISSUE #515: TUI: /pause gives no persistent indication that the run is paused, and slash-command behavior diverges between the command bar and the chat composer
### Observed behavior

`/pause` and `/steer` do reach the backend, but the TUI presents them so poorly that they appear broken, which is how the problem was originally reported. Reproduced live: `/pause` sent from the command bar during round-1 planning produced only a transient mid-screen overlay reading `pause: pending` (trivially missed, and dismissed by the next keypress), and after the planning call finished the run genuinely froze (the codex subprocess exited and the token meter stopped moving for minutes) while the header continued to read `VibeSys · running · orchestrator · round-1-plan · 118k/400k tokens` with no paused indicator anywhere on screen. There is no way to tell a paused run from a slow agent call. `/resume` then showed `resume: consumed` and the run instantly jumped to the implementer, confirming pause had been in effect the whole time while the UI said `running`. The backend never publishes a status transition on pause: `pause_after_call()` sets an in-memory flag (`src/vibesys/server/supervisor.py:1009-1012`), the block happens silently at the next `start_agent_execution` (supervisor.py:1046-1048), and only `snapshot()` derives `"paused"`, which the header does not reflect live.

The two input surfaces also diverge. The command bar and the chat composer parse commands with different tables and different fallbacks in `clients/tui/src/commands.ts`: chat autocomplete lists only `/clear`, `/model`, `/resume` (`suggestChatSlashCommands`, commands.ts:129-132), so typing `/pa` in chat shows an empty suggestion menu even though `/pause` works there when typed in full; `/resume` in a chat composer means "switch chat thread" (commands.ts:144) and shadows the global run-resume, so a paused run cannot be resumed from chat at all, including when the chat pane is F4-zoomed and the command column is hidden (app.ts:383); `/clear` and `/model` typed in the command bar return `Unknown command`; a typo like `/Pause` yields `Unknown command` in the command bar but a chat-help card in chat because the chat name regex is case-insensitive while table membership is case-sensitive (commands.ts:145-149, 169); and acks for `/pause`, `/steer`, and `/help` submitted from the modal chat render in `OverlayView` at zIndex 10 behind the chat modal at zIndex 20, so the operator sees no confirmation at all until the modal closes. Separately, the empty-input error banner (`Usage: /steer <message>`) is drawn at the top of the screen and the `/help` modal opens overlapping its footer row, the known overlay z-order collision.

### Expected behavior

Pausing a run is a state change and must be visible as state: the header status should read `pausing…` from the moment the request is acknowledged and `paused` once the boundary is reached, the pending/consumed transition should be observable without keeping a transient modal open, and `/steer` should confirm what text was queued and when it was consumed. The same command typed in any input surface should be parsed by the same rules, produce the same feedback in a visible location, and autocomplete consistently; where a name is genuinely ambiguous (`/resume`), the UI must still leave some path to the global command from every surface.

### Reproduction

1. Stage a standalone copy of the queue-rs example: `cp -r examples/data-structures/repositories/queue-rs /tmp/vibesys-runs/queue-hunt && cd /tmp/vibesys-runs/queue-hunt && rm -rf .git && git init && git add -A && git commit -m seed`.
2. Launch the TUI in tmux at 200x50: `node clients/tui/dist/launcher.js --project /tmp/vibesys-runs/queue-hunt --task spsc --profiler none --cli-provider codex --max-rounds 2` (with `[model] name = "gpt-5.6-sol"` in `agent.toml` so the codex CLI accepts the model).
3. While the orchestrator planning call is in flight, type `/pause` in the command bar and press Enter. Observe the only feedback is a transient `pause: pending` overlay and the header stays `running`.
4. Wait for the planning call to finish. Observe the token meter freezes and `ps` shows no codex child process, while the header still reads `running` indefinitely.
5. Type `/resume` and observe `resume: consumed` and the immediate jump to the implementer phase, proving the run had been paused with no indication.
6. Press Ctrl+W to focus the experiment chat, type `/pa`, and observe the empty suggestion menu; type `/pause` in full and observe it works via forwarding.
7. In the chat composer, type `/resume` and observe it opens the chat-thread resume menu instead of resuming the run.
8. Type `/steer` with no argument in the command bar and observe the `Usage: /steer <message>` banner at the top of the screen; open `/help` while the banner is up and observe the modal overlapping the banner footer row.

### Affected subsystem

CLI and packaging

### Environment

Commit: 1113353 (branch ayan/design-log). 
OS: Linux 6.1.0-50-amd64 (Debian). 
Terminal: tmux, 200x50, default dark theme. 
Provider: codex-cli 0.151.0, model gpt-5.6-sol. 
Backend: cpu, `--profiler none`. 
Workload: queue-rs spsc, `--max-rounds 2`.

### Impact

Blocks normal use

### Relevant logs

```shell
Header while actually paused (codex subprocess gone, token meter frozen for ~4 minutes):
VibeSys · running · orchestrator · round-1-plan · 118k/400k tokens · experiments

Ack overlays (only feedback, transient, mid-screen):
│ pause: pending   │  ... later ...  │ resume: consumed │

Empty /steer error colliding with /help modal:
╭─ Input error ────────────────────────────────────────────────╮
│ Usage: /steer <message>                                      │
│ [× Dismiss] · Esc: dismiss ·╭─ Help ─────────────────────────╮
```

#284

### Pre-submission checks

zatchbell1311-wq:
## Problem
When /pause is acknowledged, the only feedback is a transient mid-screen
overlay that disappears on the next keypress. The header continues to read
`running` for the entire duration the run is actually frozen, making a
paused run indistinguishable from a slow agent call. Additionally, command
ack overlays (zIndex 10) render behind the chat modal (zIndex 20), so
confirmations for /pause, /steer, and /help submitted from the modal chat
are never visible.

Closes #515

## Solution
Add a `runPaused: boolean` field to `SessionState`. The field is set to
true when a `command.pause` request is acknowledged and cleared to false
when `command.resume` is acknowledged. `statusText` prefixes `paused · `
to the header string when `runPaused` is true, giving a persistent visible
indicator for the duration of the pause.

Raise `OverlayView` zIndex from 10 to 25 so command acks always render
above the chat modal (zIndex 20).

## Architecture
No new components. Three files changed:
- session-model.ts — adds `runPaused` to SessionState interface and
  initialSessionState; statusText reads it to prefix the header
- session-controller.ts — sets/clears runPaused in submitCommand after
  the backend ack returns
- ui/overlay.ts — zIndex 10 → 25

### Correctness properties
- Header reads `paused · <status> · <agent> · <round>` from the moment
  /pause is acked until /resume is acked
- runPaused is false in initialSessionState so fresh sessions are unaffected
- OverlayView always renders above ChatOverlayView

### Testing
bun test --max-concurrency 1 src
480 pass, 0 fail, 3238 expect() calls across 20 files

### DIFF
--- a/clients/tui/src/session-controller.ts
+++ b/clients/tui/src/session-controller.ts
@@ -905,6 +905,11 @@ export class SocketSessionController implements SessionController {
     try {
       const response = await this.client.request(parsed.request);
       const rendered = renderResponse(parsed.request, response, parsed.responseView);
+      if (parsed.request.type === 'command.pause') {
+        this.#setState({...this.#state, runPaused: true});
+      } else if (parsed.request.type === 'command.resume') {
+        this.#setState({...this.#state, runPaused: false});
+      }
       if (rendered !== null) this.#setState(showDetail(this.#state, rendered));
     } catch (error) {
       this.#setState(reportCaughtError(this.#state, error, 'request'));
--- a/clients/tui/src/session-model.ts
+++ b/clients/tui/src/session-model.ts
@@ -78,6 +78,8 @@ export interface SessionState {
   themePicker: ThemePicker | null;
   /** Root-level error state, independent of the active transcript or log view. */
   errorBanner: ErrorBannerState | null;
+  /** True from the moment /pause is acknowledged until /resume is acknowledged. */
+  runPaused: boolean;
 }
 
 export type ErrorSeverity = 'recoverable' | 'fatal';
@@ -299,6 +301,7 @@ export function initialSessionState(themeName: ThemeName = DEFAULT_THEME_NAME): 
     chatDockFits: true,
     themePicker: null,
     errorBanner: null,
+    runPaused: false,
   };
 }
 
@@ -1742,7 +1745,8 @@ export function showDetail(
 }
 
 export function statusText(state: SessionState): string {
-  const base = `${state.core.status} · ${state.core.agentKind ?? 'starting'} · ${state.core.roundLabel ?? 'no round yet'}`;
+  const pausePrefix = state.runPaused ? 'paused · ' : '';
+  const base = `${pausePrefix}${state.core.status} · ${state.core.agentKind ?? 'starting'} · ${state.core.roundLabel ?? 'no round yet'}`;
   if (state.core.usage === null) return base;
   const used = formatTokenCount(state.core.usage.inputTokens);
   const meter =
--- a/clients/tui/src/ui/overlay.ts
+++ b/clients/tui/src/ui/overlay.ts
@@ -41,7 +41,7 @@ export class OverlayView {
       borderStyle: 'rounded',
       borderColor: theme.info,
       backgroundColor: theme.elevatedSurface,
-      zIndex: 10,
+      zIndex: 25,
     });
   }
 
--- a/src/vibesys/loops/agent/hypotheses.py
+++ b/src/vibesys/loops/agent/hypotheses.py
@@ -271,6 +271,7 @@ def project_round_evidence(
     *,
     prior_rounds: Sequence[RoundRecord],
     legacy_directions: Mapping[str, Literal["max", "min"]] | None = None,
+    noise_fraction: float = 0.0,
 ) -> Hypothesis:
     """Return a hypothesis updated with one completed round's evidence."""
     if record.hypothesis_id != hypothesis.hypothesis_id:
@@ -295,6 +296,7 @@ def project_round_evidence(
                     measurement.direction if measurement is not None else record.perf_direction
                 ),
                 benchmark_expected=record.official_evaluation,
+                noise_fraction=noise_fraction,
             )
         )
     else:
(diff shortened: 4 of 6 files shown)


vic-lsh:
FYI @zatchbell1311-wq: this is now fixed on the backend in #578.

Reviewing this PR surfaced two things a client-side flag could not get right: the `/pause` ack fires when the backend has only *recorded* the request, while the agent call in flight keeps running, so the header said "paused" during execution; and nothing cleared the flag when the run ended, so it read "paused · completed" for the rest of the session. Both are symptoms of mirroring backend-authoritative state on the client, which `docs/contributing/tui-architecture.md` rules out ("Only backend messages change core state"). #578 makes the run lifecycle a tested state machine on the Python side with an explicit `pausing` state, publishes every transition as a typed `run_status_changed` event, and has the TUI fold that event into its header.

Your z-order fix was a genuinely separate bug and it was right, so it is carried over in #578 with credit to you, plus a test renderer assertion that the overlay sits above the chat modal and below the theme picker. Thanks for finding and reporting both of these.

vic-lsh:
Closing in favour of #578, which fixes the pause indicator at its source on the backend and carries your overlay z-order change over with credit.

PR #558, opened 2026-09-02
fix(loops): share gate failure-output truncation lengths as named constants

### LINKED ISSUE #549: Share the gate failure-output truncation lengths as named constants
### Workstream

Core/runtime

### Problem

The accuracy gate and the benchmark gate each independently hardcode how much failure text is logged (last 1000 chars) versus fed back to the agent (last 4000 chars), and a third magnitude (last 8000 chars) is repeated across three more call sites. The logged and fed-back lengths must stay consistent to give the agents a coherent view of failures, but nothing links the copies, so changing one surface silently diverges the others. Evidence: `src/vibesys/loops/gates.py:77` (`output[-1000:]`) and `:78` (`output[-4000:]`), mirrored at `src/vibesys/loops/agent/loop.py:2160-2161`, with `output[-8000:]` at `loop.py:1754,1885,2131`.

### Desired outcome

A single shared definition of each truncation length (for example `GATE_LOG_TAIL_CHARS`, `GATE_FEEDBACK_TAIL_CHARS`, `GATE_RECORD_TAIL_CHARS`) imported by every gate, so the agent-visible failure window is defined once and stays consistent.

### Acceptance criteria

- [ ] The three truncation magnitudes exist as named constants in one module.
- [ ] Both gates and all record sites import them; no bare `[-N:]` slice literals for failure output remain.
- [ ] Loop tests that assert on failure feedback still pass.

### Scope and non-goals

In scope: 
- the failure-output truncation literals in the gates and loop. 

Not in scope: 
- changing the actual truncation values or the feedback format.

Keep the values byte-for-byte identical during the extraction so behavior is unchanged; this is a pure de-duplication.

### Verification approach

`uv run pytest tests/vibesys/loops`, plus a grep proving no residual failure-output slice literals.

#284

zatchbell1311-wq:
## Problem
The accuracy gate and benchmark gate each independently hardcode failure
output truncation lengths as bare slice literals: last 1000 chars logged,
last 4000 chars fed back to the agent, last 8000 chars recorded. Seven
copies across gates.py and loop.py are unlinked by the type system, so
changing one silently diverges the others and gives agents an incoherent
view of failures.

Closes #549

## Solution
Define GATE_LOG_TAIL_CHARS=1000, GATE_FEEDBACK_TAIL_CHARS=4000, and
GATE_RECORD_TAIL_CHARS=8000 as named constants in gates.py. All 7 bare
[-N:] slice literals in gates.py and loop.py are replaced with the
constants. Pure de-duplication — no behavior change.

## Architecture
No new modules. Constants live in gates.py (the natural owner of gate
policy). loop.py extends its existing gates import to include the three
constants alongside run_accuracy_gate.

### Correctness properties
- Values are byte-for-byte identical to the literals they replace
- A single definition site means future changes propagate automatically
  to all gates and record sites

### Testing
uv run pytest tests/vibesys/loops -x -q
548 passed, 0 failed

grep -n "output\[-[0-9]*:\]" src/vibesys/loops/gates.py src/vibesys/loops/agent/loop.py
(no output — zero residual literals remain)

### DIFF
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -45,7 +45,12 @@ from vibesys.loops.agent.model import (
 )
 from vibesys.loops.agent.state import AgentRunStateStore
 from vibesys.loops.evolve.population import Objective  # noqa: TC001  # tracked: #288
-from vibesys.loops.gates import run_accuracy_gate
+from vibesys.loops.gates import (
+    GATE_FEEDBACK_TAIL_CHARS,
+    GATE_LOG_TAIL_CHARS,
+    GATE_RECORD_TAIL_CHARS,
+    run_accuracy_gate,
+)
 from vibesys.loops.profiler import mcp_spec as profiler_mcp_spec
 from vibesys.profilers import (
     ProfilerKind,
@@ -1778,7 +1783,7 @@ def _run_framework_validation_gate(  # noqa: C901, PLR0912, PLR0915  # tracked: 
                 input_digest=input_digest,
                 passed=passed,
                 exit_code=execution.exit_code,
-                output=output[-8000:],
+                output=output[-GATE_RECORD_TAIL_CHARS:],
                 error=None if passed else "command exited nonzero",
             )
         except Exception as exc:  # noqa: BLE001  # tracked: #288
@@ -1909,7 +1914,7 @@ def _run_framework_accuracy_gate(  # noqa: PLR0913  # tracked: #288
         retry,
         command=result.command or "(not configured)",
         passed=result.passed,
-        output=result.output[-8000:],
+        output=result.output[-GATE_RECORD_TAIL_CHARS:],
     )
     ctx.snapshot_workspace(f"round-{round_number}-retry-{retry}-framework-accuracy")
     return result.feedback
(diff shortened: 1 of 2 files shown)


AyanBinRafaih:
@zatchbell1311-wq I found a couple of things that we can maybe change:

1. migrate_legacy() does not forward this value to _migrate_legacy_state(). When state.json does not exist, a configured value such as 0.03 is replaced by the default 0.0. A within-noise legacy result can therefore still become proven or disproven instead of inconclusive.

2. Please pass noise_fraction=noise_fraction at the call around lines 96-101 and add a test for migration without an existing unified state.

3. The server projection path is still missing the noise value. src/server/api/service.py:_agent_run_state() calls both migrate_legacy() and reproject_run_evidence() without the configured noise fraction. This means /experiments can still display proven/disproven when the recorded round outcome is inconclusive. Please thread the effective run value through these calls too.

Let me know what you think!

zatchbell1311-wq:
Hi Ayan, I think this feedback may be intended for a different PR. #558 is a pure de-duplication — it only extracts the three truncation length literals (GATE_LOG_TAIL_CHARS, GATE_FEEDBACK_TAIL_CHARS, GATE_RECORD_TAIL_CHARS) into named constants in gates.py and updates the 7 call sites. It doesn't touch migrate_legacy(), noise_fraction, or service.py at all. Could you double-check which PR this feedback was intended for?

AyanBinRafaih:
Yes, that's my mistake. I'll delete that to avoid confusion. I might have gotten something mixed up. Thanks for pointing it out!

vic-lsh:
Rebased this branch onto `main` before merging (`git rebase --onto origin/main f7e3ba0 pr558`) to drop two commits that belonged to #541 and #557 but had been picked up into this PR's history. The two commits kept are this PR's own: `932edde1` (extract gate failure-output truncation lengths as named constants) and `9e5a9f90` (move gate constants below imports for E402). No content of this PR's changes was altered by the rebase; `check_format.sh`, `check_lint.sh`, and `tests/vibesys/loops` (548 tests) all passed post-rebase, and CI passed on the rebased branch before merge.

PR #559, opened 2026-09-02
fix(tui): fill collapsed todo strip to box edge

### LINKED ISSUE #552: TUI: Fill the collapsed todo strip to the box edge
### Observed behavior

The collapsed todo strip truncates its summary line four columns before the box edge, leaving blank space and ellipsizing the current-todo text early, while the expanded strip fills correctly. The two disagree.

### Expected behavior

The collapsed summary should use the real content region of the collapsed box (width minus its 2 columns of padding), filling to the edge with no wasted space.

### Reproduction

1. Open a run with a todo list long enough that the collapsed summary line approaches the terminal width.
2. Observe the `▸ Todo x/y · <current item>` line is cut about four columns short of the edge.

### Affected subsystem

CLI and packaging

### Environment

### Impact

Blocks normal use

### Relevant logs

```shell
`clients/tui/src/ui/todo-strip.ts:193` computes `#contentWidth()` as `(renderedWidth ?? terminalWidth) - 6`, budgeting for the expanded layout's chrome (outer padding plus inner list border plus inner list padding). The collapsed box has no border and no inner list, only one column of padding each side, so its true region is `width - 2`; subtracting 6 cuts four columns.
```

#284

### Pre-submission checks

zatchbell1311-wq:
## Problem
The collapsed todo strip truncates its summary line 4 columns before the
box edge, leaving blank space and ellipsizing the current-todo text early.
The expanded strip fills correctly. The two disagree because #contentWidth()
hardcodes a subtraction of 6, which budgets for the expanded layout's chrome
(outer padding + inner list border + inner list padding). The collapsed box
has no border and no inner list — only 1 col of padding each side — so its
true content region is width - 2, not width - 6.

Closes #552

## Solution
Add an `expanded: boolean` parameter to `#contentWidth()`. Collapsed path
passes `false` → subtracts 2. Expanded path passes `true` → subtracts 6.
The 4-column gap is eliminated with no change to the expanded layout.

## Architecture
One file changed: clients/tui/src/ui/todo-strip.ts.
- #contentWidth(expanded) replaces the parameterless form
- Line 137 (collapsed summary): passes false
- Line 172 (expanded item): passes true

### Correctness properties
- Collapsed summary fills to width - 2 (1 col padding each side)
- Expanded items still use width - 6 (unchanged behavior)
- The two render paths no longer share a fixed chrome constant

### Testing
cd clients/tui && npm run build && bun test --max-concurrency 1 src
480 pass, 0 fail, 3238 expect() calls across 20 files

### DIFF
--- a/clients/tui/src/ui/todo-strip.test.ts
+++ b/clients/tui/src/ui/todo-strip.test.ts
@@ -33,6 +33,26 @@ describe('todo strip formatting', () => {
     expect(todoItemLine({content: 'Mystery step', status: 'deferred'}, 80)).toBe('? Mystery step');
   });
 
+  it('collapsed summary truncates to box width minus 2 columns of padding', () => {
+    // The collapsed strip has 1 col padding each side = 2 cols chrome.
+    // A long item should fill exactly to that boundary and no further.
+    const boxWidth = 40;
+    const contentWidth = boxWidth - 2;
+    const todos = [{content: 'x'.repeat(100), status: 'in_progress' as const}];
+    const line = todoSummaryLine(todos, contentWidth);
+    expect(line.length).toBeLessThanOrEqual(contentWidth);
+    expect(line.endsWith('…')).toBe(true);
+  });
+
+  it('expanded item truncates to box width minus 6 columns of chrome', () => {
+    // The expanded strip has outer padding + list border + list padding = 6 cols chrome.
+    const boxWidth = 40;
+    const contentWidth = boxWidth - 6;
+    const line = todoItemLine({content: 'x'.repeat(100), status: 'pending' as const}, contentWidth);
+    expect(line.length).toBeLessThanOrEqual(contentWidth);
+    expect(line.endsWith('…')).toBe(true);
+  });
+
   it('truncates long lines to the available width with an ellipsis', () => {
     const line = todoItemLine({content: 'x'.repeat(50), status: 'pending'}, 20);
     expect(line).toHaveLength(20);
--- a/clients/tui/src/ui/todo-strip.test.ts
+++ b/clients/tui/src/ui/todo-strip.test.ts
@@ -1,5 +1,11 @@
-import {describe, expect, it} from 'bun:test';
-import {todoItemLine, todoSummaryLine} from './todo-strip.js';
+import {afterEach, describe, expect, it} from 'bun:test';
+import {type Renderable, TextRenderable} from '@opentui/core';
+import {createTestRenderer, type TestRendererSetup} from '@opentui/core/testing';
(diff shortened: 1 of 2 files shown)


AyanBinRafaih:
@zatchbell1311-wq This is good work! Could you please add the following to this PR:

1. A fixed-width rendering test for this bug since he current test only checks that part of the todo text appears (so it might be passing with the old column width) 

2. A long collapsed item should verify that the ellipsis reaches the expected box width - 2 boundary. It would also be useful to toggle the list and verify that the expanded width remains box width - 6.

Thanks!

zatchbell1311-wq:
Thanks Ayan! Added two tests in commit 9bb472c:

collapsed summary truncates to box width minus 2 columns of padding — asserts the summary line fits within boxWidth - 2 and ends with …
expanded item truncates to box width minus 6 columns of chrome — asserts the item line fits within boxWidth - 6 and ends with …

vic-lsh:
Pushed 85bb5946 with a rendering regression test. The two width tests in 9bb472c7 pass unchanged against the pre-fix commit 12bb80c9: they compute `boxWidth - 2` / `boxWidth - 6` in the test body and pass that to the pure formatters `todoSummaryLine` / `todoItemLine`, which were never wrong, and never construct a `TodoStripView`, so the private `#contentWidth()` that this PR fixes is not exercised.

The new test builds a `TodoStripView` on the existing OpenTUI test renderer, renders a 200-character in-progress todo in a 60-column box, and compares the emitted `TextRenderable`'s plain text against the width the layout actually gave that renderable: collapsed must fill 58 columns (60 - 2 of padding), expanded 54 (60 - 6 of chrome, selection marker included). At 12bb80c9 the collapsed line measures 54 in a 58-column slot and the test fails (`Expected length: 58, Received length: 54`); at this branch's head all 7 cases in the file pass. A third case renders 8- and 2-column boxes so a box narrower than its own chrome neither throws nor paints past the box edge.

The two tautological cases are replaced; the remaining formatter tests still assert distinct behavior. Full `bun test --max-concurrency 1 src` in clients/tui: 482 pass, 1 fail, the pre-existing launcher "current working directory was deleted" sandbox failure that also fails at the branch tip without this change. `pnpm --filter @vibesys/tui check` and `pnpm check:ts` are clean.

vic-lsh:
Rebased this branch onto main (`git rebase --onto origin/main 12bb80c9`) before merging, dropping the commits that belonged to #541 and #557, plus the #558 commits that were already squash-merged into main as 80c38a34. Only the three commits proper to this PR (the fix, the author's tests, and the maintainer's rendering regression test) were merged.

PR #560, opened 2026-09-02
fix(tui): make overlay scroll when content exceeds the box

### LINKED ISSUE #554: TUI: Make the overlay scroll when its content exceeds the box
### Observed behavior

The `/design`, `/help`, and split-fallback overlays render their entire content as a single text child inside a fixed 60%-height box with no scroll region, so on a short terminal the excess is clipped and unreachable. The operator sees only the top of the design log or fallback chart and cannot page down.

### Expected behavior

Overlay content that exceeds the box should scroll (PgUp/PgDn), exactly as the error banner and right pane already do with a scroll box

### Reproduction

1. On a short terminal (for example 60% of 24 rows), run `/design` after several rounds, or open a visualization below the 100-column split width so it falls back to the overlay.
2. Observe the content is cut off with no way to reach the rest.

### Affected subsystem

CLI and packaging

### Environment

### Impact

Blocks normal use

### Relevant logs

```shell
`clients/tui/src/ui/overlay.ts:30-45` fixes `height: '60%'` and `:69-78` adds a single `TextRenderable` directly, unlike `right-pane.ts` and `error-banner.ts` which use a `ScrollBoxRenderable`.
```

#284

### Pre-submission checks

zatchbell1311-wq:
## Problem

The `/help`, `/design`, and split-fallback overlays rendered their whole content as one `TextRenderable` inside a fixed-share box with no scroll region. On a short terminal everything past the box height was clipped and unreachable: the operator saw the top of the content and had no way to page down.

Closes #554

## Solution

Content renders into a `ScrollBoxRenderable` inside `OverlayView`, the pattern `right-pane.ts` and `error-banner.ts` already use, and Page Up/Page Down scroll it while the overlay is open.

Two things the box has to get right for that to be usable:

- **The hint is part of the box, not part of the content.** `Esc to close · PgUp/PgDn: scroll` is created once in the constructor with a reserved row (`height: 1, flexShrink: 0, wrapMode: 'none', truncate: true`) and lives for the life of the view. Rebuilding it per render leaked one hint row per content change; giving it no row of its own let it overpaint the scroll viewport's last row, which hid the final line of content.
- **The box is sized in whole rows and columns.** It kept the same share of the screen it always had (70% wide, 60% tall, at 15%/18%), but as Yoga percentages those are fractional edges, and at 100x24 the rounding put the hint back on the viewport's last row even with the row reserved. `#applyGeometry` recomputes integer width, height, left, and top from the terminal size on every render, so resizes still track.

Scroll-position rule, pinned by a test: **the scroll resets to the top every time an overlay opens.** The view forgets its rendered snapshot when the overlay closes, so reopening the same overlay rebuilds the content and starts at the top instead of restoring the previous position.

On current `main` the overlay block in `keybindings.ts` already sits above `chatPaneFocused` and is fully modal, with Escape handled inside it and a blanket `key.preventDefault(); return;` for everything else. The keybinding change is therefore one `else if` for `pageup`/`pagedown` inside that block; Escape handling is untouched.

### Architecture

Ownership is unchanged: the TUI owns presentation and interaction state, and the overlay is a view over `state.overlay`.

- `ui/overlay.ts` owns the box, its scroll viewport, its hint row, and its geometry, and exposes `scrollBy(delta)`.
- `ui/app.ts` wires the `scrollOverlay` action to `overlay.scrollBy`.
- `ui/keybindings.ts` routes Page Up/Page Down to that action while an overlay is open, ahead of the panes and the hidden command input behind it.

```mermaid
flowchart LR
  key[PgUp/PgDn] --> kb[keybindings: overlay modal block]
  kb --> act[actions.scrollOverlay]
  act --> view[OverlayView.scrollBy]
  view --> scroll[overlay-scroll viewport]
```

Branch note: this branch was rebuilt on current `main` with only this PR's own commit. Its earlier lower commits belonged to #541, #557, #558, and #559, which are now closed or merged.

### Correctness properties

- Content taller than the box is reachable to its last line; the hint never overpaints a content row.
- The box shows exactly one hint row, however many times the content changes while the overlay stays open.
- The scroll position resets to the top on every open, including reopening the same overlay.
- While an overlay is open, Page Up/Page Down scroll the overlay and reach neither the transcript, the docked chat, nor the hidden command input; Escape still closes the overlay and returns to live output.
- The overlay keeps its screen share and z-order (above the chat modal, below the theme picker), now in whole rows and columns.

### Testing

Regression tests in `clients/tui/src/ui/app.test.ts` under `describe('overlay scrolling')`, using `createTestRenderer` at 100x24 with 60 numbered lines of content:

- `reaches the last line of content taller than the box`
- `keeps one hint row however often the content changes`
- `reopens the same overlay at the top`
- `scrolls the overlay rather than the docked chat behind it`

All four fail at `origin/main` and at the pre-fix commit of this branch (which reports two hint rows after the first content change and never reaches the last line); all four pass at this branch head.

| Command | Result |
| --- | --- |
| `bun test --max-concurrency 1 src` (in `clients/tui`) | 494 pass, 1 fail: the pre-existing `launcher > returns the backend argument error for an explicitly empty runs directory`, which cannot run in this sandbox ("The current working directory was deleted") |
| `pnpm check:clients` | pass |
| `pnpm check:ts` | pass (70 files, no fixes applied) |
| `pnpm check:ts-architecture` | pass (no dependency violations) |

A second, docs-only commit updates `.agents/skills/tui-bug-hunt/reference.md`, which still described the overlay footer as `Esc to close` with no scroll viewport. No backend, protocol, or prompt surface is touched.

### DIFF
--- a/.agents/skills/tui-bug-hunt/reference.md
+++ b/.agents/skills/tui-bug-hunt/reference.md
@@ -116,7 +116,9 @@ Chat composer: `Enter` submit, `Shift+Enter` newline.
   `request→Request failed`, `input→Input error`, `run→Run failed`. Equivalent
   reports fold into one banner and bump `count`.
 - **Modal overlay** (`detail`/`help`/`error`): titles Command/Help/Error;
-  footer `Esc to close`. Used for acks, `/help`, perf modal fallback.
+  footer `Esc to close · PgUp/PgDn: scroll` on its own reserved row, over a
+  scroll viewport that PgUp/PgDn page while the overlay is open and that starts
+  at the top on every open. Used for acks, `/help`, perf modal fallback.
 - **argparse diagnostics** print to the terminal for `-h`/`validate` (bypass
   TUI). In server mode a backend argparse failure is surfaced as an error-banner
   diagnostic (stage `argument_parsing`, exit code, hint) rather than exiting.
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -1968,6 +1968,112 @@ describe('OpenTUI presentation', () => {
   });
 });
 
+describe('overlay scrolling', () => {
+  /** Content taller than the overlay box at any terminal size it is used at. */
+  function longOverlayContent(): string {
+    return Array.from(
+      {length: 60},
+      (_, index) => `overlay line ${String(index).padStart(2, '0')}`,
+    ).join('\n');
+  }
+
+  function hintRows(frame: string): number {
+    return frameRows(frame).filter(row => row.includes('PgUp/PgDn: scroll')).length;
+  }
+
+  it('reaches the last line of content taller than the box', async () => {
+    const testRenderer = await createTestRenderer({width: 100, height: 24});
+    const controller = new FakeController({
+      ...initialSessionState(),
+      overlay: {kind: 'help', content: longOverlayContent()},
+    });
+    const app = createOpenTuiApp(testRenderer.renderer, controller);
+    registerCleanup(testRenderer.renderer, app);
+
+    const first = await testRenderer.waitForFrame(value => value.includes('overlay line 00'));
+    expect(first).not.toContain('overlay line 59');
+
+    // No named page key in the mock input; send the raw terminal sequence.
+    for (let index = 0; index < 12; index += 1) testRenderer.mockInput.pressKey('\x1B[6~');
+    const scrolled = await frameAfter(testRenderer);
+
+    // The last row of content is the one the hint used to overpaint.
+    expect(scrolled).toContain('overlay line 59');
+    expect(scrolled).not.toContain('overlay line 00');
+
+    for (let index = 0; index < 12; index += 1) testRenderer.mockInput.pressKey('\x1B[5~');
+    expect(await frameAfter(testRenderer)).toContain('overlay line 00');
+  });
+
+  it('keeps one hint row however often the content changes', async () => {
+    const testRenderer = await createTestRenderer({width: 100, height: 24});
+    const controller = new FakeController({
+      ...initialSessionState(),
+      overlay: {kind: 'detail', content: 'ack 0'},
+    });
+    const app = createOpenTuiApp(testRenderer.renderer, controller);
+    registerCleanup(testRenderer.renderer, app);
+    expect(hintRows(await frameAfter(testRenderer))).toBe(1);
+
+    for (let index = 1; index <= 6; index += 1) {
+      controller.publish({
+        ...controller.state,
+        overlay: {kind: 'detail', content: `ack ${index}`},
+      });
+      const frame = await frameAfter(testRenderer);
+      expect(frame, `after ${index} content changes`).toContain(`ack ${index}`);
+      expect(hintRows(frame), `after ${index} content changes`).toBe(1);
+    }
+  });
+
+  it('reopens the same overlay at the top', async () => {
+    const testRenderer = await createTestRenderer({width: 100, height: 24});
+    const overlay = {kind: 'help' as const, content: longOverlayContent()};
+    const controller = new FakeController({...initialSessionState(), overlay});
+    const app = createOpenTuiApp(testRenderer.renderer, controller);
+    registerCleanup(testRenderer.renderer, app);
+    await testRenderer.waitForFrame(value => value.includes('overlay line 00'));
+
+    for (let index = 0; index < 12; index += 1) testRenderer.mockInput.pressKey('\x1B[6~');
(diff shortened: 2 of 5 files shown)


vic-lsh:
@zatchbell1311-wq a note on branch layout, since it affected this whole series: #541, #557, #558, #559, and #560 were each built on top of the previous one, so every PR's diff included the earlier PRs' commits even though the changes were independent. That meant #558 and #559 had to be rebased before they could be merged on their own, and this PR now conflicts with main because its base commits moved. For unrelated changes, please branch each PR directly from `main` so it can be reviewed and merged in any order. When one PR truly depends on another, base it on that PR's branch and register the series as a native GitHub stack (`gh stack link`), so the dependency is explicit and GitHub retargets it when the base merges. For this PR, a rebase onto current `main` with just the overlay-scroll commit should resolve the conflict.

zatchbell1311-wq:
Rebased onto current main — single commit on top of the current tree. Resolved the keybindings.ts conflict from the old stack base and fixed Biome formatting on the scrollOverlay blocks.

vic-lsh:

vic-lsh:
Thanks @zatchbell1311-wq — the scroll box is the right shape for this, so I finished the branch rather than sending it back. I rebuilt it on current `main` with only this PR's own commit (the lower commits belonged to #541/#557/#558/#559, which are now closed or merged) and added one commit on top of yours.

Findings addressed:

- **[P1] Renderable leak: one hint row per content change.** `#clear()` was narrowed to the scroll box's children, but `'Esc to close · PgUp/PgDn: scroll'` was still appended to `this.output` on every render, so six content changes left six hint rows stacked in the box. The hint is now created once in the constructor and lives for the life of the view; `#clear()` only rebuilds the scrolled content, and `applyTheme` recolors the existing hint.
- **[P2] The last row of content was unreachable.** The hint had no row of its own, so it overpainted the scroll viewport's final row: paging to the bottom of 60 lines showed line 58 and a garbled row where line 59 and the hint were drawn on top of each other. The hint now has `height: 1, flexShrink: 0, wrapMode: 'none', truncate: true`, as in `error-banner.ts`. That alone was not enough: the box was sized with Yoga percentages (`height: '60%'`, `top: '18%'`), and at 100x24 the fractional edges rounded the hint back onto the viewport's last row. The box is now sized and placed in whole rows and columns from the terminal size, recomputed on every render so resizes still track.
- **[P3] Scroll-reset rule.** It reset only when kind or content changed and never cleared `#renderedKind`/`#renderedContent` on the `overlay === null` path, so closing and reopening the same overlay kept the old scroll position. One rule is now pinned and tested: **the scroll position resets to the top every time an overlay opens**, implemented by forgetting the rendered snapshot when the overlay closes.
- **[P3] Scroll box options.** Aligned with `right-pane.ts` and `error-banner.ts`: `stickyScroll: false`, `viewportCulling: true`, `verticalScrollbarOptions: {showArrows: false}`, and content that is `flexShrink: 0` with `wrapMode: 'word'` (without `flexShrink: 0` the content is squeezed into the viewport and there is nothing to scroll).
- **[P3] Keybindings placement and Biome.** On current `main` the overlay block already sits above `chatPaneFocused` and is fully modal, with Escape handled inside it and a blanket `key.preventDefault(); return;`. The change reduces to one `else if` for `pageup`/`pagedown` inside that block, so the chunk that moved Escape handling is gone, and with it the two single-line `if` statements that failed `pnpm check:ts`.

Regression tests (`clients/tui/src/ui/app.test.ts`, `describe('overlay scrolling')`, `createTestRenderer` at 100x24 over 60 numbered lines):

- `reaches the last line of content taller than the box` — first frame shows `overlay line 00` and not `overlay line 59`; after PgDn the last line is on screen; PgUp returns to the top.
- `keeps one hint row however often the content changes` — exactly one hint row across six content changes.
- `reopens the same overlay at the top` — scroll to the bottom, close, reopen the same overlay, back at line 00.
- `scrolls the overlay rather than the docked chat behind it` — with the docked chat focused and a help overlay open, PgDn scrolls the overlay, the chat scroll box stays at 0, and focus stays on the chat.

All four fail at `origin/main` and at your commit (there the hint-row assertion reports 2 rows after the first content change, and line 59 is never reachable), and pass at the branch head. Also run: `bun test --max-concurrency 1 src` in `clients/tui` (494 pass; the one failure is the pre-existing launcher test that cannot run in this sandbox), `pnpm check:clients`, `pnpm check:ts`, `pnpm check:ts-architecture`, all clean.

vic-lsh:
Follow-up: added one more commit that updates `.agents/skills/tui-bug-hunt/reference.md`, which still described the overlay footer as `Esc to close` with no scroll viewport. Docs-only; no behavior change on top of the commit described above.

PR #569, opened 2026-09-03
feat(tui): add a dev-only replay harness driven by recorded runs

### LINKED ISSUE #542: TUI: drive rendering work from recorded runs instead of a live agent run

Nano-AI:
Closes #542.

## Problem

Every TUI rendering change costs a real optimization run to look at. The TUI will not start without a live Python backend (`src/index.ts:13-14`), so checking a layout or theme change means an agent invocation, tokens, and minutes before the first interesting event. Terminal states, typed `tool_result` payloads, and older-schema records cannot be produced on demand at all.

## Solution

`dev/mock-server.ts` speaks the server side of the control protocol and replays a recorded `run-events.jsonl`. The binary under test is the unmodified `dist/index.js`, so no product code changes. Journal lines are already the on-wire `RunEvent`, so replay is re-enveloping, not translation.

Two real captures, scrubbed and committed plain. `bad-cpp-round1` is the only one with a full lifecycle; `queue-rs-payloads` carries a typed payload on all 105 `tool_result` events and ends in `run_failed`, so it exercises the error banner.

```bash
pnpm --dir clients/tui build
clients/tui/dev/mock-ui.sh --speed 0        # interactive
clients/tui/dev/mock-ui.sh --tmux 150x40    # scripted capture
```

The full TUI at 150x40 in the dark theme, replaying a recorded run with no backend running

Nothing ships: `dev/` is outside `rootDir`, absent from `dist` and from `pnpm pack`'s 66-entry tarball, and a `src/` to `dev/` import is rejected by a dependency-cruiser rule.

## Verification

`tests/server/test_tui_dev_harness.py` declares each fixture `CURRENT` or `LEGACY`: every line must validate as a `RunEvent`, and the current-schema one must round-trip byte for byte. The mock's eleven response bodies live in `dev/mock-responses.json`, so the test validates the bytes actually sent, which caught three out of contract.

**Known limit:** CI installs neither node nor bun, so the mock check validates the file, not a live exchange. Verified manually over the socket.

| Check | Result |
| --- | --- |
| `pytest tests/server/test_tui_dev_harness.py` | 17 passed |
| `pnpm --filter @vibesys/tui test` | 553 pass, 0 fail, 24 files |
| `check_format` / `check_lint` / `check_doc_links` | clean |

Rebased onto `main` at 8e09827.


AyanBinRafaih:
@Nano-AI This is great! Could you please fix the findings below:

1. The runner backgrounds the mock server and then execs the TUI. The server only initiates its own shutdown from onLastSubscriberGone, which is never called if the client fails before sending subscribe.  A missing/incompatible runtime, an exception during TUI initialisation, or another pre-subscription exit therefore leaves the server adopted by PID 1 with its socket and log still present. The interactive path also returns the client’s failure code without killing SERVER_PID. Please retain an owner process or another cleanup mechanism until subscription succeeds, and terminate SERVER_PID if the client exits before the server observes a subscriber. This needs a process-level regression test covering a post-bind, pre-subscribe client failure.

2. `query.snapshot` updates the template’s sequence, status, agent, and round, but leaves active_executions as the static empty array from mock-responses.json. This is observably wrong with --bootstrap 20 --paused on bad-cpp-round1: the bootstrap includes invocation_started at sequence 10 and no matching completion, so the event batch reconstructs an active execution. Boot performs the snapshot query and subscription concurrently. If the sequence-20 snapshot arrives after that batch, reduceSnapshot accepts it because its sequence is equal to the folded cursor, then replaces activeExecutions with {} from the empty checkpoint. The resulting paused frame loses the active agent state even though the fixture is stopped inside an invocation. Please derive active_executions from the delivered prefix, or include an equivalent checkpoint on the event batch. It might be a good idea to add an integration test where an equal-sequence snapshot lands after a mid-invocation bootstrap.

Nano-AI:
@AyanBinRafaih Thanks, both findings are fixed.

1. The mock now watches the launched client PID and shuts down if the client exits before subscribing, cleaning up its owned socket and log. A process-level test covers a post-bind, pre-subscribe failure.

2. Snapshots and event batches now use the same active-execution checkpoint derived from delivered events, so an equal-sequence snapshot cannot erase a running execution.

Your `bad-cpp-round1` repro also exposed a real harness gap: its legacy identity and lifecycle events must be canonicalized before replay, as they are on the server read path. The harness now applies the translation needed by both bundled fixtures and checks it against the Python implementation. The active-execution regression is covered with both the legacy fixture and `queue-rs-payloads --bootstrap 40 --paused`.

PR #570, opened 2026-09-03
fix(tui): render transcript tables and fenced code blocks readably

### LINKED ISSUE #562: TUI: render transcript tables and fenced code blocks readably

Nano-AI:
Closes #562.

> Stacked on #569. Base is `adi/tui-mock-harness`, so the diff is this change alone. GitHub retargets to `main` when #569 merges.

## Problem

Tables and fenced code blocks are the two markdown constructs the transcript claims to support and does not render usably.

Tables take the renderer's defaults, so columns stretch to the pane width, a six-column table overflows its card, and the border is a fixed `#888888` that ignores the theme. Fenced blocks arrive with no foreground and no background, visually identical to the prose around them.

Zoomed round-1 transcript: a rendered markdown table 46 columns wide inside a 195-column pane, and two fenced code blocks on the theme's code surface

## Solution

Tables size to content, wrap on word boundaries, fit inside the card, and take the theme's border colour. Fenced blocks draw on the theme's code surface.

Two things worth a close look:

**The override is marked `codeBlockOnly`.** Without it a prose-only message with 100 paragraphs creates 100 native renderables instead of one. `createMarkdownCodeBlockRenderer` could not be the dispatcher, only the source of that marker: it is a per-language table that bails on any fence without a recognized info string, which is most of a transcript.

**The block is styled, not replaced.** Mutating the `CodeRenderable` that `defaultRender()` returns is what keeps `canUpdateInPlace` true, which restores the margin after a fence and preserves streaming, concealment and the tree-sitter client.

## Verification

Tests build a real `MarkdownRenderable` wired like `ConversationView`, with a variant omitting the override as the control. Neutering the marker fails 2 of them; restoring it passes all 17. The `markdown.jsonl` fixture is synthetic because no recorded run contains a table or a fenced block; it includes a table split mid-row across three chunks to exercise streaming.

| Check | Result |
| --- | --- |
| `pnpm --filter @vibesys/tui test` | 573 pass, 0 fail, 25 files |

**Follow-up, not fixed here:** `applyCodeBlockRenderable` resets a code block's colours on in-place update. Unreachable today because the card is rebuilt on every change, but live if the transcript ever updates markdown in place.

Rebased onto `adi/tui-mock-harness` at c71b9c8 (itself on `main` at 8e09827).


vic-lsh:

AyanBinRafaih:
@Nano-AI Could you please resolve this:

1. `renderNode` is only consulted for top-level Markdown blocks. OpenTUI’s createListChildRenderable() handles nested code tokens directly without invoking _renderNode. A fence nested inside a list, such as a list item followed by an indented mystery fence, therefore retains the default CodeRenderable colors and drawUnstyledText: false. When the grammar is unavailable, the nested block remains blank during the fallback or is visually indistinguishable from surrounding prose. This violates the PR’s readability guarantee for unknown and unlabelled languages. `marked` parses this input as a list item containing a code token, and OpenTUI 0.4.3’s nested-child dispatch confirms that the new callback is skipped. Please cover nested fences, including unknown languages, and apply the same background, foreground, and unstyled-text fallback in the nested list path.

Nano-AI:
@AyanBinRafaih You are right about the nested dispatch path. In the current coalesced transcript mode, nested fences are styled as markdown captures rather than separate `CodeRenderable` objects. In `top-level` mode, OpenTUI reaches the unstyled list-child path you identified.

The renderer now handles list tokens too and applies the same code-surface and plain-text fallback styling to nested fenced blocks. Tests cover unknown and unlabelled fences in both the current coalesced mode and the actual list-child path.

I kept `coalesced` as the default because changing it affects table layout, margins, and creates one renderable per top-level block. I also corrected the comment about bundled grammars: OpenTUI ships five; unknown languages still need the fallback.

PR #571, opened 2026-09-03
fix(tui): give the header real content and its own housing

### LINKED ISSUE #517: TUI: replace the raw status/metadata line above the rounds strip with a curated header

Nano-AI:
Closes #517.

## Problem

The header interpolated backend identifiers verbatim, so a live run read:

```
VibeSys · running · implementer · round-1-retry-1-implementer · 2/200k tokens · Preallocated lock-free SPSC ring · r1
```

`round-1-retry-1-implementer` is a loop-internal key, duplicating both the agent kind beside it and the rounds strip below. The token meter rendered 118 percent during a real run, because `input_tokens` is the last call's context pressure while `context_window` is a static lookup. At 100 columns the line clipped mid-title, losing the segment saying what the run was trying.

It also rendered as a bare one-row line while every other region sits in a bordered box, in a single flat accent, on a lighter surface than the panes beneath it.

Five header captures stacked vertically, at 200, 60, 34, 22 and 16 columns, segments dropping whole in priority order

## Solution

**Content.** `phase-label.ts` parses every loop's labels into operator-facing words; an unrecognized label degrades to the agent kind rather than printing an identifier. Segments drop whole in priority order rather than clipping, and width is measured in terminal cells via `text-width.ts`, so CJK and emoji budget correctly. A terminal status always outranks a local override.

**Run state.** `runStateText` defers to main's `runStatusLabel`, so the backend's status is the only source of truth. An earlier revision of this branch carried a client-local `pauseOverride` set from the `/pause` ack, which review correctly flagged for reading `paused` while the run was still going. #578 landed that transition as a real `pausing` status, so the override and its ack handling are gone rather than fixed: two sources of truth for run state is the worse bug. `session-controller.ts` is no longer touched by this branch at all.

**Housing.** The header is a pane: three rows, bezel to bezel, no fill, no blank separator, so its bottom bezel meets the next pane's top one. An earlier attempt used a surface band and a rule, which read as a floating strip and was replaced. The header frame, and then `chat-pane.ts`, set no background at all, so both fell through to the root's `canvas` while every pane beside them sits on `elevatedSurface`; both now name the surface they sit on.

**Hierarchy.** One `TextRenderable` per span, so roles carry tones: brand in accent, run state in a verdict colour and bold, phase and title as content, metadata greyed. Every tone clears its theme's contrast floor in all eight themes.

**Cost:** two extra rows. The housing is bezel, one content row, bezel, so it costs two rows the bare status line did not; on a short terminal that is two rows of transcript. At 100x30 the design pane can no longer show all ten rounds at once: the oldest is on screen at open and the newest is a page away. The pane scrolls, so nothing is unreachable, but that is narrower than what #497's `keeps the newest design round on screen in the pane at 100x30` guaranteed; it passes on the base and fails only after this branch's commits. Rather than raising the terminal size to hide the cost, the test is re-premised at the same 100x30 as `keeps the newest design round reachable in the pane at 100x30`: `Round 1` on screen at open, then three PgDn presses, then `Round 10` and `src/round-10.rs`.

**Known limit:** the state word keeps its text and its bold weight when it truncates, but loses its role colour. `stateColor` matches on the span's rendered text, so `completed` is `success` `#22c55e` at 60 columns while `complet…` falls through to `textPrimary` `#e2e8f0` at 22. Both rows are in the image. Brand and run state do survive to 22 columns as claimed; the state arrives uncoloured. The honest fix is carrying the role's semantic state on the span rather than re-deriving it from rendered text, which is a header refactor and not this PR.

## Verification

Tests cover the backend's `pausing` and `paused` statuses reading correctly through `runStateText`, CJK and emoji budgeting, grapheme-safe truncation, the width ladder at 200/100/60/34/22/16 columns, the tone per role, and contrast for every tone in every theme. Backend labels are unchanged on the wire, so round attribution in `run-map.ts` is unaffected.

One bug found and fixed during the hierarchy work: one renderable per span let OpenTUI shrink every span at once, rendering `V...ys·r...ng` at 24 columns. Fixed with `flexShrink: 0` and hidden unused spans, pinned by a test at that width.

| Check | Result |
| --- | --- |
| `pnpm --filter @vibesys/tui test` | 629 pass, 0 fail, 27 files |

Rebased onto `main` at 8e09827. An earlier rebase, onto 84dfe92, took two upstream changes into this branch. #576 removed `CoreState.terminal` for a typed status behind `hasRunEnded`, so a follow-up commit migrates `runStateText` and drops the tests' `terminal: true` literals. #578 made the run lifecycle a state machine that publishes every transition, which is what removed the pause override above.

One bug came out of that reconciliation: `runStatusLabel` renders `pausing` as `pausing…`, while `stateColor` matched the bare literal `'pausing'` against rendered span text. Deferring to the label without touching `stateColor` would have dropped a pause out of `theme.warning` into body text with no compile error and no failing test. `stateColor` now compares against `runStatusLabel('pausing')`, so the two cannot drift.

Note on structure: commits 1 to 5 do not compile against the new base, since commit 6 is the `hasRunEnded` migration. The repo squash-merges, so this does not reach `main`; say the word if you want it folded for bisectability.

### DIFF
--- a/clients/tui/dev/fixtures/markdown.experiments.json
+++ b/clients/tui/dev/fixtures/markdown.experiments.json
@@ -1,0 +1,21 @@
+[
+  {
+    "hypothesis_id": "h1",
+    "title": "Preallocated lock-free SPSC ring replaces the Mutex<VecDeque> hot path",
+    "claim": "Removing the mutex from the producer/consumer path removes the dominant cost.",
+    "action": "Replace the queue with a bounded SPSC ring and keep the C ABI identical.",
+    "first_round": 1,
+    "last_round": 1,
+    "rounds": [{"round": 1, "passed": true, "reviewed": true, "hypothesis_outcome": "proven"}],
+    "resolved_outcome": "proven",
+    "judge_verdict": "accepted",
+    "perf_metric": 301.4,
+    "perf_unit": "ms",
+    "perf_metric_name": "total_ms",
+    "perf_direction": "min",
+    "perf_baseline_value": 412.7,
+    "perf_delta_pct": -27.0,
+    "kept": true,
+    "active": false
+  }
+]
--- a/clients/tui/dev/mock-server.ts
+++ b/clients/tui/dev/mock-server.ts
@@ -266,6 +266,24 @@ function activeExecutionsFrom(delivered: RunEventRecord[]): ActiveExecutionCheck
   return [...active.values()];
 }
 
+/**
+ * Hypotheses for the experiment log, from `<fixture>.experiments.json` beside
+ * the fixture when one exists.
+ *
+ * Recorded journals hold events, not the projected experiment table, and the
+ * header and log render the hypothesis title rather than anything in the event
+ * stream. A sidecar keeps that title pinnable without inventing event types the
+ * backend never emits.
+ */
+function loadExperiments(fixturePath: string): unknown[] {
+  const sidecar = `${fixturePath.replace(/\.gz$/, '').replace(/\.jsonl$/, '')}.experiments.json`;
+  try {
+    return JSON.parse(readFileSync(sidecar, 'utf8')) as unknown[];
+  } catch {
+    return [];
+  }
+}
+
 /** Milliseconds to wait before `next`, from the recorded timestamps. */
 function gapMs(previous: RunEventRecord, next: RunEventRecord, options: Options): number {
   if (options.speed === 0) return 0;
@@ -470,6 +488,7 @@ function main(): void {
   // out: every read path a client can reach translates legacy lifecycle events,
   // so the replay owes the TUI the translated stream, not the recorded one.
   const events = canonicalJournalEvents(readJournalRecords(options.fixture));
+  const experiments = loadExperiments(options.fixture);
   const replay = new Replay(events, options);
   process.stderr.write(
     `mock: ${String(events.length)} events from ${options.fixture}\n` +
@@ -541,7 +560,10 @@ function main(): void {
           return;
         }
         case 'query.experiments': {
-          writeLine(socket, ok(id, responseBody(type)));
+          writeLine(
+            socket,
+            ok(id, withValues(responseBody(type), {experiments, experiments_ready: true})),
+          );
           return;
         }
         case 'query.events': {
--- a/clients/tui/src/session-model.test.ts
+++ b/clients/tui/src/session-model.test.ts
@@ -1,6 +1,6 @@
 import {describe, expect, it, test} from 'bun:test';
 import type {RunEvent, RunStatus} from '@vibesys/backend-client';
-import {type CoreRunStatus, hasRunEnded} from '@vibesys/core-state';
+import {hasRunEnded} from '@vibesys/core-state';
 import type {SessionState} from './session-model.js';
 import {
   applyActiveExecutionCheckpoint,
@@ -32,6 +32,7 @@ import {
   openPane,
   openThemePicker,
   reportError,
+  runStatusLabel,
   selectExperimentActivity,
   selectNextAgent,
   selectNextRound,
@@ -41,7 +42,6 @@ import {
   setPaneContent,
   setTheme,
   showDetail,
-  statusText,
   togglePaneZoom,
   toggleTodos,
   unownedExperimentRounds,
@@ -50,6 +50,7 @@ import {
   visiblePhases,
   visibleTodos,
 } from './session-model.js';
+import {runStateText, usageText} from './ui/header.js';
 
 describe('event batch projection', () => {
   it('keeps the existing banner while resumed history ends in a running session', () => {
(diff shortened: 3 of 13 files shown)


vic-lsh:

AyanBinRafaih:
@Nano-AI A few findings:

1. `selectedAgentKind` is interpolated directly as selected ${state.selectedAgentKind}. This is not hypothetical: the plain loop seeds perf_eval as a selectable phase in clients/core-state/src/run-map.ts, and selection tores the raw phase kind. Selecting that phase therefore renders selected perf_eval in the curated header, reintroducing the backend identifier that this change is intended to replace with operator-facing language. Please route this segment through the same canonical display mapping used by the phase description, or suppress the segment when there is no user-facing label. Add a test selecting the real perf_eval phase.

2. The header frame now paints theme.elevatedSurface, but headerSpanStyle and the all-theme tests calculate contrast against theme.canvas. Those colors differ in several built-in themes. Using the repository’s own contrastRatio calculation against the actual header background: solarized-dark error is about 3.71:1, solarized-light accent is about 3.05:1, with success and warning about 3.10:1 and catppuccin-latte warning is about 2.62:1 and accent about 3.74:1. The rendered brand and state tones can therefore fall below the claimed 4.5:1 text floor while the canvas-based test passes. Please derive the header colors against elevatedSurface, or render the header on canvas, and make the all-theme assertion use the frame’s actual background.

Nano-AI:
@AyanBinRafaih The first finding is fixed: `perf_eval` now maps to `measuring`, selection uses the same mapping (`filtered to measuring`), and unknown kinds are omitted. Tests cover the real `perf_eval` phase from the run-map projection.

The contrast coupling is fixed too. `headerBackground(theme)` now supplies both the frame fill and the contrast basis, and header colors are adjusted against that surface. Tests verify rendered cells in all eight themes and simulate the surface inversion proposed in #574.

One correction: the reported sub-4.5 ratios came from raw `ThemeSpec` literals. The resolved built-in theme colors already cleared the floor, so no current color value changed. The structural fix prevents the renderer and contrast check from drifting later.

PR #572, opened 2026-09-03
fix(tui): make pane focus unambiguous and legible in every theme

### LINKED ISSUE #561: TUI: pane focus is hard to distinguish in five built-in themes

Nano-AI:
Closes #561. Regression of #433, which PR #434 recoloured rather than resolved.

## Problem

Two problems with one cause: nothing owned what focus means.

**Which surface is focused had no single authority.** `experiment-log` and the transcript asked `focusedPane(state)`, `agent-map` asked `state.roundFocus`, and `command-input` was driven by a pane-focus expression despite the command box being shared by every pane in its column. Two boxes lit at once.

**How focus is shown was carried by colour alone.** Across the eight built-in themes, focused-versus-resting border contrast is 4.33 at best and under 1.5 in five. `high-contrast-light` was inverted: focus at 11.22 against the canvas was *quieter* than the 12.63 resting border, so focusing a pane made it recede.

The chat pane focused in the dark theme: its frame in borderFocus with a marker in the title gutter, the Experiments pane resting, the Command box grey. Captured before the composer fix below, so the nested Message box still carries a marker here

## Solution

Every surface compares its own `PaneId` against `focusedPane`, documented in `focus.ts` as the sole authority. Focus then shows on two channels, colour last:

- A marker in a **reserved** title gutter, so focus swaps a glyph rather than inserting one and the label never shifts.
- `high-contrast-light` focus becomes navy `#000080` (16.01), removing the inversion.

A heavier frame was tried as a third channel and dropped. Unicode has no heavy rounded corner: the arcs at U+256D-2570 are light-only and the heavier weights are square, so the swap traded rounded corners for square ones. That reads as the pane becoming a different kind of object rather than a louder one. Focus never changes a pane's shape.

No other palette changes: every resting border already sits below the 3:1 non-text floor, so there is no headroom to quiet one further. The high-contrast palettes are deliberately small, so no retune fixes this, and colour cannot be the primary channel anyway: WCAG 1.4.1 and 2.4.7.

The command box calls the same helpers but only ever in the resting form. It takes the title so its label aligns with the panes above, never the marker.

**Review round.** The three findings were one bug: a surface wearing focus it did not own, or owning focus it did not show. One rule settles them. The treatment belongs to the frame of the pane `focusedPane` names, a box inside such a pane never repeats it, and a surface that takes the keys is a pane whatever its shape. The expanded todo list already receives Up and Down from `keybindings`, so `todos` becomes a `PaneId`, `focusedPane` names it, and the pane it opened over goes back to rest. Below the split width the fallback overlay takes the pane it stands in for instead of titling itself Command.

**Nested boxes.** The invariant now holds per box, not only per pane. `chat-composer.ts` takes the resting frame, so a focused chat pane carries `borderFocus` and the marker on its own frame alone, and which composer holds the cursor is said by the cursor and by the hint line under the box. "Exactly one box carries cyan" is true, including for the chat pane the screenshot shows.

**Conventions doc.** This branch also adds `docs/contributing/tui-conventions.md`. Review caught it presenting an unimplemented keymap as current policy and describing a border-weight focus channel that does not exist, so it is now split into **Rules in force**, each checkable against this tree, and **Proposed: movement and naming**, explicitly not implemented. Auditing the rest of the file found four more claims that were false: arrows and Tab were inverted (today `←`/`→` move between regions and `↑`/`↓`/`Tab` move within one), `/` was described as search when it is the command prefix and no search exists, and `?`, `q` and `F1` were documented with no handlers behind them. The Tab-based keymap it argues for lives on the unmerged `adi/tui-keymap-defaults`; `Ctrl+W` is what this tree binds.

A third commit corrects `focus.ts` itself. Its module docstring still claimed three channels including border style, and `paneBorderStyle` still explained the border as "a style swap", while the function ignores its argument and returns the rounded constant. `PANE_BORDER` already recorded the real history; the docstrings above it were where the doc's error came from. Comments only.

## Verification

`check_doc_links.py` passes on 326 files. Note it does not validate same-page anchors, tested by pointing a link at a heading that does not exist, so the two intra-page anchors were checked by hand. `focus.test.ts` pins the invariant **for every built-in theme**: a focused pane differs from a resting one by at least one non-colour channel, and no theme makes a focused border quieter than a resting one. Reverting the source while keeping the tests fails exactly the new assertions. The review round adds five renderer tests, all failing on the pre-fix tree: a no-op notification with the experiment-log detail open, the expanded todo list under each possible parent focus and against zoom, the fallback overlay below `splitFits`, and a focused chat pane with its composer at rest.

| Check | Result |
| --- | --- |
| `pnpm --filter @vibesys/tui test` | 673 pass, 0 fail, 28 files |

Rebased onto `main` at 84dfe92. That commit (#560) rewrote this same `overlay.ts` to draw its content in a `ScrollBoxRenderable` behind a reserved hint row, which conflicted with the narrow-width fallback above. Resolved by keeping #560's scroll architecture and routing the fallback through it: `#renderPane` calls `#applyGeometry()` and feeds the same `#body()` helper as an ordinary overlay, so the pane inherits scrolling and the `Esc to close / PgUp/PgDn` hint instead of the two literal lines it used to append. Both sides' tests are present and pass, including #560's `scrolls the overlay rather than the docked chat behind it`. The branch has since been rebased again, onto `main` at 8e09827, and that resolution carries forward unchanged.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -149,7 +149,8 @@ stacking another surface on top.
 `Ctrl+W` moves focus one column to the right and wraps, through whichever
 columns are actually on screen. Every pane reads the same authoritative focus
 state: the focused pane carries the theme's focus border and a `▸` in its title,
-while every other pane uses the neutral border. Page Up and Page Down scroll
+while every other pane uses the neutral border. The treatment belongs to a pane
+frame, so a box inside one, the chat's composer for instance, never repeats it. Page Up and Page Down scroll
 whichever pane has focus, and Escape on the right pane closes it and restores
 the full-width view. Chat and transcript state survive the pane closing.
 
@@ -158,13 +159,16 @@ content row; pressing `F4` again restores the previous split and focus. It does
 not create a second pane or move its data, so hypothesis and agent selection,
 chat drafts, and pane scroll positions remain in place. The shortcut works for
 the hypothesis list, agent graph, transcript, experiment chat, and performance
-pane. Modal dialogs keep `F4` to themselves until they close.
+pane. It does nothing over an expanded todo list, which is as tall as its own
+contents and has no use for the row. Modal dialogs keep `F4` to themselves until
+they close.
 
 Pane widths are computed from the terminal, so a wide terminal gives the
 visualization real room while the left pane keeps a readable floor. Below 100
 columns there is not enough width for both, and visualizations fall back to the
-modal they used before panes existed. The layout re-flows on resize in either
-direction.
+modal they used before panes existed. That modal is the same surface as the
+pane, so it keeps the pane's title and its focus marker rather than reading as a
+generic dialog. The layout re-flows on resize in either direction.
 
 `/help`, `/theme`, and errors stay modal.
 
@@ -193,8 +197,8 @@ Inside a hypothesis the footer shows keyboard navigation. `←` and `→` step
 focus across the rounds rail, the agents graph, and the transcript, `↑` and `↓`
 move within whichever holds it, `[` and `]` select rounds from anywhere, Tab and
 Shift+Tab select agents, Page Up/Page Down scroll the transcript, and F2 (or
-Ctrl+T) expands the todo box, which then takes the arrow keys until Escape closes
-it.
+Ctrl+T) expands the todo box, which then takes the arrow keys, and the focus
+marker with them, until Escape closes it.
 F3 (or Ctrl+P) expands the latest prompt in the current selection. Function keys
 are offered alongside the Control chords because a terminal is free to keep a
 Control chord for itself, and on macOS several do.
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -108,12 +108,15 @@ list rises out of the command input rather than across the chat.
 
 The cursor starts in the command input, and `Ctrl+W` moves both it and the pane
 keys to the chat and back; the chat's instruction line says so (`Ctrl+W to type
-here`) until it holds them, and the focused input carries the focus border, so
-where a keystroke lands is never a guess. Only the chat composer accepts
-ordinary questions; the command input accepts slash commands. Page Up and Page Down scroll the focused pane, and
-Escape hands the keys back to the table. Arrows, Enter, and the rest of the
-table's keys are unaffected by the dock. A slash command typed into the chat
-input runs through the same command path as the command box.
+here`) until it holds them, and the chat pane carries the focus border while it
+does, so where a keystroke lands is never a guess. The command box is shared by
+every pane in its column rather than being one of them, so it keeps the neutral
+border in every theme and never competes with the pane that holds the keys.
+Only the chat composer accepts ordinary questions; the command input accepts
+slash commands. Page Up and Page Down scroll the focused pane, and Escape hands
+the keys back to the table. Arrows, Enter, and the rest of the table's keys are
+unaffected by the dock. A slash command typed into the chat input runs through
+the same command path as the command box.
 
 `/chat` leaves the command surface on this view, since the chat is already
 beside the table: it is absent from `/help` and from the completions, though it
--- a/clients/tui/src/session-model.ts
+++ b/clients/tui/src/session-model.ts
@@ -1286,7 +1286,12 @@ export function focusPane(state: SessionState, focus: PaneFocus): SessionState {
   return {...state, layout: {...state.layout, focus}};
 }
 
-/** The semantic pane currently receiving pane navigation keys. */
+/**
+ * The semantic pane currently receiving pane navigation keys, and the single
+ * authority for the focus treatment: every pane view compares its own `PaneId`
+ * against this, so at most one of them can be lit at a time. Surfaces that are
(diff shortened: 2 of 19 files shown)


vic-lsh:

AyanBinRafaih:
@Nano-AI Could you please fix the following:

1. `render()` now applies the generic Experiments title before the identity/width cache check. Detail rendering later replaces that title with Hypothesis <id>. If the detail view receives a same-state notification, the cache returns immediately after resetting the title to Experiments, while the hypothesis detail content remains visible. This is reachable because focusPane('left') can return the existing state, but SessionController.#setState still notifies every listener.
The frame consequently changes from, for example, Hypothesis H-07 to Experiments after clicking an already-focused log area. Please apply the focus chrome without overwriting the selected title, or preserve/reapply the detail title before the cache return. Add a renderer test for a no-op state notification while detail is open.

2. The expanded todo list unconditionally receives `paneTitle(..., true)` and borderFocus. Expanding it does not change roundFocus, while keybindings.ts routes Up/Down to selectNextTodo() whenever todosExpanded is true. If the transcript is focused and the operator presses Ctrl+T, the transcript keeps its focus marker and border while the nested todo list also displays the same focused marker and border. Both surfaces claim focus, although only the todo list receives the arrow keys. Either model the expanded todo list as the semantic focus and clear the parent treatment, or keep the nested box neutral and use a distinct selection indicator. Please add a renderer test with the list expanded under each possible parent focus.

3. When `splitFits()` is false, app.ts hides RightPaneView and renders the visualization content through OverlayView. focusedPane(state) can still be performance, but this code applies the marker and focused border only to the hidden RightPaneView. Running /perf at the narrow fallback width therefore shows an overlay titled Command, with the generic info border and no focus marker. The visible keyboard target has none of the non-color focus treatment this PR introduces. Please render the fallback as the focused performance surface, including its actual title, or give that modal an explicit equivalent focus treatment. Add a renderer test at a width below splitFits() that opens the performance pane and checks the visible overlay title and focus marker.

Nano-AI:
@AyanBinRafaih Thanks, all three findings are valid. The detail title can reset after a no-op notification, the expanded todo list can claim focus alongside its parent, and the narrow visualization fallback styles the hidden pane instead of the visible modal. I’ll fix each visible surface and add renderer regressions for the reported states.

Nano-AI:
@AyanBinRafaih Thanks, all three are fixed in bf0a070. They turned out to be one bug: a surface wearing focus it did not own, or owning focus it did not show. The rule now is that the treatment belongs to the frame of the pane `focusedPane` names, a box inside such a pane never repeats it, and a surface that takes the keys is a pane whatever its shape.

1. The experiment-log title is derived from the same state as the body, in the one place that runs on every notification, so the cache return can no longer leave `Experiments` over a hypothesis detail. Covered by `keeps the hypothesis title through a no-op state notification`.

2. `todos` is now a `PaneId`. `focusedPane` names the expanded list, the pane it opened over goes back to rest, and `keybindings` and the chrome read the same predicate, so they cannot disagree. Covered by `moves the focus treatment onto the expanded todo list, under every parent focus` and `keeps zoom and the expanded todo list from claiming the row together`.

3. The narrow fallback overlay takes the pane it is standing in for and asks `focusedPane` the same question every other pane view does, so the visible keyboard target carries the real title and the marker instead of `Command`. Covered by `shows the narrow fallback as the focused visualization, not a Command box`.

The nested Message composer the PR body listed as a known limit is the same class as finding 2 and is fixed with it: the composer takes the resting frame and the chat pane frame carries the marker alone, so which composer holds the cursor is said by the cursor and the hint line. Covered by `keeps the composer inside a focused chat pane at rest`.

All five tests fail on the pre-fix tree. The PR body is updated: the known limit is gone and the verification row is now `bun test --max-concurrency 1 clients/tui/src`, 611 pass, 0 fail. The screenshot predates this commit, so the composer still shows a marker there; its caption says so.

PR #592, opened 2026-09-03
TUI: Sync chat trajectory evidence incrementally and show the wait

vic-lsh:
## Problem

Two independent costs on every experiment-chat question, both found and first
fixed by @AyanBinRafaih in #484:

1. **The trajectory snapshot was rebuilt from scratch per question.**
   `TrajectoryEvidence.refresh` `rmtree`d the whole `trajectory/` directory and
   re-copied every state file and every run log before the agent was invoked.
   The logs grow for the length of the run, so the cost of asking a question
   grows with the run, and it is paid while the operator is already waiting.

2. **Nothing on screen said a question was in flight.** The composer's title
   read a static ` Message · awaiting agent ` for however long the agent took,
   so a 30-second answer and a stuck one looked identical.

#484 is being split: its provider-session-reuse half is being rebuilt on the
`AgentSessionKey` / `SessionStore` contract that landed with #523, and depends
on nothing here. This PR carries the two halves that do not touch sessions, so
they can land on their own.

### Incremental evidence sync

`refresh` no longer deletes the snapshot. It writes the state files whose bytes
differ, copies the log files whose size or mtime differs, and then prunes every
entry the pass did not produce, so it converges to exactly the tree the old
full rebuild produced. `shutil.copy2` replaces `copyfile` so the destination
carries the source mtime and the next sync has that comparison to make. The
sync lock and the symlink guards are unchanged.

The size-and-mtime skip rests on run logs being append-only, and the docstring
now says so, along with how it degrades: on a filesystem with coarse
timestamps, a same-size rewrite landing inside the mtime granularity would be
missed until the next write moves either. Nothing under `log_dir` rewrites in
place today, and the cost of being wrong is one stale evidence file, not a
wrong answer about run state.

### Composer spinner

The composer title carries the braille spinner and the elapsed wait while a
question is in flight. Both are reused rather than re-implemented: the frames
and cadence are now exported from `activity-bar.ts` (the same animation the
activity bar runs), and the wait is formatted by `elapsedLabel` from
`previews.ts`.

The timer is driven by `ChatComposerView.syncPending(pending, onScreen)`, which
both chat surfaces call from `render` **before** their visibility gate, the way
`ActivityBarView.render(state, visible)` syncs its timer whether or not the bar
is shown. That is the correctness point: `activate()` runs only while the
surface is visible, so a question that completes while the dock is hidden (the
terminal narrowed, the modal closed) would otherwise leave the interval running
for the life of the process and leave the elapsed epoch pinned to a wait that
already ended. The epoch moves only when a question starts, not when a surface
returns, so hiding and reshowing mid-question keeps the wait the operator has
actually been watching. `destroy()` clears the pending state along with the
timer, so a composer activated again after teardown starts its next question
from a fresh epoch instead of a dead one.

### Architecture

Ownership is unchanged on both sides. `TrajectoryEvidence` still owns the
snapshot directory and is still the only writer under the sync lock; the
composer still owns its own title, and the surfaces still own visibility.

```mermaid
flowchart LR
  Q["chat question"] --> R["TrajectoryEvidence.refresh"]
  R --> W["_write_snapshot<br/>skip byte-equal state files"]
  R --> C["_copy_files<br/>skip size+mtime-equal logs"]
  W --> P["_prune_stale<br/>drop what this pass did not produce"]
  C --> P

  Pane["ChatPaneView.render(state, visible)"] --> S["composer.syncPending(pending, onScreen)"]
  Modal["ChatOverlayView.render(state)"] --> S
  S --> T["spinner timer<br/>runs only while pending and on screen"]
  S --> E["elapsed epoch<br/>moves only on a new question"]
  App["app.destroy()"] --> D["pane/overlay destroy()"] --> S
```

### Correctness properties

- The incremental sync converges to the same tree as the old full rewrite:
  unchanged files are left alone, changed files are rewritten through the same
  temp-file-and-replace, and stale files and the directories they emptied are
  removed.
- Writes stay atomic from a reader's point of view: every file still lands via
  `.<name>.tmp` then `replace`.
- The sync lock, the symlink guards, and the suffix allowlist are unchanged, so
  the set of files chat can see is unchanged.
- The spinner timer runs only while a question is pending *and* its surface is
  on screen. An answer arriving while the surface is hidden stops it.
- The elapsed wait is per question, not per render: a hidden-then-reshown
  surface reports the original wait, and a new question reports its own.
- `destroy()` leaves the composer able to animate again if it is activated
  afterwards.
- No third copy of the spinner frames or the elapsed formatter: both are
  imported from their existing owners.

### Testing

```
uv run pytest tests/server -q --no-cov            -> 236 passed
./scripts/check_format.sh                          -> 481 files already formatted
./scripts/check_lint.sh                            -> All checks passed
./scripts/check_types.sh                           -> All checks passed
clients/tui: bun test --max-concurrency 1 src      -> 500 passed, 0 failed (21 files)
pnpm check:clients                                 -> clean
pnpm check:ts                                       -> 71 files checked, no fixes
pnpm check:ts-architecture                          -> no dependency violations
CI=true pnpm install --offline --frozen-lockfile    -> lockfile up to date
```

New coverage:

- `tests/server/test_chat_evidence.py`: an unchanged log is not rewritten in
  place (same inode and ctime) while a file the sync did not produce is pruned;
  an appended log is re-copied and carries the source mtime forward; a
  directory the run no longer produces is removed.
- `clients/tui/src/ui/chat-composer.test.ts`: the pure title formatter, plus
  three `createTestRenderer` tests driving a real `ChatPaneView` through
  pending → hidden → complete. Confirmed against the pre-fix wiring: syncing
  the timer after the visibility gate fails the first two ("Received:
  ' Message · ⠋ 0s '" instead of ' Message ', and "' Message · ⠋ 40s '"
  instead of a fresh epoch), and a `destroy()` that clears only the interval
  fails the third.

Not covered: a live run measuring the sync's saved wall time per question. The
manual timing @AyanBinRafaih reported on #484 is the evidence there.

### DIFF
--- a/clients/tui/src/ui/activity-bar.ts
+++ b/clients/tui/src/ui/activity-bar.ts
@@ -5,8 +5,9 @@ import {visibleActiveExecutions} from '../session-model.js';
 import {agentRuntimeLabel} from './agent-runtime-label.js';
 import type {Theme} from './theme.js';
 
-const SPINNER_FRAMES = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
-const SPINNER_INTERVAL_MS = 120;
+/** The one braille spinner every in-flight indicator animates. */
+export const SPINNER_FRAMES = ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'];
+export const SPINNER_INTERVAL_MS = 120;
 /** Status for executions visible in the selected agent conversation. */
 export class ActivityBarView {
   readonly output: TextRenderable;
--- a/clients/tui/src/ui/app.ts
+++ b/clients/tui/src/ui/app.ts
@@ -471,6 +471,8 @@ export function createOpenTuiApp(
       roundStrip.destroy();
       agentMap.destroy();
       experimentLog.destroy();
+      chat.destroy();
+      chatPane.destroy();
       root.destroyRecursively();
       markdownStyle.destroy();
     },
--- a/clients/tui/src/ui/chat-composer.test.ts
+++ b/clients/tui/src/ui/chat-composer.test.ts
@@ -1,0 +1,129 @@
+import {afterEach, describe, expect, it, setSystemTime} from 'bun:test';
+import {BoxRenderable, type Renderable} from '@opentui/core';
+import {createTestRenderer, type TestRendererSetup} from '@opentui/core/testing';
+import type {SessionController} from '../session-controller.js';
+import {initialSessionState, type SessionState} from '../session-model.js';
+import {createChatDraft, pendingComposerTitle} from './chat-composer.js';
+import {ChatPaneView} from './chat-pane.js';
+import {createMarkdownStyle} from './styles.js';
+import {resolveTheme} from './theme.js';
+
+describe('pending composer title', () => {
+  it('shows a spinner frame and the elapsed wait', () => {
+    expect(pendingComposerTitle(0, 0)).toBe(' Message · ⠋ 0s ');
+    expect(pendingComposerTitle(1, 5_000)).toBe(' Message · ⠙ 5s ');
+  });
+
+  it('wraps the frame index so a long wait keeps animating', () => {
+    expect(pendingComposerTitle(10, 1_000)).toBe(pendingComposerTitle(0, 1_000));
+    expect(pendingComposerTitle(23, 1_000)).toBe(pendingComposerTitle(3, 1_000));
+  });
+});
+
+/**
+ * The spinner's state lives in the composer, but only the surfaces around it
+ * know whether they are on screen, so these drive a real `ChatPaneView` and
+ * read the title its box ended up with. A question that completes while the
+ * dock is hidden is the case `activate` alone cannot see: it does not run
+ * while hidden, so nothing stops the timer or moves the elapsed epoch.
+ */
+describe('composer spinner across a hidden surface', () => {
+  const cleanup: Array<() => void> = [];
+  const EPOCH = Date.UTC(2026, 8, 3, 12, 0, 0);
+
+  afterEach(() => {
+    for (const destroy of cleanup.splice(0).reverse()) destroy();
+    setSystemTime();
+  });
+
+  /** The composer only calls back on input, which none of these tests send. */
+  const controller = {
+    focusPane: () => {},
+    submitChat: () => {},
+  } as unknown as SessionController;
+
+  function stateWith(pending: boolean): SessionState {
+    return {...initialSessionState(), chatPending: pending};
+  }
+
+  function composerTitle(node: Renderable): string | null {
+    for (const child of node.getChildren()) {
+      if (child instanceof BoxRenderable && child.id.endsWith('-composer-box')) {
+        return child.title ?? null;
+      }
+      const found = composerTitle(child);
+      if (found !== null) return found;
+    }
+    return null;
+  }
+
+  interface Dock {
+    render: (pending: boolean, visible: boolean) => void;
+    title: () => string | null;
+    destroy: () => void;
+  }
+
+  async function dockedChat(): Promise<Dock> {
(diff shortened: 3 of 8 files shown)


vic-lsh:
This is one half of #484, split out so it can land on its own. @AyanBinRafaih wrote both halves; this one is his incremental trajectory sync and his composer spinner, with the review findings applied. The other half — chat provider-session reuse — stays on #484 and is being rebuilt there on the `AgentSessionKey` / `SessionStore` / `RESET_REQUIRED` contract that landed with #523, so it now depends on that PR and nothing here depends on it. Splitting means the sync and the spinner are not held behind a driver-contract rework, and #484's review can be about sessions only.

What changed relative to #484's version of these files, in the review's terms.

**[P1] The spinner interval leaked, and reported a wait that had already ended.** `#setPending(false)` was reachable only from `activate()`, and `activate()` runs only while the surface is visible: `ChatPaneView.render` and `ChatOverlayView.render` both return early when hidden. So a question that completed while the dock was hidden (terminal narrowed to the modal, chat closed) left the 120ms interval running for the life of the process, and left `#pending` true — so the next question found no rising edge, kept the dead epoch, and reported the earlier question's elapsed time. Now `syncPending(pending, onScreen)` is called from both surfaces *before* their visibility gate, following `ActivityBarView.render(state, visible)`, which syncs its timer whether or not the bar is shown. The timer runs only while pending and on screen; the epoch moves only when a question starts, so hiding and reshowing mid-question keeps the wait the operator has actually been watching.

**[P1] `destroy()` left the composer unable to animate again.** It cleared the interval but not `#pending`, so a later `activate` saw `pending && !this.#pending` as false and never restarted the timer. It now clears the pending state with the timer.

**[P2] Third copies of the spinner and the elapsed formatter.** `SPINNER_FRAMES` / `SPINNER_INTERVAL_MS` are now exported from `activity-bar.ts` and imported, and the wait is formatted by `elapsedLabel` from `previews.ts` rather than a second implementation of the same `Ns` / `Nm Ns` rule.

**[P2] Test that fails on the old code.** `chat-composer.test.ts` keeps the pure title assertions and adds three `createTestRenderer` tests driving a real `ChatPaneView` through pending → hidden → complete. Checked against the pre-fix wiring: syncing after the visibility gate fails two of them (' Message · ⠋ 0s ' where ' Message ' is expected, and ' Message · ⠋ 40s ' where a fresh epoch is expected), and the old `destroy()` fails the third.

**[P3] The append-only assumption was unstated.** `_copy_files` skips a log whose size and mtime match its source, which is only sound because run logs are appended to and never rewritten. The docstring now says that, says `copy2` is what makes the comparison possible on the next pass, and says how it degrades: on a filesystem with coarse timestamps a same-size rewrite landing inside the mtime granularity would be missed until the next write moves either. The cost of being wrong is one stale evidence file, never a wrong answer about run state.

Also renamed the sync's `fresh` set to `current` and added a comment at the call site stating the convergence property (same tree as the old full rebuild), plus two more evidence tests: an appended log is re-copied and carries the source mtime forward, and a directory the run no longer produces is pruned.

Checks on this head: `tests/server` 236 passed, `bun test src` 500 passed, format / lint / types / `check:clients` / `check:ts` / `check:ts-architecture` all clean.

vic-lsh:
Note on this PR's red check: the four `Build <platform>` jobs fail on a break already on `main`, not on anything here. #523 added `implementer_driver=ctx.agent_client.driver_name` at `src/vibesys/loops/agent/loop.py:3240`, and `StubAgentClient` (a standalone class, `cast` to `AgentClient` in the factory) has no `driver_name`, so the release job's `--stub-agent` smoke run dies with `AttributeError: 'StubAgentClient' object has no attribute 'driver_name'` right after the first judge. It reproduces with `vibesys --headless --stub-agent --agent-backend cli --exp-name smoke --max-rounds 1 --no-skills --backend cpu --profiler none` on a clean `main`, and it needs its own fix rather than one folded into this PR. Every check that covers this diff — python-checks, test (3.12), tui, typecheck, sandbox-linux/macos, queue-native, ci-budget — passes.

PR #598, opened 2026-09-04
TUI: advertise expected agent roles in run_started

### LINKED ISSUE #584: TUI: Source the expected per-loop agent roles from the backend contract
### Workstream

Core/runtime

### Problem

The client hardcodes, per outer loop, which agent roles to seed as pending placeholders in the agent graph, duplicating backend knowledge with no shared source or assertion. The server already emits kinds the client list does not know about (for example a `benchmark` kind), and an unknown outer loop returns an empty list with no seeding at all, so any server-side rename or new role silently drops the pending placeholder. Evidence: `clients/core-state/src/run-map.ts:245-250` hardcodes `agent → [orchestrator, implementer, judge, profiler]`, `plain → [implementer, judge, perf_eval]`, `evolve → [implementer, judge, profiler]`, while `src/vibesys/loops/__init__.py:10` already notes `plain` does not run a profiler today and the loops emit a wider kind set.

### Desired outcome

The backend advertises the expected roles per outer loop (for example in `run_started` data) and the client folds that, or the mapping is centralized in the generated protocol, so client and server share one definition and new roles appear automatically.

### Acceptance criteria

- [ ] Expected roles per outer loop come from the backend contract, not a client literal.
- [ ] A newly added backend role shows a pending placeholder without a client edit.
- [ ] An unknown outer loop degrades gracefully with a visible indicator, not a silent empty seed.

### Scope and non-goals

In scope: 
- the role source of truth and the seeding path. 

Not in scope: 
- redesigning the agent-graph layout.

Prefer extending the protocol payload over a second hand-maintained map; keep the change backward compatible for existing recorded runs.

### Verification approach

`pnpm test:clients` for the fold, plus a real `runtui.sh` run per outer loop confirming pending placeholders match the roles the loop actually runs.

#284

AyanBinRafaih:
## Problem

Fixes #584. Which agent roles a loop runs per round was a client-side guess: `expectedRoles(outerLoop)` in `clients/core-state/src/run-map.ts` hardcoded the role list for `agent`, `plain`, and `evolve`, and the frontend seeded one pending placeholder per role for each round from that table. The backend, which actually decides which agents run, never advertised the set. The two could silently drift: a loop gaining or losing a role, a renamed role, or a new loop kind would leave the TUI seeding stale placeholders (roles that will never run shown as pending forever) or seeding nothing at all for an unknown loop, with no diagnostic and no way for the backend to correct the display. This inverts the repo's contract direction, where backend event schemas are the frontend contract and frontends consume published semantic data instead of duplicating backend knowledge.

## Solution

The backend now owns and advertises the contract. A new module, `src/vibesys/loops/roles.py`, holds the single authoritative loop-to-roles mapping, ground-truthed from the actual `LoopContext.invoke(kind=...)` call sites in each loop (agent: `orchestrator`, `implementer`, `judge`, plus the conditional `profiler`; plain: `implementer`, `judge`, `perf_eval`; evolve: `implementer`, `judge`, `profiler`; conditional roles are included because a pending placeholder is the right display for a role that may still run). The module is deliberately import-light so `entrypoints.headless` can read it while emitting `run_started`, before any loop package loads, and `expected_agent_roles()` rejects an unknown loop kind loudly. Two safeguards keep the registry honest. A source-scanning drift test derives each loop's roles from the code itself (the `kind="..."` literals at the invoke call sites plus the shared `invoke_profiler` wrapper) and asserts equality with the registry, so an invoke added or removed without updating the registry fails tests. And `dispatch` filters `profiler` out of the advertised tuple when the CLI-resolved profiler kind is `NONE`, since a run started with `--profiler none` never invokes the profiler and would otherwise show a placeholder that can never resolve. Both `RunStartedData` models (`src/vibesys/run/events.py` and `src/server/events.py`) gain `expected_roles: tuple[str, ...] = ()`, defaulting empty so every recording that predates the field still validates, and `headless.dispatch` populates it. The protocol schema and TS bindings are regenerated, never hand-edited. On the client, the fold stores the advertised roles as `RunMapState.expectedRoles` / `CoreState.expectedRoles` and seeds placeholders from them; the old table survives only as `legacyExpectedRoles`, documented as frozen and consulted solely for recordings that predate the contract. For a loop neither side knows, the exported selector `expectedRolesForSeeding` returns null and the round degrades gracefully to exactly the phases its events actually carry, since `ensurePhase` still creates every observed role.

### Architecture

The roles registry is application configuration with one owner: the loops package describes what it runs, `headless` wires that into the `run_started` event, and the frontend consumes the published field. `run-map.ts` keeps ownership of seeding but no longer owns the knowledge of what to seed; `core-state.ts` carries `expectedRoles` through `reduceEventPrefix` with the same keep-newest rule as `outerLoop`, so prefix compaction cannot drop the contract. The wire bridge (`TypeAdapter` in `src/server/integration.py`) round-trips the field between the duplicated core and server event models, covered by a new bridge test in both directions (with the field, and legacy payloads without it).

```mermaid
flowchart TD
    RS[(vibesys/loops/roles.py<br/>EXPECTED_AGENT_ROLES:<br/>one authoritative mapping)] --> H[entrypoints/headless.py dispatch]
    IS[LoopContext.invoke call sites<br/>agent / plain / evolve loops] -. ground truth,<br/>registry test pins coverage .-> RS
    H --> EV[run_started event<br/>expected_roles advertised]
    EV --> W[wire bridge + protocol schema<br/>regenerated TS bindings]
    W --> F[core-state fold:<br/>RunMapState.expectedRoles]
    F --> SEL{expectedRolesForSeeding}
    SEL -->|advertised roles| SEED[seed one pending<br/>placeholder per role per round]
    SEL -->|"null roles (pre-field recording)"| LEG[legacyExpectedRoles:<br/>frozen fallback table]
    SEL -->|unknown loop| DEG[no seeding: round shows<br/>exactly the observed phases]
    LEG --> SEED
```

## Verification

Automated: two new client tests fail against the unfixed fold (a novel loop's advertised roles were ignored, and a known loop could not advertise a role the legacy table lacked); backend tests pin the registry against dispatch's loop set, the emission for all three loops, the wire bridge round trip, and legacy-recording compatibility; the committed-schema parity test passes. The full Python and TS pipelines pass. Manual: a real codex-provider agent-loop run on this branch showed `expected_roles` on the wire and the seeded placeholders in the TUI.

### Correctness properties

- The backend is the single source of truth for per-round roles: `run_started` advertises the same tuple the loop's `invoke` call sites can produce, and a registry test forces the mapping to cover exactly the loop kinds `dispatch` accepts, so adding a loop without declaring its roles fails tests.
- The registry cannot silently drift from the code: a source-scanning test re-derives each loop package's roles from its invoke literals and `invoke_profiler` calls and asserts set equality with `EXPECTED_AGENT_ROLES`.
- A disabled profiler is not advertised: with `--profiler none`, the `run_started` tuple omits `profiler`, so the TUI seeds no placeholder for a role the loop is guaranteed to skip.
- Recordings made before the field exist keep working: the empty default validates through both event models and the wire bridge, and the client falls back to the frozen `legacyExpectedRoles` table for them, preserving today's rendering byte for byte.
- A frontend never invents roles for a loop the backend advertised: advertised roles win over the legacy table whenever present and non-empty.
- Unknown-loop degradation is visible and safe: with no advertised and no legacy roles, nothing is seeded, and every role that actually runs still appears because observed events create their phases.
- `expected_roles` survives prefix compaction (`reduceEventPrefix` keeps the newest non-null value, mirroring `outerLoop`), so a client that folds a snapshot plus a suffix sees the same contract as one that folded every event.

### Testing

- `clients/core-state/src/run-map.test.ts`, new "expected phase seeding" describe block (6 tests): a novel loop `swarm` with novel roles seeds placeholders for each (pre-fix: only the one observed phase existed, `expect` received `[["implementer","active"]]` instead of the three-role expectation); a known loop advertising an extra `benchmark` role seeds it without any client edit (pre-fix: missing); legacy fallback for role-less `run_started`; an empty advertised list treated as absent; unknown loop `mystery` returning null from `expectedRolesForSeeding` with observed-only phases; and roles folding onto state. The first two fail before the fix, all pass after.
- `clients/core-state/src/core-state-prefix.test.ts`: the prefix-merge test now pins `expectedRoles` behavior across `reduceEventPrefix`.
- `tests/entrypoints/test_headless.py::test_expected_role_registry_covers_every_outer_loop` (registry equals dispatch's `_OUTER_LOOPS` and `_LOOP_COMMANDS`, no empty role sets) and `test_dispatch_advertises_expected_roles_on_run_started` parametrized over `agent`, `plain`, and `evolve`.
- `tests/vibesys/loops/test_roles.py` (new): the drift test scans each loop package's source for `kind="..."` invoke literals and `invoke_profiler(` calls and asserts the derived per-loop sets equal the registry exactly (agent: orchestrator/implementer/judge/profiler, plain: implementer/judge/perf_eval, evolve: implementer/judge/profiler).
- `tests/entrypoints/test_headless.py::test_dispatch_omits_profiler_role_when_profiler_is_disabled` (parametrized over `agent` and `evolve`): dispatching with `--profiler none` advertises a tuple without `profiler`; the default-profiler presence case is already pinned by the advertise test above.
- `tests/server/test_integration.py::test_run_started_expected_roles_round_trip_through_the_wire_bridge`: the core payload bridges to the wire model with the field intact, and a legacy payload without the field validates as empty.
- `uv run pytest tests/server/test_protocol_schema.py` (1 passed), `uv run pytest tests/server` (266 passed), `uv run pytest tests/entrypoints tests/vibesys/loops` (862 passed), `uv run ruff check`, `uv run ty check`, `./scripts/check_format.sh`: all clean.
- `pnpm check:ts-architecture` (no dependency violations), `pnpm check:clients`, `pnpm test:clients` (81 core-state, 549 tui, all pass), `pnpm build:clients`: all clean. Generated files were produced only by `pnpm --dir clients/backend-client generate:protocol`.
- Codex-provider harness run (`runtui.sh start 200x50 spsc`, model `gpt-5.6-sol`, agent loop, 1 round): the persisted `run_started` event carried `"expected_roles": ["orchestrator", "implementer", "judge", "profiler"]`, the round view's Agents pane showed the seeded placeholders ("5 agents · 1 active · 1 done · 3 waiting") while only the orchestrator had run, `backend.log` was clean, and teardown on quit was clean. The plain and evolve loops are covered by the parametrized dispatch test rather than separate live runs.

Closes #584.

### DIFF
--- a/clients/backend-client/src/generated/protocol.generated.ts
+++ b/clients/backend-client/src/generated/protocol.generated.ts
@@ -285,6 +285,7 @@ export type Kind9 = "run_started";
 export type OuterLoop = string;
 export type Input = string;
 export type MaxRounds = number;
+export type ExpectedRoles = string[];
 export type Kind10 = "run_interrupted";
 export type Reason = string;
 export type Signal = string | null;
@@ -816,6 +817,7 @@ export interface RunStartedData {
   outer_loop: OuterLoop;
   input: Input;
   max_rounds: MaxRounds;
+  expected_roles?: ExpectedRoles;
   [k: string]: unknown;
 }
 export interface RunInterruptedData {
--- a/clients/backend-client/src/generated/protocol.schema.json
+++ b/clients/backend-client/src/generated/protocol.schema.json
@@ -2820,6 +2820,14 @@
         "max_rounds": {
           "title": "Max Rounds",
           "type": "integer"
+        },
+        "expected_roles": {
+          "default": [],
+          "items": {
+            "type": "string"
+          },
+          "title": "Expected Roles",
+          "type": "array"
         }
       },
       "required": [
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -268,6 +268,9 @@ describe('prefix merges across the chunk boundary', () => {
     expect(full.status).toBe('running');
     expect(merged.status).toBe('connecting');
     expect(merged.outerLoop).toBe('plain');
+    // `runStartedEvent` advertises no roles, so this also pins the legacy
+    // table fallback for recordings that predate `expected_roles`.
+    expect(full.expectedRoles).toBeNull();
     expect(full.phases.map(phase => phase.kind)).toEqual(['implementer', 'judge', 'perf_eval']);
     expect(merged.phases.map(phase => phase.kind)).toEqual(['implementer']);
     expect(merged.transcript.map(entry => entry.content)).toEqual(['work more']);
@@ -414,6 +417,7 @@ function generateRunEvents(seed: number, options: {typedTools: boolean}, rounds 
       outer_loop: 'agent',
       input: '/synthetic/target',
       max_rounds: rounds,
+      expected_roles: [...AGENT_KINDS],
     },
   });
 
--- a/clients/core-state/src/core-state.ts
+++ b/clients/core-state/src/core-state.ts
@@ -132,6 +132,11 @@ export interface CoreState {
   agentKind: string | null;
   roundLabel: string | null;
   outerLoop: string | null;
+  /**
+   * The agent roles the backend advertised in `run_started`; null on
+   * recordings that predate the field (see `expectedRolesForSeeding`).
+   */
+  expectedRoles: readonly string[] | null;
   maxRounds: number | null;
   rounds: RoundSummary[];
   phases: AgentPhase[];
@@ -169,6 +174,7 @@ export function initialCoreState(): CoreState {
     agentKind: null,
     roundLabel: null,
     outerLoop: null,
+    expectedRoles: null,
     maxRounds: null,
     rounds: [],
     phases: [],
@@ -350,6 +356,7 @@ export function reduceEventPrefix(
     agentKind: state.agentKind ?? older.agentKind,
     roundLabel: state.roundLabel ?? older.roundLabel,
     outerLoop: state.outerLoop ?? older.outerLoop,
+    expectedRoles: state.expectedRoles ?? older.expectedRoles,
     maxRounds: state.maxRounds ?? older.maxRounds,
     rounds: mergeRoundLists(older.rounds, state.rounds),
     phases: mergePhaseLists(older.phases, state.phases),
@@ -541,6 +548,7 @@ function foldEvent(state: CoreState, event: RunEvent, folder: TranscriptFolder |
   if (event.round_label) next.roundLabel = event.round_label;
   const runMap = applyRunMapEvent(next, event);
   next.outerLoop = runMap.outerLoop;
+  next.expectedRoles = runMap.expectedRoles;
   next.rounds = runMap.rounds;
   next.phases = runMap.phases;
 
--- a/clients/core-state/src/run-map.test.ts
+++ b/clients/core-state/src/run-map.test.ts
@@ -1,6 +1,11 @@
 import {describe, expect, it} from 'bun:test';
 import type {RunEvent} from '@vibesys/backend-client';
-import {applyRunMapEvent, type RunMapState, roundAgentElapsedMs} from './run-map.js';
+import {
+  applyRunMapEvent,
+  expectedRolesForSeeding,
+  type RunMapState,
+  roundAgentElapsedMs,
+} from './run-map.js';
 
 describe('run map projection', () => {
   it('tracks concurrent same-role executions independently', () => {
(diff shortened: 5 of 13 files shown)


AyanBinRafaih:
Review follow-up, fixed in 4e4d04a:

- **Registry drift (P3)**: nothing forced `EXPECTED_AGENT_ROLES` to track the actual invoke sites, so an added or removed `ctx.invoke(kind=...)` would silently desync the advertised roles. New `tests/vibesys/loops/test_roles.py` scans each loop package's source for `kind="..."` invoke literals plus `invoke_profiler(` calls and asserts set equality with the registry per loop.
- **Disabled profiler advertised (P3)**: a run started with `--profiler none` still advertised `profiler`, seeding a TUI placeholder that can never resolve. `dispatch` now filters `profiler` out of the tuple when `args.profiler is ProfilerKind.NONE`, pinned by a parametrized test over the agent and evolve loops.

Verification: `uv run pytest tests/entrypoints tests/vibesys/loops` 862 passed; ruff, ty, and format checks clean.

PR #599, opened 2026-09-04
TUI: distinguish carried-forward rounds from measured ones

### LINKED ISSUE #585: TUI: Distinguish carried-forward rounds from freshly measured ones
### Workstream

CLI/DX

### Problem

A round whose metric was carried forward rather than freshly profiled is plotted identically to a measured round, so the operator reads stale or inherited numbers as new measurements and misjudges the trajectory. The backend already ships the distinguishing flag: `PerformanceRound.profile_skipped` (`src/vibesys/server/protocol.py:240`, set from the record at `src/vibesys/server/service.py:201`, computed in the loop at `src/vibesys/loops/agent/loop.py:3174,3398` and treated there as not-a-fresh-measurement). No client reads it: the rail and perf pane consume `perf_metric`, `perf_unit`, `passed`, `perf_delta_pct` but never `profile_skipped`.

### Desired outcome

Skipped rounds get a dim or hollow-marker treatment (or a distinct glyph) in the round rail and the performance sparkline, with a legend note, and the delta column suppresses or footnotes deltas whose baseline was `profile_skipped`.

### Acceptance criteria

- [ ] `profile_skipped` is folded into round state and reaches the rail and perf pane.
- [ ] Skipped rounds render visibly distinct from measured rounds, legible without color.
- [ ] Deltas computed against a skipped baseline are marked as such.

### Scope and non-goals

In scope: 
- surfacing the existing flag. 

Not in scope: 
- changing how the loop decides to skip profiling.

Reuse the existing marker vocabulary so the rail stays readable; no new backend field is required.

### Verification approach

`pnpm test:clients` for the fold and rail label, plus a run with a carried-forward round.

#284

AyanBinRafaih:
## Problem

Fixes #585. A round where the loop skipped profiling rendered identically to a freshly measured round: the rail showed the same `✓` in the same tone, so an operator scanning a run could not tell which rounds actually produced a performance reading and which ran without one. The loop already records the distinction (`profile_skipped` on the round record, surfaced through the `performance_rounds` query), but no client surface read it, and the round state the rail folds from the event stream could not read it: `round_finished` did not carry the flag, and the architecture keeps query results outside core state (`core-state` is a pure fold over snapshot and events, and query payloads are explicitly not folded). Meeting the acceptance criterion "folded into round state" therefore requires the event contract to publish what the loop already records.

An important semantic pins the scope of the fix: in the loop, `profile_skipped=True` strictly implies `perf_metric is None` on both the round record and the event (one constructor, one emission site), and the `performance_rounds` query filters to rounds with a reading. A skipped round therefore records no perf reading at all, never appears in the performance chart's series, and never has a delta. Nothing is "carried forward"; the round simply measured nothing, and the only surface that can mark it is the rail row itself.

## Solution

The backend addition is one additive field: `RoundFinishedData.profile_skipped: bool = False` in both event modules (`src/vibesys/run/events.py` and `src/server/events.py`), populated at the agent loop's `ROUND_FINISHED` emission from the same `profile_skipped` local that feeds the round record, so the event and the `performance_rounds` query agree by construction. The default keeps every legacy recording valid, and the protocol schema and TS bindings are regenerated rather than hand-edited. On the client, `core-state` folds the flag onto the round as an optional `profileSkipped` on a new `RoundState extends RoundSummary`, patched after the run map settles, which leaves `run-map.ts` untouched and lets the existing spread-based merges carry the property through `reduceEventPrefix` unchanged.

Presentation is deliberately confined to the one surface a skipped round can reach: in the rounds rail, a completed round with `profileSkipped` shows a hollow `○` in place of `✓` and dims the row to `textSubtle` after the selection, running, planned, and failed precedences. A failed round keeps `✗`, because how the round ended outranks how it measured, and that precedence is now pinned by a test. The performance chart is untouched: it plots only fresh readings (solid `●`), because a skipped round has no `perf_metric` and cannot enter its series; likewise no delta marker exists, because a round without a reading has no delta. An earlier revision of this branch added hollow chart markers, a chart legend, and a `~` delta prefix for skipped rounds; review showed those code paths were unreachable given the backend invariant above, and they were removed along with the tests that fabricated the impossible payloads (a `perf_metric` on a skipped round). Wording in the field comments, README, and test docstrings was corrected from "carried forward" to "no fresh profile ran this round".

### Architecture

The loop owns the fact (did a fresh profile run this round), the event contract publishes it, `core-state` folds it, and the TUI decides how to draw it, which is the repo's standard unidirectional flow with no presentation leaking into the backend. The one contract change is additive and lands in both duplicated event models in lockstep, with the wire round trip and legacy-payload compatibility tested at the boundary.

```mermaid
flowchart TD
    L[agent loop round finish<br/>profile_skipped local<br/>implies perf_metric is None] --> REC[RoundRecord] --> Q[performance_rounds query<br/>filters to rounds with a reading]
    L --> EV[round_finished event<br/>profile_skipped: new additive field]
    EV --> CS[core-state fold:<br/>RoundState.profileSkipped]
    CS --> RAIL[round rail: hollow glyph on<br/>completed rounds, textSubtle dim,<br/>failed keeps its glyph]
    Q --> CHART[performance chart:<br/>fresh readings only, unchanged]
```

## Verification

Automated: new client tests fail against the unfixed code (the rail glyph and dim, the core-state fold, and the prefix-boundary carry), a renderer test pins that failure outranks the skip marker, and a new Python test proves the loop emits the flag into the journaled events; the committed-schema parity test passes. The full Python and TS pipelines pass. Manual: a real codex-provider agent-loop run on this branch exercised the new emission and fold on the live event stream.

### Correctness properties

- The `round_finished` event and the `performance_rounds` query cannot disagree about whether a round measured: both are fed by the same local at the same emission site.
- `profile_skipped=True` implies no perf reading, on the record and on the event alike, so a skipped round contributes no chart point and no delta; the rail treatment is the complete rendering surface, and no client code path assumes the impossible combination of a skipped round with a metric.
- Legacy recordings and old clients are unaffected: the field defaults to `False`, a raw legacy payload without it validates, and an absent `profileSkipped` renders exactly as before.
- The distinction is legible without color: the `○` versus `✓` glyph carries the information, with the `textSubtle` dim as reinforcement.
- Failure display outranks measurement provenance: a failed round renders `✗` regardless of `profileSkipped`, pinned by a renderer test.
- The flag survives prefix compaction: a client folding a snapshot plus a suffix sees the same `profileSkipped` as one that folded every event.
- `run-map.ts` is untouched; the fold extension lives in `core-state.ts` as an optional property, so no existing consumer of `RoundSummary` changes behavior.

### Testing

- `clients/tui/src/ui/round-rail.test.ts`: a completed profile-skipped round with no perf data renders `○` (not `✓`) with `textSubtle` foreground while a fresh round keeps `✓` and the primary tone; and a failed round with `profileSkipped: true` renders `✗`, not `○`. The precedence test was mutation-checked by temporarily removing the `completed` gate in `statusGlyph`, which makes it fail.
- `clients/core-state/src/core-state.test.ts` (2 new): the flag lands on the round when true and stays unset when absent; `clients/core-state/src/core-state-prefix.test.ts` (1 new): the flag survives `foldAsPrefix` with tail activity on a later round. All fail before the fix (`Expected: true, Received: undefined`).
- `tests/server/test_events.py::TestRoundFinishedProfileSkipped`: wire round trip with the flag true, and a legacy raw payload without the field validating as `False`. `tests/vibesys/loops/agent/test_orchestrate.py::test_round_finished_events_carry_profile_skipped`: a two-round run with metrics `[None, 321.5]` journals `round_finished` events whose flags are `[True, False]`, matching the records; fails before the fix with `KeyError: 'profile_skipped'`.
- `uv run pytest tests/server/test_protocol_schema.py` (1 passed), `uv run pytest tests/server` (267 passed), `uv run pytest tests/vibesys/loops/agent` (352 passed), and after the review commit (whose Python changes are comment-only) `uv run pytest tests/vibesys/loops/agent/test_orchestrate.py tests/server/test_events.py` (167 passed); `uv run ruff check`, `uv run ty check`, `./scripts/check_format.sh`: all clean.
- `pnpm check:ts-architecture`, `pnpm check:clients`, `pnpm test:clients` (25 backend-client, 78 core-state, 551 tui, all pass), `pnpm build:clients`: all clean. Generated files came only from `pnpm --dir clients/backend-client generate:protocol`.
- `clients/tui/README.md` documents the `○` glyph and the dim, and states that a round without a fresh profile records no reading and therefore never carries a delta or a chart point.
- Codex-provider harness run (`runtui.sh start 200x50 spsc`, model `gpt-5.6-sol`, agent loop, 1 round, `--profiler none`): healthy run through the modified emission path with a clean `backend.log`; the deterministic skipped-round rendering itself is pinned by the unit tests above, since a single-round smoke cannot produce the multi-round sequence on demand.

Closes #585.

### DIFF
--- a/clients/backend-client/src/generated/protocol.generated.ts
+++ b/clients/backend-client/src/generated/protocol.generated.ts
@@ -326,6 +326,7 @@ export type Attempts = number;
 export type JudgeVerdict = "pass" | "fail" | "skipped";
 export type PerfMetric = number | null;
 export type PerfUnit = string | null;
+export type ProfileSkipped = boolean;
 export type Kind20 = "tool_call";
 export type Tool1 = string;
 export type CallId = string | null;
@@ -359,7 +360,7 @@ export type Round = number;
 export type PerfMetric1 = number;
 export type PerfUnit1 = string;
 export type Passed = boolean;
-export type ProfileSkipped = boolean;
+export type ProfileSkipped1 = boolean;
 export type Performance = PerformanceRound[];
 export type ObjectiveMetric = string | null;
 export type ObjectiveUnit = string | null;
@@ -909,6 +910,7 @@ export interface RoundFinishedData {
   judge_verdict: JudgeVerdict;
   perf_metric?: PerfMetric;
   perf_unit?: PerfUnit;
+  profile_skipped?: ProfileSkipped;
   [k: string]: unknown;
 }
 export interface ToolCallData {
@@ -972,7 +974,7 @@ export interface PerformanceRound {
   perf_metric: PerfMetric1;
   perf_unit: PerfUnit1;
   passed: Passed;
-  profile_skipped?: ProfileSkipped;
+  profile_skipped?: ProfileSkipped1;
 }
 /**
  * What the performance plot measures and how to read it.
--- a/clients/backend-client/src/generated/protocol.schema.json
+++ b/clients/backend-client/src/generated/protocol.schema.json
@@ -2456,6 +2456,11 @@
           ],
           "default": null,
           "title": "Perf Unit"
+        },
+        "profile_skipped": {
+          "default": false,
+          "title": "Profile Skipped",
+          "type": "boolean"
         }
       },
       "required": [
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -200,6 +200,22 @@ describe('prefix merges across the chunk boundary', () => {
     ]);
   });
 
+  it('keeps the carried-forward flag on a round the chunk finished', () => {
+    const events = [
+      chunkEvent(1, 'work'),
+      roundFinishedEvent(2, {profile_skipped: true}),
+      {...chunkEvent(3, 'next'), round_label: 'round-2-implementer'},
+    ];
+
+    // The flag lands in the chunk fold; the tail only touches round 2, so the
+    // boundary merge must hand round 1 through with the flag intact.
+    const merged = foldAsPrefix(events, 2);
+
+    expect(merged).toEqual(reduceEventBatch(initialCoreState(), events));
+    expect(merged.rounds[0]?.profileSkipped).toBe(true);
+    expect(merged.rounds[1]?.profileSkipped).toBeUndefined();
+  });
+
   it('merges a chunk diagnostic with its tail update by id', () => {
     const events = [
       diagnosticEvent(1, 'diag-1', 'warning', 'Agent stalled', null),
@@ -704,7 +720,7 @@ function runStartedEvent(sequence: number): RunEvent {
   };
 }
 
-function roundFinishedEvent(sequence: number): RunEvent {
+function roundFinishedEvent(sequence: number, extra: {profile_skipped?: boolean} = {}): RunEvent {
   return {
     sequence,
     timestamp: timestamp(sequence),
@@ -717,6 +733,7 @@ function roundFinishedEvent(sequence: number): RunEvent {
       judge_verdict: 'pass',
       perf_metric: null,
       perf_unit: null,
+      ...extra,
     },
   };
 }
--- a/clients/core-state/src/core-state.test.ts
+++ b/clients/core-state/src/core-state.test.ts
@@ -895,6 +895,23 @@ describe('batched transcript folding', () => {
   });
 });
 
+describe('the carried-forward profile flag', () => {
+  it('lands on the round whose round_finished event skipped profiling', () => {
+    const state = reduceEvent(initialCoreState(), roundFinishedEvent(1, {profile_skipped: true}));
+
+    expect(state.rounds).toHaveLength(1);
+    expect(state.rounds[0]?.status).toBe('completed');
+    expect(state.rounds[0]?.profileSkipped).toBe(true);
+  });
+
+  it('stays unset when the event lacks the field, as legacy streams do', () => {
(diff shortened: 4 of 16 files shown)


AyanBinRafaih:
Review follow-up, fixed in 4f31723:

- **Unreachable chart surfaces (P1)**: in the loop, `profile_skipped=True` strictly implies `perf_metric is None` (one round-record constructor, one `ROUND_FINISHED` emission site), and `performance_rounds` filters to rounds with a reading. A skipped round therefore never enters the chart series and never has a delta, so the hollow chart markers, the chart legend line, and the `~` delta prefix were dead code, and their tests fabricated impossible payloads (a metric on a skipped round). All of that is deleted; the rail treatment (hollow `○` on completed skipped rounds plus the `textSubtle` dim) is the complete surface and stays.
- **Wrong semantics in wording (P2)**: nothing is "carried forward"; the round simply measured nothing. README, both `profile_skipped` field comments, and the orchestrate test docstring now say no fresh profile ran this round.
- **Missing precedence test (P3)**: a failed round with `profileSkipped` must render `✗`, not `○`; now pinned by a renderer test (mutation-checked against a removed gate).

Verification: TS architecture, tsc, biome, tests (78 core-state, 551 tui), and builds clean; `tests/vibesys/loops/agent/test_orchestrate.py` + `tests/server/test_events.py` 167 passed; format check clean. PR body rewritten to match.

PR #601, opened 2026-09-04
TUI: make tail bootstrap use one atomic watermark

### LINKED ISSUE #587: TUI: Make tail bootstrap use one atomic watermark
### Observed behavior

A tail subscription calculates its history floor from one sequence watermark, then reads the checkpoint from a potentially newer journal state.

The server currently obtains `snapshot.sequence`, sends the `subscribed` message, calculates `latest_sequence - tail` from that earlier snapshot, and then calls `subscription_checkpoint()` against the current store. If thousands of events arrive between those operations, the calculated floor can be far too low. A request for the newest 1,000 events can therefore receive the entire event history.

This defeats the bounded-startup behavior introduced by [PR #468](https://github.com/uw-syfi/vibesys/pull/468). On a long run it can reintroduce a large server parse, a large JSON batch, a large frontend fold, and delayed first paint.

### Expected behavior

The floor, through-sequence watermark, event window, run-level spine, and active-execution checkpoint should describe one atomic journal state. Regardless of concurrent append activity, a bootstrap with `tail=N` should return at most the newest `N` ordinary events at the captured watermark, the documented pre-floor run-level spine, and events appended after that watermark in later live batches.

### Reproduction

1. Create a server with an event journal at sequence 100.
2. Subscribe using `tail=500`.
3. Pause the handler after `snapshot()` returns but before `_write_bootstrap()` takes its checkpoint.
4. Append several thousand events.
5. Resume the handler.
6. Observe that the floor was calculated from sequence 100 while the checkpoint includes the much newer store, causing the bootstrap to include far more than 500 ordinary events.

### Affected subsystem

### Environment

OS: Linux (debian) 
Backend: Unix JSONL supervision transport  
Client: TUI tail bootstrap  
Run shape: active run producing events concurrently with subscription startup

### Impact

Blocks normal use

### Relevant logs

```shell
A regression test should record the snapshot watermark, checkpoint watermark, requested tail, and ordinary bootstrap count. The expected count must remain bounded by the requested tail plus explicitly documented spine records.
```

#284, #468, #470

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #587. A tail subscription's bootstrap was assembled from two different journal states. The transport handler called `api.snapshot()` to get a sequence watermark, computed the history floor as `latest_sequence - tail` from that watermark, and only then called `subscription_checkpoint()` against the current store. Any events appended between those two reads made the computed floor stale: a request for the newest 1,000 events could receive the entire history. That defeats the bounded-startup contract introduced by #468, and on a long run it reintroduces exactly what that PR removed: a large server-side parse, a huge JSON batch on the wire, a large frontend fold, and delayed first paint. The same racy sequence was duplicated in the in-loop rebootstrap path that re-tails when the durable store attaches mid-subscription.

## Solution

The bootstrap read is moved behind the API boundary into one method that captures everything under a single lock acquisition. `RunApi.subscription_bootstrap(after_sequence, tail)` takes `self._condition` once and, while holding it, reads the latest sequence, computes the floor (`after_sequence` when `tail` is `None`, else `max(after_sequence, latest - tail)`), takes the checkpoint at that floor with the documented pre-floor run-level spine, and reads the run id and active executions. It returns a frozen `SubscriptionBootstrap` dataclass carrying `run_id`, `floor`, `through_sequence`, `events`, and `active_executions`, all describing one journal state. Appends acquire the same condition, so no event can land between the floor computation and the checkpoint. The transport handler (`_stream` in `unix_jsonl.py`) now calls this method for both the initial bootstrap and the in-loop rebootstrap, and no longer calls the racy `snapshot()` at all; the `subscribed` message's `latest_sequence` comes from the same captured watermark. The module-level `_history_floor` helper is deleted, and the one piece of it that was presentation policy rather than journal state (report floor 0 to the client when no tail was requested, because nothing was withheld) moves into `_write_bootstrap` with its rationale.

Moving the bootstrap ahead of the `subscribed` handshake changed one failure ordering: a bootstrap exception used to surface after the handshake (the replay was read later) but would now surface before it, and the client probes `tail` support by dialing and treating a pre-handshake failure as a server without the field, which would send it into a full-history retry against the same fault. The handler therefore acknowledges the accepted subscribe first when the bootstrap raises (run id and watermark from the API) and then lets the failure propagate as the usual `stream_failed` protocol error, preserving the wire ordering the client's fallback logic assumes.

### Architecture

The watermark arithmetic now lives where the lock lives: `RunApi` in `src/server/api/service.py` owns the journal's condition variable, so it is the only layer that can promise atomicity across the floor computation and the checkpoint. The transport layer (`src/server/transport/unix_jsonl.py`) is reduced to framing: it asks the API for one `SubscriptionBootstrap` and serializes it, keeping only the reported-floor presentation rule. No protocol message shape changes, so the frontend is untouched.

```mermaid
sequenceDiagram
    participant C as TUI client
    participant T as transport _stream
    participant A as RunApi.subscription_bootstrap
    participant J as journal (under condition)
    C->>T: subscribe(after_sequence, tail)
    T->>A: subscription_bootstrap(after, tail)
    activate A
    Note over A,J: one condition acquisition,<br/>appends blocked
    A->>J: latest_sequence_locked()
    A->>A: floor = max(after, latest - tail)
    A->>J: checkpoint_locked(floor, bootstrap_spine)
    A->>J: run_id_locked(), active_locked()
    deactivate A
    A-->>T: SubscriptionBootstrap(run_id, floor,<br/>through_sequence, events, active)
    T-->>C: subscribed(latest_sequence=through)
    T-->>C: event batch, at most tail ordinary events + spine
    Note over T: same atomic call reused by the<br/>in-loop rebootstrap when the durable<br/>store attaches mid-subscription
```

## Verification

Automated: a deterministic regression test reproduces the exact pre-fix interleaving (3,000 events landing between the old snapshot read and the checkpoint) and fails on the unfixed code with `3040 <= 40`, recording the snapshot watermark, checkpoint watermark, requested tail, and ordinary bootstrap count that the issue asked for; a unit test pins the atomic method's contract; a stress test races a live writer thread against a real socket subscription. Full server suite, lint, type check, and format check pass. Manual: a real codex-provider TUI run on this branch landed and streamed live events normally with a clean `backend.log`.

### Correctness properties

- The floor, through-sequence watermark, event window, pre-floor spine, run id, and active-execution checkpoint of a bootstrap all describe one journal state, captured under a single acquisition of the same condition that appends take.
- A bootstrap with `tail=N` contains at most `N` ordinary events (sequence above the floor) regardless of concurrent append volume; everything at or below the floor is a documented spine record type.
- `floor == through_sequence - tail` when the journal is longer than the tail, and the `subscribed` message's `latest_sequence` equals the batch's `through_sequence`, so the client's cursor and the replay window cannot disagree.
- Events appended after the captured watermark arrive only in later live batches, never folded into the bootstrap.
- Without `tail`, the client receives everything after its own cursor and the reported floor stays 0, preserving the wire behavior old clients expect.
- A bootstrap failure reaches the client only after the `subscribed` handshake, as a `stream_failed` protocol error, so the client's tail-support probe cannot misread a transient replay fault as an old server and retry the full history.

### Testing

- `tests/server/test_transport_subscription.py::test_bootstrap_tail_stays_bounded_when_events_land_mid_bootstrap`: wraps `subscription_bootstrap` to append a 3,000-event burst after the handler commits to bootstrapping but before the locked read; the bootstrap stays bounded at 40 ordinary events with the correct floor, through-sequence, and spine-only pre-floor records. (On the pre-fix code the equivalent race was injected through the then-live `snapshot()` hook and failed with `assert 3040 <= 40`.)
- `tests/server/test_transport_subscription.py::test_bootstrap_failure_surfaces_after_the_handshake`: monkeypatches `subscription_bootstrap` to raise and asserts the wire carries `subscribed` first and then a `protocol_error` with code `stream_failed`; without the handshake-first guard the failure arrives before the handshake and the test fails.
- `tests/server/test_transport_subscription.py::test_subscription_bootstrap_captures_one_atomic_state`: pins the returned `SubscriptionBootstrap` for both `tail=40` (bounded window, exact spine sequence) and `tail=None` (floor 0, full replay).
- `tests/server/test_transport_subscription.py::test_bootstrap_tail_stays_bounded_under_concurrent_appends`: a writer thread appends up to 20,000 live events while a real socket client subscribes with `tail=10`; the batch never exceeds 10 ordinary events.
- `tests/server/test_api_transport.py`: the stream-failure test now triggers its fault through a live append instead of the removed bootstrap-time `snapshot()` call, keeping the stream-error path covered.
- `uv run pytest tests/server`: 269 passed. `uv run ruff check`, `uv run ty check`, `./scripts/check_format.sh`: clean.
- Codex-provider harness run (`runtui.sh start 200x50 spsc`, model `gpt-5.6-sol`, 1 round): healthy landing, live streaming through the new bootstrap path, no tracebacks or protocol errors in `backend.log`.

Closes #587.

### DIFF
--- a/src/server/api/service.py
+++ b/src/server/api/service.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import threading
+from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
 from server.api.design import DesignLog
@@ -56,6 +57,17 @@ if TYPE_CHECKING:
     from server.journal import EventJournal
     from server.settings import InteractiveSetupDefaults
     from vibesys.loops.agent.model import AgentRunState
+
+
+@dataclass(frozen=True)
+class SubscriptionBootstrap:
+    """One journal state captured atomically for a subscription bootstrap."""
+
+    run_id: str
+    floor: int
+    through_sequence: int
+    events: list[RunEvent]
+    active_executions: list[ActiveAgentExecution]
 
 
 class RunApi:
@@ -226,6 +238,29 @@ class RunApi:
                 after_sequence, bootstrap_spine=bootstrap_spine
             )
             return through_sequence, events, self._executions.active_locked()
+
+    def subscription_bootstrap(
+        self, after_sequence: int, tail: int | None
+    ) -> SubscriptionBootstrap:
+        """Capture one atomic bootstrap state for a new or restarted subscription.
+
+        Appends acquire the same condition, so computing the tail floor and
+        reading the checkpoint under one acquisition keeps the replay bounded
+        by ``tail`` no matter how many events land during the bootstrap.
+        """
+        with self._condition:
+            latest = self._journal.latest_sequence_locked()
+            floor = after_sequence if tail is None else max(after_sequence, latest - tail)
+            through_sequence, events = self._journal.checkpoint_locked(
+                floor, bootstrap_spine=tail is not None
+            )
+            return SubscriptionBootstrap(
+                run_id=self._journal.run_id_locked(),
+                floor=floor,
+                through_sequence=through_sequence,
+                events=events,
+                active_executions=self._executions.active_locked(),
+            )
 
     def events(self, after_sequence: int = 0, before_sequence: int | None = None) -> list[RunEvent]:
         """Read journal events within the requested sequence bounds."""
--- a/src/server/transport/unix_jsonl.py
+++ b/src/server/transport/unix_jsonl.py
@@ -25,22 +25,9 @@ from server.api.protocol import (
 from vibesys.unix_socket import validate_socket_path
 
 if TYPE_CHECKING:
-    from server.api.service import RunApi
+    from server.api.service import RunApi, SubscriptionBootstrap
 
 _REQUEST_ADAPTER = TypeAdapter(ProtocolRequest)
-
-
-def _history_floor(request: SubscribeRequest, latest_sequence: int) -> tuple[int, int]:
-    """Return the replay floor, and the floor to report until the next bootstrap.
-
-    Without ``tail`` the reported floor stays 0: the client asked for
-    everything from its own cursor onward, so nothing was withheld and old
-    clients see the field's default.
-    """
-    if request.tail is None:
-        return request.after_sequence, 0
-    floor = max(request.after_sequence, latest_sequence - request.tail)
-    return floor, floor
 
 
 class _RequestHandler(socketserver.StreamRequestHandler):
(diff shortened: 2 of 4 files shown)


AyanBinRafaih:
Review follow-up, fixed in d015f45:

- **Failure ordering regression (P2)**: moving the bootstrap ahead of the `subscribed` handshake meant a bootstrap exception now reached the client before the handshake. The client probes `tail` support by dialing and treats a pre-handshake failure as a server without the field, so a transient replay fault would trigger a full-history retry against the same fault. The handler now acknowledges the accepted subscribe first (run id and watermark from the API) and then lets the failure propagate as the usual post-handshake `stream_failed` protocol error. Pinned by a new socket-level test asserting the wire order `subscribed` then `protocol_error`, which fails pre-fix.
- **Dead monkeypatch (P3)**: `test_bootstrap_tail_stays_bounded_when_events_land_mid_bootstrap` still patched `api.snapshot`, which the bootstrap path no longer calls (0 invocations), so the race it was written for went unexercised. It now wraps `subscription_bootstrap` to inject the 3,000-event burst between the handler committing to bootstrap and the locked read, and asserts the wrapper actually ran.

Verification: `uv run pytest tests/server` 269 passed; ruff, ty, and format checks clean.

PR #602, opened 2026-09-04
TUI: keep runtime alive until the last active subscriber closes

### LINKED ISSUE #588: TUI: Keep a reconnected interface attached until its active subscription closes
### Observed behavior

The server uses one sticky `threading.Event` to represent subscription disconnection. Every subscription handler sets it when that handler exits, and the event is never cleared.

After [PR #525](https://github.com/uw-syfi/vibesys/pull/525), the TUI can disconnect and establish a new event subscription. The old handler has already set the global disconnect event, however. When the optimization run finishes, `ServerRuntime.run()` calls `wait_for_subscriber_disconnect()`, which returns immediately because of the old disconnect. The runtime then closes the server and retained chat resources even though the replacement TUI subscription is still active.

The same mechanism is incorrect for multiple simultaneous presentation clients: one client disconnecting allows runtime teardown while another remains attached.

### Expected behavior

Runtime teardown must be tied to a clearly defined owning subscription or to the current active subscription count. A disconnect followed by reconnect must not shorten the lifetime of the new stream. If multiple subscribers are supported, the runtime must not close until the owning or last active subscriber closes. If only one subscriber is supported, additional subscriptions should be rejected explicitly.

### Reproduction

1. Start `ServerRuntime` with a run callback blocked on a synchronization event.
2. Connect subscription A.
3. Disconnect A while the run is active.
4. Connect subscription B and confirm that it receives live events.
5. Allow the run to finish while B remains connected.
6. Observe that the earlier disconnect causes immediate runtime teardown and closes B.

### Affected subsystem

### Environment

OS: Linux (debian)
Backend: Unix JSONL supervision server  
Client: TUI with reconnect enabled  
Related feature: retained post-run experiment chat

### Impact

Blocks normal use

### Relevant logs

```shell
Expected timeline: subscription A connected, A disconnected, B connected, run finished, B remains active, B disconnected, runtime exits. A failure occurs when runtime exits immediately after the run despite B remaining active.
```

#284, #472, #525, #426

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #588. The server represented "the client disconnected" as one sticky `threading.Event` on `UnixJsonlServer`: every subscription handler set it when that handler's stream ended, and nothing ever cleared it. Since #525 the TUI can drop its event subscription and establish a new one, so the old handler's exit permanently poisoned the flag. When the optimization run finished, `ServerRuntime.run()` called `wait_for_subscriber_disconnect()`, which returned immediately because of the long-gone first subscription, and the runtime closed the server and the retained post-run chat resources out from under the live replacement subscription. The same single-bit design was wrong for multiple simultaneous presentation clients: one client disconnecting allowed runtime teardown while another was still attached.

## Solution

Disconnect stops being a sticky bit and becomes reference counting over active subscription streams. A new `SubscriptionTracker` (`src/server/transport/subscriptions.py`) owns two separate questions the old event conflated: "has any client ever subscribed?" (a `threading.Event` that is set once and intentionally never resets, backing the existing startup gate `wait_for_subscriber`) and "is any subscription active right now?" (a `Condition`-guarded counter). The request handler wraps each stream in `tracker.track()`, a context manager that increments on entry and decrements in a `finally`, notifying waiters when the count reaches zero, so the count stays correct on clean ends, client resets, and handler exceptions alike. `wait_for_subscriber_disconnect()` blocks until the count is zero and then lingers through a settle window (`RECONNECT_SETTLE_SECONDS`, 1.0s) before declaring the server subscriber-free: a dropped client redials on a backoff whose first delay is 500ms, so returning the instant the count hits zero would let teardown unlink the socket underneath that redial. A reconnect that lands inside the window notifies the condition and puts the waiter back to blocking, and with several clients attached the runtime waits for the last one. The window stays well under the launcher's 2s backend exit grace, so a deliberate quit still tears down without a SIGTERM. Teardown ordering is otherwise unchanged, and the multi-client policy the issue asked to make explicit is the one the transport already supports: concurrent subscribers are served, and the runtime outlives the last of them.

### Architecture

The tracker is transport-layer state with one owner: `UnixJsonlServer` creates it, hands it to the threaded socket server, and exposes only the two wait methods the runtime already consumed, so `ServerRuntime` needed no change at all. Handler threads touch the tracker only through `track()`, which pairs increment and decrement structurally rather than by convention. The new module keeps `unix_jsonl.py` about framing and streaming while the lifetime accounting lives in one small, independently testable unit.

```mermaid
flowchart TD
    subgraph handler threads
        A[subscription A stream] -- "with track(): +1 ... -1" --> T
        B[subscription B stream] -- "with track(): +1 ... -1" --> T
    end
    T[(SubscriptionTracker<br/>ever_subscribed: sticky Event<br/>active: Condition-guarded count)]
    S[UnixJsonlServer] -- owns --> T
    R[ServerRuntime.run] -- "wait_for_subscriber(timeout)" --> T
    R -- "wait_for_subscriber_disconnect()<br/>blocks until active == 0<br/>holds for a 1s settle window" --> T
    R --> X[close server, release retained<br/>chat resources, teardown]
    T -. "notify_all when the last<br/>active stream closes" .-> R
```

## Verification

Automated: a runtime-level regression test drives the issue's exact timeline (subscribe, disconnect mid-run, reconnect, run finishes, the reconnected subscriber still receives the terminal event before teardown) and a transport-level unit test pins that the disconnect wait blocks while any stream is active and releases when the last one closes; both fail on the unfixed code. Full server suite, lint, type check, and format check pass. Manual: a real codex-provider TUI run on this branch landed and streamed normally, and quitting the TUI tore the backend down cleanly (session directory removed, no surviving backend process).

### Correctness properties

- Runtime teardown is gated on the count of currently active subscription streams, not on whether any stream ever ended: a reconnect that overlaps the old stream, or lands within the 1s settle window after it closes, keeps the runtime alive; the count-zero gap of a slower reconnect is bounded by the window, which covers the client's 500ms first redial backoff.
- With multiple concurrent subscribers, `wait_for_subscriber_disconnect()` returns only after the last active stream has been closed for the full settle window; with none ever connected, the startup gate `wait_for_subscriber(timeout)` behaves exactly as before (sticky, never resets).
- The settle window (1.0s) stays under the launcher's 2s backend exit grace, so a deliberate quit still completes teardown before the launcher escalates to SIGTERM.
- The active count is exact under all stream exits (clean end, broken pipe, handler exception) because increment and decrement are paired in one context manager with the decrement in `finally`.
- Waiters are woken through the same condition that guards the counter, so there is no window where the count is zero but a waiter sleeps forever.
- No protocol or frontend change: the fix is entirely in the transport's lifetime accounting, and `ServerRuntime`'s call sites are untouched.

### Testing

- `tests/server/test_runtime.py::test_runtime_waits_for_reconnected_subscriber_before_teardown`: a client thread subscribes, drops the stream, subscribes again, then the run finishes; the test asserts the reconnected subscriber observes `run_finished` and that the runtime had not returned while it was attached. Before the fix the runtime tore down immediately after the run because of the first disconnect; after the fix it waits for the live stream.
- `tests/server/test_transport_subscription.py::test_disconnect_wait_blocks_until_last_subscriber_closes`: with two subscribed clients over the real socket, `wait_for_subscriber_disconnect()` still blocks after the first disconnects and returns promptly after the second does. Fails before the fix (returns after the first disconnect), passes after.
- `tests/server/test_subscription_tracker.py::test_disconnect_wait_bridges_a_redial_inside_the_settle_window`: a tracker unit test where the only stream closes and a new one opens 50ms later, inside the settle window; the waiter must stay blocked across the gap and release only after the replacement closes and its own window drains. Fails on the windowless code (the waiter returns in the gap), passes after. It lives in its own module because it needs no socket harness.
- `uv run pytest tests/server/test_runtime.py tests/server/test_transport_subscription.py`: 20 passed. `uv run pytest tests/server`: 267 passed.
- `uv run ruff check`, `uv run ty check`, `./scripts/check_format.sh`: clean.
- Codex-provider harness run (`runtui.sh start 200x50 spsc`, model `gpt-5.6-sol`, 1 round): healthy landing and live run; after quitting with Ctrl+C the launcher exited, the session directory was removed, and no backend or frontend process survived.

Closes #588.

### DIFF
--- a/src/server/transport/subscriptions.py
+++ b/src/server/transport/subscriptions.py
@@ -1,0 +1,49 @@
+"""Subscription lifetime accounting for the transport server."""
+
+from __future__ import annotations
+
+import threading
+from contextlib import contextmanager
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+
+class SubscriptionTracker:
+    """Count active event subscriptions across handler threads.
+
+    A disconnect must not end the server's lifetime while another subscription
+    is still streaming: a client that reconnects after a dropped connection,
+    or a second concurrent client, holds the count above zero until its own
+    stream closes. ``wait_for_subscriber`` answers the separate question of
+    whether any client has ever subscribed, and never resets.
+    """
+
+    def __init__(self) -> None:  # noqa: D107  # tracked: #288
+        self._ever_subscribed = threading.Event()
+        self._condition = threading.Condition()
+        self._active = 0
+
+    @contextmanager
+    def track(self) -> Generator[None]:
+        """Count one subscription stream for the duration of the block."""
+        with self._condition:
+            self._active += 1
+        self._ever_subscribed.set()
+        try:
+            yield
+        finally:
+            with self._condition:
+                self._active -= 1
+                if self._active == 0:
+                    self._condition.notify_all()
+
+    def wait_for_subscriber(self, timeout: float) -> bool:
+        """Wait until any client has established an event stream."""
+        return self._ever_subscribed.wait(timeout)
+
+    def wait_for_none_active(self) -> None:
+        """Block until no subscription streams remain active."""
+        with self._condition:
+            self._condition.wait_for(lambda: self._active == 0)
--- a/src/server/transport/subscriptions.py
+++ b/src/server/transport/subscriptions.py
@@ -9,6 +9,14 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from collections.abc import Generator
 
+# How long ``wait_for_none_active`` lingers after the count reaches zero
+# before declaring the server subscriber-free. A dropped client redials on a
+# finite backoff whose first delay is 500ms, so returning the instant the
+# count hits zero would let teardown unlink the socket underneath that
+# redial. The window must stay well under the launcher's 2s backend exit
+# grace so a deliberate quit still tears down without a SIGTERM.
+RECONNECT_SETTLE_SECONDS = 1.0
+
 
 class SubscriptionTracker:
     """Count active event subscriptions across handler threads.
@@ -30,6 +38,9 @@ class SubscriptionTracker:
         """Count one subscription stream for the duration of the block."""
         with self._condition:
             self._active += 1
+            # Wake a disconnect waiter sitting in its settle window so the
+            # reconnect extends the server's lifetime immediately.
+            self._condition.notify_all()
         self._ever_subscribed.set()
         try:
             yield
@@ -43,7 +54,17 @@ class SubscriptionTracker:
         """Wait until any client has established an event stream."""
         return self._ever_subscribed.wait(timeout)
 
-    def wait_for_none_active(self) -> None:
-        """Block until no subscription streams remain active."""
+    def wait_for_none_active(self, settle_seconds: float | None = None) -> None:
+        """Block until no stream has been active for ``settle_seconds``.
+
+        The settle window bridges a client's redial backoff: a reconnect that
+        lands inside it keeps the wait blocked, so a transient drop cannot
+        hand teardown a socket the client is about to dial again.
+        """
+        if settle_seconds is None:
+            settle_seconds = RECONNECT_SETTLE_SECONDS
         with self._condition:
-            self._condition.wait_for(lambda: self._active == 0)
+            while True:
+                self._condition.wait_for(lambda: self._active == 0)
+                if not self._condition.wait_for(lambda: self._active > 0, timeout=settle_seconds):
+                    return
--- a/src/server/transport/unix_jsonl.py
+++ b/src/server/transport/unix_jsonl.py
@@ -22,6 +22,7 @@ from server.api.protocol import (
     SubscribedMessage,
     SubscribeRequest,
 )
+from server.transport.subscriptions import SubscriptionTracker
 from vibesys.unix_socket import validate_socket_path
 
 if TYPE_CHECKING:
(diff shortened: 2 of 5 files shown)


AyanBinRafaih:
Review follow-up, fixed in e0fbc66 and a4bb341:

- **Redial race (P2)**: `wait_for_none_active` returned the instant the count hit zero, but a dropped client redials on a backoff whose first delay is 500ms, so teardown could unlink the socket underneath the redial. The wait now lingers through a settle window (`RECONNECT_SETTLE_SECONDS`, 1.0s, kept well under the launcher's 2s backend exit grace) and `track()` notifies the condition so a reconnect inside the window puts the waiter back to blocking. New test `test_disconnect_wait_bridges_a_redial_inside_the_settle_window` fails on the windowless code.
- a4bb341 moves that tracker unit test into its own module, `tests/server/test_subscription_tracker.py`: it needs no socket harness, and keeping it in `test_transport_subscription.py` created an import-block merge conflict with #601, which touches the same file. The two branches now merge cleanly in either order.

Verification: 20 targeted server tests passed; ruff, ty, and format checks clean. PR body updated for the settle-window semantics.

vic-lsh:
FYI @AyanBinRafaih: #607 follows up on the settle window added here. The 1s window is fine on its own, but the teardown path also carries the stream loop's 1.0s disconnect poll and serve_forever's default 0.5s shutdown poll, so a deliberate quit took ~2.6s and overran the launcher's 2s grace (the tui job was skipped on this PR's merge commit and first tripped on #603's). #607 trims the two polls to 0.1s each and leaves the settle window and the launcher unchanged; measured teardown is now ~1.3s.

PR #621, opened 2026-09-05
Loop: stop dropping unreviewed frontier claims from the pareto warning

### LINKED ISSUE #547: TUI: Stop dropping older unreviewed frontier claims past the eighth
### Observed behavior

The context block that warns the agent about measured frontier claims still awaiting independent review is capped to the eight most recent claims. When more than eight claims await review, the older ones silently vanish from the agent's context, and those are precisely the untrusted commits the warning exists to flag.

### Expected behavior

The agent should see every unreviewed frontier claim (or a deterministically prioritized set with the omitted count surfaced), so it never adopts an unreviewed commit as a trusted parent just because the warning scrolled off.

### Reproduction

1. Drive a run long enough to accumulate more than eight measured-but-unreviewed frontier claims.
2. Inspect the orchestrator context: only the last eight appear; earlier untrusted claims are absent.

### Affected subsystem

Optimization loop and agents

### Environment

### Impact

Blocks normal use

### Relevant logs

```shell
`src/vibesys/loops/agent/loop.py:452` uses a fixed slice `for record in pending[-8:]:` under the header at line 449 ("Measured frontier claims awaiting independent review"). The cap is an unnamed literal with no prioritization and no omitted-count disclosure.
```

#284

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #547. The Pareto archive warning could silently omit older measured but unreviewed claims after eight entries. Removing the limit entirely would let agent-facing file context grow with the run.

## Solution

Keep a named limit of eight pending claims. The archive sorts pending claims by round, shows the eight newest, and discloses both the count and round range of omitted older claims. The disclosure repeats that omitted claims are untrusted and cannot be adopted as parents.

### Architecture

The change is confined to `_pareto_archive_summary` in the agent loop. It preserves the existing trusted-versus-pending partition and writes a bounded, explicit warning through the existing Pareto archive path.

### Correctness properties

- Selection is deterministic: pending claims are ordered by round and the newest eight are rendered.
- When claims are omitted, the archive exposes their count and round range and says they remain untrusted.
- At or below the limit, every qualifying pending claim is rendered and no omission notice appears.

### Testing

- `uv run pytest tests/vibesys/loops/agent/test_orchestrate.py -q --no-cov -k pareto_archive`: 5 passed, 144 deselected, including trusted/pending partitioning and the new limit/disclosure regressions.
- `uv run ruff check src/vibesys/loops/agent/loop.py tests/vibesys/loops/agent/test_orchestrate.py`: passed.
- `uv run ruff format --check src/vibesys/loops/agent/loop.py tests/vibesys/loops/agent/test_orchestrate.py`: passed.
- `git diff --check`: passed. The full orchestrate suite exceeded the local execution window and is not counted as passing.

### DIFF
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -391,7 +391,7 @@ def _pareto_archive_summary(records: list[RoundRecord], space: MetricSpace) -> s
             "have not passed independent review, or because its numbers are the "
             "implementer's own report rather than a framework measurement:"
         )
-        for record in pending[-8:]:
+        for record in pending:
             assert record.commit is not None  # noqa: S101  # tracked: #288
             lines.append(
                 f"- round {record.round_number}, commit {record.commit[:12]}: "
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -168,6 +168,7 @@ _FAILED_HYPOTHESIS_OUTCOMES = (
     }
 ) | {"rejected"}
 _MAX_CONTINUATION_ROUNDS_WITHOUT_DESIGN_REVIEW = 2
+_PARETO_ARCHIVE_PENDING_CLAIM_LIMIT = 8
 
 
 def _persist_agent_run_state(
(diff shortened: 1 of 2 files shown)


Nano-AI:
Cross-reference: #636 targets #547 as well. Flagging it so a reviewer landing on both knows they are the same bug, not two. No changes requested here.

The difference, for whoever picks: this PR removes the cap entirely, which is the simpler read of the issue. #636 keeps a named cap and adds a disclosure line naming how many claims were omitted and their round range.

One thing worth checking before either merges: `_pareto_archive_summary` feeds `issue_board.write_pareto_archive`, so its output is agent-facing context rather than display. Removing the bound means that block grows with the run length.

AyanBinRafaih:
[a4c0111f](https://github.com/uw-syfi/vibesys/commit/a4c0111f18ca3a0707234d4d65ed2110d1a4f41f) addresses the [context-growth concern](https://github.com/uw-syfi/vibesys/pull/621#issuecomment-5605551586). The previous revision rendered every pending claim; this update uses a named limit of eight and deterministically selects the newest claims by round. When older claims are omitted, the warning reports the omitted count and round range and explicitly says those claims remain untrusted and cannot be adopted as parents. At or below the bound, all qualifying claims remain visible.

Updated regression tests cover the limit, omission disclosure, unchanged output for reversed input order, and the no-omission case. All five Pareto archive tests passed, including the existing trusted/pending partition tests. Ruff and formatting checks passed for the implementation and tests, along with `git diff --check`. GitHub checks on this commit are green.

PR #622, opened 2026-09-05
Sandbox: namespace the Modal evaluator lock and deployment lease per user

### LINKED ISSUE #548: TUI: Namespace the Modal evaluator lock and deployment lease per user
### Observed behavior

The Modal evaluator uses a fixed, host-wide lock file and deployment-lease file in `/tmp` with no per-user component. On a shared host, user A creates files owned by A, then user B's write or unlink of the same paths raises `PermissionError` and crashes, or B reads A's stale lease and targets a Modal deployment it cannot access.

### Expected behavior

Both the lock and the lease should be namespaced under a per-uid runtime directory (for example `$XDG_RUNTIME_DIR` or `gettempdir()/f"vibesys-{os.getuid()}"`), so multi-user hosts do not cross-contaminate or crash.

### Reproduction

1. On a shared host, run a Modal-evaluated task as user A (creates the lock and lease).
2. Run the same as user B; observe a `PermissionError` on the lock write/unlink, or a stale-lease mis-target.

### Affected subsystem

### Environment

### Impact

Blocks normal use

### Relevant logs

```shell
`src/vibesys/sandbox/modal_evaluator.py:48` sets `_LOCK_PATH = "/tmp/vibesys-modal-evaluator.lock"` (opened for write at line 83) and `:49` sets `_DEPLOYMENT_LEASE_PATH = Path("/tmp/vibesys-modal-evaluator-deployment.json")` (read at 174, replaced at 192-200, unlinked at 239), all fixed host-global names.
```

#284

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #548. A shared host-wide Modal lock and deployment lease let one user's files block another user's evaluations or expose a lease for a deployment the second account cannot access. The runtime directory guard also followed symlinks, allowing a pre-created symlink to pass ownership checks through its target.

## Solution

Store both coordination files under `$XDG_RUNTIME_DIR/vibesys`, falling back to `gettempdir()/vibesys-<uid>`. Before opening the lock or writing a lease, create the runtime directory with mode `0700` and inspect its entry with `lstat`. Reject symlinks, non-directories, and directories owned by another uid.

Path construction remains pure at import time. Directory creation and validation happen only at the existing lock and lease write sites. A new import regression verifies this already-present behavior, addressing the second review concern without adding import-time work.

### Architecture

The sandbox evaluator owns the runtime paths and their validation. Existing exclusive locking, cleanup, and atomic lease replacement remain unchanged. No protocol or frontend changes are involved.

### Correctness properties

- Lock and lease paths are namespaced per user.
- The runtime directory entry must be an actual directory owned by the current uid; a symlink to an owned directory is rejected.
- Importing the module neither creates nor validates the runtime directory.
- Per-user lock serialization and atomic lease writes retain their existing behavior.

### Testing

- `uv run pytest tests/vibesys/sandbox/test_modal_evaluator.py -q --no-cov`: 59 passed, including symlink rejection and an import probe tracking filesystem operations.
- `uv run ruff check src/vibesys/sandbox/modal_evaluator.py tests/vibesys/sandbox/test_modal_evaluator.py`: passed.
- `uv run ruff format --check src/vibesys/sandbox/modal_evaluator.py tests/vibesys/sandbox/test_modal_evaluator.py`: passed.
- `git diff --check`: passed. No live Modal deployment was performed for this revision.

Closes #548.

### DIFF
--- a/src/vibesys/sandbox/modal_evaluator.py
+++ b/src/vibesys/sandbox/modal_evaluator.py
@@ -32,6 +32,7 @@ import shlex
 import subprocess
 import sys
 import tarfile
+import tempfile
 import threading
 import time
 import urllib.error
@@ -47,8 +48,6 @@ _MODAL_DEPLOYMENT = re.compile(
     r"https://modal\.com/apps/[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+/deployed/"
     r"([a-zA-Z0-9_.-]+)"
 )
-_LOCK_PATH = "/tmp/vibesys-modal-evaluator.lock"  # noqa: S108  # tracked: #288
-_DEPLOYMENT_LEASE_PATH = Path("/tmp/vibesys-modal-evaluator-deployment.json")  # noqa: S108  # tracked: #288
 _CANDIDATE_REVISION_ENV = "VIBESYS_CANDIDATE_REVISION"
 _RELEASE_DEPLOYMENT_ENV = "VIBESYS_RELEASE_MODAL_DEPLOYMENT"
 _MAX_DIAGNOSTIC_CHARS = 20_000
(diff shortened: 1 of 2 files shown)


Nano-AI:
Cross-reference: #638 targets #548 as well, with the same design (XDG_RUNTIME_DIR, else a uid-suffixed dir under the system temp dir). Flagging it so a reviewer landing on both knows they are the same bug. No changes requested here.

Two differences that may be worth folding in regardless of which one lands:

1. `_ensure_runtime_dir` uses `stat`, which follows a symlink and then checks the target uid. `lstat` inspects the entry itself, so a symlink planted at the runtime-dir path is rejected rather than followed. That is the specific pre-creation attack the `S108` suppression stood in for.
2. Computing the paths at module scope is fine; the `mkdir` and the ownership check at import time are the risk. `sandbox/run_environment.py:58` imports this module eagerly just for `encode_setup_command`, so a read-only temp dir or a hostile entry at that path would raise during an import that never touches Modal. Before the change these were inert string constants and import was side-effect free. #638 keeps the path helper as pure path arithmetic and moves the filesystem work into a `functools.cache`-wrapped function called at the point of use.

AyanBinRafaih:
[c8b62428](https://github.com/uw-syfi/vibesys/commit/c8b62428d4f4c849a1c260baee474c2357a32566) addresses the [review finding](https://github.com/uw-syfi/vibesys/pull/622#issuecomment-5605553318) that a pre-created symlink could pass runtime-directory ownership checks. The evaluator now uses `lstat` and rejects symlinked entries, non-directories, and directories owned by another uid before opening the lock or writing a lease. Runtime paths remain per-user and directory creation and validation stay at the existing operation sites.

Confirmed that runtime-directory creation and validation were already deferred until use at the prior revision. Added a regression test to preserve that import behavior, alongside a symlink-rejection regression. The Modal evaluator suite passed 59 tests, including both new cases. Ruff, formatting checks, and `git diff --check` passed. GitHub checks on this commit are green.

Issue #617, opened 2026-09-04
TUI dev harness: scratch cleanup races the mock server's exit and fails with EBUSY on NFS-backed tmpdirs


AyanBinRafaih:
### Observed behavior

On a host whose `/tmp` is NFS-backed, `pnpm test:clients` fails in `clients/tui/dev/harness.test.ts` with `EBUSY: resource busy or locked, rm '/tmp/vs-mock-...'` raised from the scratch-directory cleanup (`harness.test.ts:58-64`). Which case fails varies across runs (observed on "replays a legacy capture with the executions the read path translates into it" and on the `bad-cpp-round1` replay), and each failure leaves an undeletable `/tmp/vs-mock-*` directory behind; they accumulate across runs.

The root cause is a teardown ordering race, not host misbehavior. `afterEach` runs the `cleanups` stack LIFO, so `server.kill('SIGKILL')` (pushed at `harness.test.ts:138`) does run before `rmSync(directory, {recursive: true, force: true})` (pushed at `harness.test.ts:63`), but the kill is fire-and-forget: nothing awaits the mock server's exit, so the process can still hold its journal file and socket open inside the scratch directory when `rmSync` executes. On a local tmpfs, unlinking open files succeeds and the race is invisible. On NFS, the client silly-renames open files to `.nfsXXXX` instead of unlinking them, the directory is left non-empty, and removing it fails with `EBUSY`.

### Expected behavior

Teardown waits for the mock server process to actually exit before removing the scratch directory (or retries the removal until the process has released its files), so the harness suite passes regardless of whether the tmpdir filesystem allows unlinking open files. No `/tmp/vs-mock-*` directories are left behind.

### Reproduction

1. On a machine where `os.tmpdir()` resolves to an NFS mount (or any filesystem that silly-renames open files), run `pnpm test:clients`.
2. Observe one or more `dev/harness.test.ts` cases fail with `error: EBUSY: resource busy or locked, rm '/tmp/vs-mock-...'` at the cleanup in `harness.test.ts:58` while all assertions before teardown passed.
3. Observe leftover `/tmp/vs-mock-legacy-*` / `/tmp/vs-mock-*` directories after the run.

### Affected subsystem

CLI and packaging

### Environment

Commit: 12e807a (current `main`)
OS: Linux 6.1 (Debian), shared multi-user host with NFS-backed `/tmp`
Hardware: x86_64 build server
Backend: not involved (dev mock harness only)
Profiler: not involved
[...]

PR #624, opened 2026-09-05
TUI: redraw the running round in its rendered width inside the compact rail

### LINKED ISSUE #551: TUI: Redraw the running round in its true width inside the compact rail
### Observed behavior

On a mid-width terminal (85 to 99 columns) with a live run, once per second the active round's row in the rounds rail is regenerated in full width (for example `▸r7 ⟳ run 1m 5s`) even though the rail was drawn in its compact 13-column form. The row has no truncation, so every second it overflows or wraps, corrupting the fixed one-row-per-round layout the rail depends on for its overflow counters.

### Expected behavior

The per-second elapsed refresh should redraw the running round in whatever width mode the rail is currently in, so the compact rail stays compact and the overflow math stays correct.

### Reproduction

1. `runtui.sh start 92x40 spsc` and let a round go active.
2. Watch the active round's row; once per second it re-expands to full width and breaks the compact column.

### Affected subsystem

CLI and packaging

### Environment

### Impact

Blocks normal use

### Relevant logs

```shell
`clients/tui/src/ui/round-rail.ts` renders with the correct `compact` flag at line 256, but the `setInterval` refresh callback (around line 305, inside `#syncElapsedTimer`) calls `this.#roundLabel(round, state, ..., false)` with the compact argument hardcoded `false`, ignoring the rail's actual mode.
```

#284

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #551. On a mid-width terminal (85 to 99 columns) the rounds rail renders in its compact 13-column form, but the per-second elapsed-time refresh redrew the active round's row at full width: the `setInterval` callback in `RoundRailView#syncElapsedTimer` (`clients/tui/src/ui/round-rail.ts`) called `#roundLabel(round, state, ..., false)` with the compact argument hardcoded to `false`, while the initial render at line 256 passed the rail's real mode. The regenerated row (for example `▸r7 ⟳ run 1m 5s`) carries no truncation, so once per second it overflowed or wrapped the compact column, corrupting the fixed one-row-per-round layout the rail's overflow counters depend on.

## Solution

The timer now redraws in the same width mode the row was rendered in. `#runningRound`, the memo the timer reads (round, state, and the `TextRenderable` it rewrites), gains a `compact` field captured at render time in `#roundLabelText`, and the interval callback passes that captured flag to `#roundLabel` instead of the literal `false`. This is the root-cause shape of the fix: the refresh path had access to everything about the running round except the mode it was drawn in, so the mode joins the memo rather than being re-derived or plumbed through a side channel. A rail re-render on resize rebuilds `#runningRound` with the new mode, so the timer always agrees with the layout that is actually on screen, in both directions of a width change.

### Architecture

The change is confined to the `@vibesys/tui` package's rail widget; no core-state, backend-client, protocol, or Python surface is involved, and the fold that produces `activeAgentStarts` is untouched.

```mermaid
flowchart TD
    S["SessionState fold (core-state): activeAgentStarts on the running round"] --> R["RoundRailView.render(state, width, rows)"]
    R --> L["#roundLabelText: draws the row, captures {round, state, text, compact} in #runningRound"]
    L --> T["#syncElapsedTimer: 1s interval rewrites text.content"]
    T --> C["#roundLabel(round, state, visible, compact) with the captured mode (was hardcoded false)"]
    C --> V["compact rail stays 13 columns; full rail keeps its elapsed suffix"]
```

## Verification

Two regression tests pin the refresh behavior in both width modes; the compact one fails on the unfixed code and passes on this branch. The full TS check suite passes at the current `main` tip, and a live codex-provider run at 92x40 holds the compact rail stable across elapsed ticks.

### Correctness properties

- The elapsed-time refresh is idempotent with respect to layout: rewriting the running round's row never changes its width mode, so a compact rail stays within its 13-column budget on every tick and the one-row-per-round invariant behind the overflow counters holds.
- The full-width rail keeps its elapsed suffix (`run Ns`) across ticks; the fix does not dim or truncate the wide form.
- The mode the timer uses is exactly the mode of the last real render: a resize that flips the rail between compact and full rebuilds the memo, so the timer can never redraw in a stale mode in either direction.
- Timer lifecycle (armed only while a running round has an active agent timing, cleared on destroy) is unchanged; only the label call inside the tick changed.

### Testing

- New regression tests in `clients/tui/src/ui/round-rail.test.ts`: "keeps the compact label after the elapsed timer refreshes at a compact width" and "keeps the elapsed suffix after the timer refreshes at full width". They render a live round through a real `createTestRenderer`, wait past a 1-second tick, and read back the rewritten row. Before the fix (source reverted via a saved patch, never `git stash`): 1 failed with `Expected to not contain: " run " / Received: "▸r1 ⟳ run 1s"` (26 pass, 1 fail). After: 27 pass, 0 fail.
- `pnpm check:ts` (Checked 85 files), `pnpm check:ts-architecture` (no dependency violations, 90 modules), `pnpm check:clients`, `pnpm build:clients`: all pass at the current `main` tip (12e807a).
- `pnpm test:clients`: clean run, 686 pass, 0 fail. An intermittent `EBUSY: resource busy or locked, rm '/tmp/vs-mock-*'` in `clients/tui/dev/harness.test.ts` reproduces on this NFS-backed host with the change fully reverted and is tracked separately as #617; it is environmental, not from this change.
- Live harness run (codex provider, model `gpt-5.6-sol`, queue-rs spsc, `--profiler none`) at 92x40: with a round active in the round view, three captures two seconds apart (the pane's elapsed chip advances 7m 15s to 7m 18s, so timer ticks landed between frames) all show the compact rail row as `▸r1⟳` inside its intact 13-column box, never re-expanded to the full `run Ns` form.
- Merges cleanly against current `main` (`git merge-tree --write-tree`).

Closes #551.

### DIFF
--- a/clients/tui/src/ui/round-rail.test.ts
+++ b/clients/tui/src/ui/round-rail.test.ts
@@ -234,6 +234,64 @@ describe('RoundRailView row budget', () => {
   });
 });
 
+describe('RoundRailView elapsed timer refresh', () => {
+  /** A single active round with a live agent, so the rail arms the elapsed timer. */
+  function activeRoundState(): SessionState {
+    const base = railState(1);
+    return {
+      ...base,
+      selectedRound: 1,
+      core: {
+        ...base.core,
+        rounds: [
+          {
+            number: 1,
+            status: 'active',
+            startedAt: new Date().toISOString(),
+            activeAgentStarts: {worker: new Date().toISOString()},
+          },
+        ],
+      },
+    };
+  }
+
+  function textOf(text: TextRenderable): string {
+    const content = (text.content as {chunks?: {text?: string}[]} | undefined)?.chunks ?? [];
+    return content.map(chunk => chunk.text ?? '').join('');
+  }
+
+  test('keeps the compact label after the elapsed timer refreshes at a compact width', async () => {
+    const {renderer} = await createTestRenderer({width: 120, height: 40});
+    const view = new RoundRailView(
+      renderer,
+      {} as unknown as SessionController,
+      resolveTheme(null),
+    );
+    view.render(activeRoundState(), RAIL_COMPACT_WIDTH, 10);
+    // The elapsed timer ticks on a real one-second interval; wait past a tick so the
+    // refresh runs, then read the row it rewrote.
+    await new Promise(resolve => setTimeout(resolve, 1100));
+    const text = textOf(view.output.getChildren()[0] as TextRenderable);
+    view.destroy();
+    expect(text).not.toContain(' run ');
+    expect(text.length).toBeLessThanOrEqual(RAIL_COMPACT_WIDTH);
+  });
+
+  test('keeps the elapsed suffix after the timer refreshes at full width', async () => {
+    const {renderer} = await createTestRenderer({width: 120, height: 40});
+    const view = new RoundRailView(
+      renderer,
+      {} as unknown as SessionController,
+      resolveTheme(null),
+    );
+    view.render(activeRoundState(), RAIL_FULL_WIDTH, 10);
+    await new Promise(resolve => setTimeout(resolve, 1100));
+    const text = textOf(view.output.getChildren()[0] as TextRenderable);
+    view.destroy();
+    expect(text).toContain(' run ');
+  });
+});
+
 describe('RoundRailView profile-skipped rounds', () => {
   test('marks a completed profile-skipped round hollow and dim', async () => {
     const base = railState(3);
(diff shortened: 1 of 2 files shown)


Nano-AI:
Cross-reference: #637 targets #551 as well. Flagging it so a reviewer landing on both knows they are the same bug, not two. No changes requested here.

Of the two, this PR looks like the better approach: storing `compact` on `#runningRound` replays exactly the flag the row was drawn with. #637 recomputes it from `#renderedWidth` at refresh time, which would disagree if the width ever changed without a re-render. It threads no new field, which is the only advantage it has.

PR #620, opened 2026-09-05
TUI: render readable tool cards from typed payloads and demote lifecycle chatter

AyanBinRafaih:
## Problem

Fixes #516. Round transcripts showed raw tool argument JSON, shell wrappers with confusing nested quoting, and verbose results. Provider lifecycle messages occupied bordered cards and resembled failures. Codex also emitted duplicate, unanswered tool calls: AgentShim 0.5.x maps generic item starts and completions to `ToolUseEvent`.

The original callback-level deduplication lost legitimate actions. Two consecutive `file_change` items with different IDs and identical arguments produced only one journal call. MCP completion payloads also change when results or errors arrive, so argument equality cannot identify an item's lifecycle.

## Solution

Tool previews dispatch by tool and typed result payload. Shell previews remove recognized shell wrappers while preserving command quoting; file changes summarize actions and paths; unknown tools retain the argument fallback with whole-value shape summaries for large structured arguments. Command results show exit status, duration, and size when collapsed, and separated stdout/stderr when expanded. JSON results show a shape summary or formatted contents. The expand/collapse contract and raw-content fallback remain intact.

Non-failure diagnostics and status messages render as bare rows. Failure markers retain prominent cards, with backend tone taking precedence. Tool and output text wrap. The callback emitter classifies driver lifecycle markers as `diagnostic`; core-state joins diagnostic chunks with newlines.

Codex event repair now lives in `src/vibesys/_agent_cli/codex.py`, where raw lifecycle kinds and item IDs remain available. The session registers generic starts in AgentShim's item maps and converts completion echoes into `ToolResultEvent`. Original arguments stay on the call; completion output and errors appear in the result. The neutral driver forwards every callback without argument-signature deduplication. This PR includes the same adapter and adapter tests as #628, so either PR works independently.

### Architecture

```mermaid
flowchart TD
    A[Codex JSON lifecycle and item ID] --> B[Codex session adapter: pair calls and results by ID]
    B --> C[Neutral driver events]
    C --> D[Application logger and journal]
    D --> P[Tool and result previews]
    E[Callback lifecycle classification] --> F[Core-state diagnostic folding]
    F --> G[TUI diagnostic and failure styling]
    P --> V[Conversation rendering and wrapping]
    G --> V
```

The Codex adapter owns provider lifecycle repair, callbacks own channel semantics, core-state owns event folding, and the TUI owns presentation. No wire schema or installed-package changes are required.

### Correctness properties

[...]

PR #628, opened 2026-09-05
Agents: pair codex tool calls with results in the driver

### LINKED ISSUE #618: Agent driver: agentshim 0.5.0 codex adapter emits a duplicate ToolUseEvent per non-command item, doubling tool_call events
### Observed behavior

On codex-provider runs, every non-command agent action (`file_change`, `todo_list`, `mcp_tool_call`, `web_search`) reaches VibeSys twice. The installed `agentshim` 0.5.0 codex event adapter (`agentshim/codex/events.py:70-82`) returns a byte-identical `ToolUseEvent` for both the `item.started` and `item.completed` lifecycle notifications on every item type except `command_execution`: a `status` local is computed and then unused, and the two arms construct the same event. `agentshim/codex/agent.py:89` then fires `on_tool_call` once per `ToolUseEvent`. VibeSys mints a fresh `call_id` per callback, so one action becomes two `tool_call` wire events with distinct ids in `run-events.jsonl`, and every consumer shows the action twice: the TUI transcript, the headless renderer, replay, and the run log. Because `file_change` never yields a tool result, both transcript cards additionally stay open forever (one visible symptom of #516).

### Expected behavior

One agent action produces exactly one `on_tool_call` callback and therefore one `tool_call` event. The adapter should emit the `ToolUseEvent` on a single lifecycle edge for generic item types (for example return `None` for `item.started` and keep the `item.completed` emission), matching the split handling it already has for `command_execution`. Once agentshim is fixed and the dependency floor bumped, the codex-scoped dedup that #516's fix adds in `src/vibesys/agents/drivers/agentshim.py` (`_AgentShimEventHandler` drops a call identical to the immediately preceding unanswered call) should be removed.

### Reproduction

1. Run any codex-provider round that performs a `file_change`/`apply_patch` (queue-rs `spsc`, 1 round is enough).
2. Inspect `run-events.jsonl`: each such action appears as two consecutive `tool_call` events identical in `tool` and `args` but with different call ids; the TUI round transcript renders two duplicate cards.
3. The defect is also pinned without a live run by `tests/vibesys/agents/drivers/test_agentshim_driver.py::test_codex_item_lifecycle_does_not_open_two_turns_for_one_action` (added with #516's fix), which replays the started/completed pair a real codex item produces.

### Affected subsystem

Optimization loop and agents

### Environment

Commit: 12e807a (current `main`)
agentshim: 0.5.0 from PyPI (`pyproject.toml` pins `agentshim>=0.5.0`)
Provider: codex CLI, model `gpt-5.6-sol`
OS: Linux 6.1 (Debian), x86_64

### Impact

Workaround exists

### Relevant logs

```shell
# run-events.jsonl shape produced by one file_change action (ids differ, everything else identical):
{"type": "tool_call", "call_id": "<uuid-A>", "tool": "file_change", "args": {"changes": [{"path": "src/lib.rs", "kind": "delete"}]}, ...}
{"type": "tool_call", "call_id": "<uuid-B>", "tool": "file_change", "args": {"changes": [{"path": "src/lib.rs", "kind": "delete"}]}, ...}
```

#516

### Pre-submission checks

AyanBinRafaih:
## Problem

Fixes #618. AgentShim 0.5.x reports generic Codex item starts and completions as tool calls, leaving duplicate calls without results in the journal. The original argument-equality workaround also failed when MCP completions populated `result` or `error`, and could confuse distinct items with identical arguments.

## Solution

Pair events inside the existing Codex CLI adapter while raw `item.started` / `item.completed` kinds and stable item IDs are available. `_PairedCodexSession` registers generic starts in AgentShim's existing item maps, converts generic completion echoes into `ToolResultEvent`, and delegates callback emission and duration calculation to the base session. Completed generic entries are removed from the maps.

The original start arguments remain on the call; completion output and errors become result text. Completion-only items receive a synthetic call followed by their result. Command execution and other callbacks retain AgentShim's existing handling. A parser that already supplies `ToolResultEvent` also follows the same registered-ID path.

### Architecture

```mermaid
flowchart TD
    A[Codex JSON item lifecycle and ID] --> B[VibeSys Codex session adapter]
    B --> C[AgentShim ToolUseEvent / ToolResultEvent paired by ID]
    C --> D[Neutral driver events]
    D --> E[Application logger, journal, and clients]
```

The compatibility code stays in `src/vibesys/_agent_cli/codex.py`; installed packages and the wire schema are unchanged. PR #620 includes the identical adapter and tests, so either PR works independently.

### Correctness properties

- Changed completion payloads produce one call with the original arguments and one result containing the completion output or error.
- Consecutive identical items, overlapping starts, and completion-only items remain separate by item ID.
- Command executions and native result events retain their existing call/result handling.
- Generic state belongs to one session and is removed after completion.

### Testing

- Before the fix, the MCP changed-result, interleaved-item, and completion-only regressions failed against `1aa055cb` through the installed parser and real `_process_stdout` path.
- `python -m pytest tests/vibesys/_agent_cli/test_codex.py tests/vibesys/agents/drivers/test_agentshim_driver.py tests/vibesys/agents/test_callbacks.py -q --no-cov`: 138 passed, using the existing Python environment with this worktree's source roots on `PYTHONPATH`.
- Changed-file `ruff check`, `ruff format --check`, and `ty check`: passed. `git diff --check`: passed.
- A broader `_agent_cli` plus `agents` run stalled and was interrupted. It is not counted as passing. No new live provider run was performed for this revision.

Closes #618.

### DIFF
--- a/src/vibesys/_agent_cli/codex.py
+++ b/src/vibesys/_agent_cli/codex.py
@@ -1,5 +1,6 @@
+import time
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from agentshim.codex import CodexGenerationSession
 from agentshim.events import AgentEventHandler
@@ -9,6 +10,55 @@ from .cli_agent import CLICodingAgent
 
 if TYPE_CHECKING:
     from agentshim.executor import CommandExecutor
+
+_COMMAND_TOOL = "execute"
+
+
+class _PairedCodexToolEvents:
+    """Repair the codex adapter's tool-event stream into call/result pairs.
+
+    agentshim through 0.5.1 maps a generic codex item's ``item.started`` and
+    ``item.completed`` both to tool-use, so one ``file_change`` (or
+    ``web_search``, ``mcp_tool_call``, ...) reaches ``on_tool_call`` twice
+    with identical arguments and never reaches ``on_tool_result``. This
+    wrapper treats the second identical call for a still-open item as that
+    item's completion: it swallows the duplicate and emits the missing
+    ``on_tool_result`` instead. Command executions (``execute``) already
+    arrive correctly paired and pass through untouched, as does every other
+    callback. Under an adapter that reports completions as results, no
+    duplicate ever arrives and the wrapper forwards everything unchanged.
+    """
+
+    def __init__(self, inner: AgentEventHandler) -> None:
+        self._inner = inner
+        # One open (args, start time) per tool name: codex items in a turn
+        # are sequential, so a second identical call while the first is open
+        # can only be the same item's completion echo.
+        self._open: dict[str, tuple[Any, float]] = {}
+
+    def __getattr__(self, name: str) -> Any:  # noqa: ANN401  # tracked: #288
+        return getattr(self._inner, name)
+
+    def on_tool_call(self, tool: str, args: Any = None) -> None:  # noqa: ANN401  # tracked: #288
+        """Forward a tool call, folding a completion echo into its result."""
+        if tool != _COMMAND_TOOL:
+            opened = self._open.pop(tool, None)
+            if opened is not None and opened[0] == args:
+                self._inner.on_tool_result(
+                    tool=tool,
+                    stdout="",
+                    exit_code=None,
+                    duration=time.monotonic() - opened[1],
+                )
+                return
+            self._open[tool] = (args, time.monotonic())
+        self._inner.on_tool_call(tool, args)
+
+    def on_tool_result(self, *args: Any, **kwargs: Any) -> None:  # noqa: ANN401  # tracked: #288
+        """Forward a real result verbatim and close the matching open call."""
+        tool = kwargs.get("tool", args[0] if args else None)
+        self._open.pop(tool, None)
+        self._inner.on_tool_result(*args, **kwargs)
 
 
 def _toml_str(value: str) -> str:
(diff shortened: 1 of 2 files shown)


AyanBinRafaih:
Review finding at `1aa055cb`:

**[P2] Pair events by item identity before completion fields change** ([codex.py:45-54](https://github.com/uw-syfi/vibesys/blob/1aa055cbeb5b02662d191c92a7acba6031c96d92/src/vibesys/_agent_cli/codex.py#L45)).

Exact argument equality does not hold for normal Codex lifecycle events. [Codex's JSON-output test](https://github.com/openai/codex/blob/main/codex-rs/exec/tests/event_processor_with_json_output.rs#L3052-L3224) emits an MCP start with no result and the same item's completion with a populated result. AgentShim preserves those fields while dropping the stable item ID, so this branch removes the pending start, treats the completion as another call, and never emits a result. Replaying that MCP sequence through this head's real `_process_stdout` produced two `on_tool_call` callbacks and zero `on_tool_result` callbacks, leaving the duplicate/unanswered-call bug in place for MCP. Please pair at the Codex lifecycle layer using item identity and start/completion events, and test completion payloads that differ from their starts.

Validation: all 42 existing Codex tests passed; the additional MCP lifecycle reproduction exposed the missing pairing.

AyanBinRafaih:
[f46b36f0](https://github.com/uw-syfi/vibesys/commit/f46b36f0979720145ec54b201e85f7dceb987b02) addresses the [review finding](https://github.com/uw-syfi/vibesys/pull/628#issuecomment-5591555566) that generic Codex lifecycle events could create duplicate calls without results, especially when MCP completion payloads changed. The adapter now pairs starts and completions by lifecycle item identity, preserving original call arguments and forwarding completion output or errors as the matching result. Consecutive identical items, overlapping starts, and completion-only items remain distinct. The same adapter and tests are included independently in #620.

The targeted Python suite passed 138 tests, covering changed MCP result and error payloads, duplicate-looking items, completion-only items, and existing command-result handling. Ruff, formatting checks, Ty, and `git diff --check` passed. GitHub checks on this commit are green.

PR #629, opened 2026-09-05
feat: dataflow_opt modality for in-place engine superoptimization

HQingXuan:
## Problem

The differential-dataflow optimization workload benefits from retaining a measured component focus across rounds. The ordinary agent loop has profiling support, but no policy for walking ranked dataflow bottlenecks or preserving that focus through retries and subsequent hypotheses.

The runnable differential-dataflow bundle was merged separately in #739. This PR integrates against its canonical `examples/database/differential-dataflow` path instead of carrying another copy of the example.

## Solution

Add a `dataflow_opt` modality to the existing agent outer loop.

- Register the modality in the headless CLI.
- Add modality-specific orchestrator, implementer, judge, and profiler prompt fragments.
- Add typed ranked-bottleneck fields to the agent response schemas.
- Persist a bottleneck ledger that keeps one active component until its measured plateau is reached.
- Connect the ledger to the existing agent round lifecycle without changing other modalities.
- Point the modality tests and profiler fixture at the canonical differential-dataflow example merged by #739.

### Architecture

The existing `agent` outer loop remains the lifecycle owner. Selecting `dataflow_opt` activates its prompt fragments and bottleneck-ledger policy. The example bundle and profiler remain under `examples/database/differential-dataflow`; this PR does not duplicate those inputs.

## Verification

The focused modality, ledger, and attribution tests cover registration, guarded prompts, state persistence, plateau behavior, and profiler parsing. GitHub CI is rerunning after the stacked-PR boundary change.

### Correctness properties

- Existing agent modalities render and run unchanged when `dataflow_opt` is not selected.
- One active component persists until its measured plateau is reached.
- Failed or unmeasured attempts do not advance the bottleneck walk.
- The implementation consumes the canonical example path merged by #739.

### Testing

- `tests/vibesys/loops/agent/test_dataflow_opt_modality.py`
- `tests/vibesys/loops/agent/test_bottleneck_ledger.py`
- `tests/vibesys/loops/agent/test_attribute_cpu.py`
- GitHub Actions test, lint, type-check, sandbox, and example-validation workflows


Nano-AI:
> [!NOTE]

**Summary:** Adds a `dataflow_opt` modality to the `database` domain. It includes a framework-owned bottleneck ledger that steers each round from callgrind attribution, and a `differential-dataflow-cpu-bench` example bundle with equivalence, fuzz, determinism, crash-recovery, and TSan gates.

**Findings**

1. **[P1] The attribution command omits the required `--engine-cmd`, so the bottleneck walk never engages** (`src/vibesys/loops/agent/bottleneck_ledger.py:114-134`, `examples/differential-dataflow-cpu-bench/profiler/attribute_cpu.py:282-287`)
   `run_attribution` runs `uv run python profiler/attribute_cpu.py --output-json <path>`. `attribute_cpu.py` declares `--engine-cmd` with `required=True` and no default (`checker.py` and `benchmark.py` default it), so argparse exits 2 on every round. `run_attribution` logs `attribution exited 2` and returns `None`. The ledger then stays empty and `select_active_component` returns `None`, so `active_component`, `ledger_text`, and `ranked_bottlenecks` are empty. Every `{% if %}`-guarded walk block drops out and `advance_after_round` is skipped (`loop.py:2353-2376`, `3468`). A `dataflow_opt` run therefore behaves like an unsteered run, while `profiler.j2` tells the agent the framework selects its focus. The `run_attribution` tests use a fake `judge_backend.execute` with canned output, and `test_attribute_cpu.py` never calls `main()`, so no test covers the real CLI contract.
   Evidence: running the ledger's exact command in the bundle tree printed `error: the following arguments are required: --engine-cmd`, with exit 2.
   Suggested direction: default `--engine-cmd` to `engine/target/release/examples/bfs` like the sibling scripts, and add a test that runs the real script's argument parser.

2. **[P2] The accuracy command exits 0 when `_ref_engine/` is missing, which skips output equivalence** (`examples/differential-dataflow-cpu-bench/accuracy_checker/checker.py:166-174`, `equivalence_gate.py:125-128`, `differential_fuzz_gate.py:156-159`)
   When `_ref_engine/` is absent, both `equivalence` and `differential-fuzz` return SETUP-ERROR (exit 2). The same happens when the pristine build fails with no cached binary. `checker.py` maps SETUP-ERROR to exit 0 unless `--strict` is passed, and the manifest's `[accuracy] command` (`vibesys.input.toml:7`) does not pass it. The framework accuracy gate passes on exit 0, so the implementer can drop the only correctness oracle and leave just the LLM judge's `diff -ru` instruction as enforcement. The PR discloses that `_ref_engine` has no mechanical tamper protection. It does not disclose that removing the oracle turns the equivalence gates into passing skips.
   Evidence: with the bundle extracted without `_ref_engine/`, this command exited 0: `python3 accuracy_checker/checker.py --no-build --engine-cmd <script that prints one tuple>`. Summary: equivalence SETUP-ERROR, differential-fuzz SETUP-ERROR, determinism PASS, crash-recovery PASS, sanitizer SETUP-ERROR, then `ALL RUNNABLE CHECKS PASSED`.
   Suggested direction: treat pristine-side setup errors in `equivalence` and `differential-fuzz` as failures (a missing oracle is not an environment limitation). Keep the SETUP tolerance for the sanitizer only.

3. **[P3] The framework hardcodes the example's script path and argv instead of taking them from the manifest** (`src/vibesys/loops/agent/bottleneck_ledger.py:35`, `:114-145`, `src/vibesys/loops/agent/loop.py:2353`)
   `ATTRIBUTION_SCRIPT_REL = "profiler/attribute_cpu.py"` and the `uv run python ... --output-json` invocation are example-bundle policy inside `src/vibesys/`. `loop.py` also gains its first branch on a concrete modality name (`modality == "dataflow_opt"`). Finding 1 follows from this: the framework builds an argv for a script it does not own. The command construction and BEGIN/END marker extraction also duplicate `run_benchmark_gate` (`src/vibesys/loops/gates.py:405-411`, `458-459`). The review guide keeps example-specific behavior in the bundle, and `coding-best-practices.md` lists "scattered subprocess command construction and output parsing" under Avoid.
   Suggested direction: declare the attribution command in `vibesys.input.toml`, like `[accuracy]` and `[benchmark]` (a command plus a JSON-result argument). The loop would then enable the walk when that command is present and reuse the existing marker transport.

4. **[P3] New schema fields are either never read or reference data the agent cannot see** (`src/vibesys/schemas.py:441-461`, `:496-503`, `src/vibesys/prompts/loops/agent/orchestrator_pre_round_prompt.j2:68-75`)
   Nothing constructs `ProfilerSummary.ranked_bottlenecks` outside a test, and the ledger reads raw attribution dicts. `DATABASE` also resolves to `ProfilerKind.NONE`, so no profiler runs for `dataflow_opt`. The field is still part of `ProfilerSummary`'s response schema, which is sent to the profiler agent in every domain (`agents/cli_common.py` uses `model_json_schema()`). This is an agent-visible change outside `dataflow_opt`. Separately, the pre-round prompt asks for an override "from the ledger in `progress.md`", but nothing writes the ledger there. `ledger_text` only reaches the plan prompt, which runs after the pre-round decision.
   Suggested direction: remove `RankedBottleneck`/`ranked_bottlenecks` until a producer exists, and either write the ledger into the progress file or point the pre-round prompt at data it actually receives.

vic-lsh:
@HQingXuan thanks for opening this. I'm taking a look now

vic-lsh:
@HQingXuan FYI, I split the runnable differential-dataflow example into #739 so the workload, oracle, gates, benchmark, and manual profiler can be reviewed and merged independently of the proposed `dataflow_opt` core integration. The split also makes required reference failures fail closed and rejects abnormal TSan exits.

vic-lsh:
@HQingXuan I moved the complete outer-loop conversion into stacked PR #764. PR #629 is now the original modality implementation against the canonical differential-dataflow example from #739; #764 shows the modality-to-profile-guided refactor and its follow-up resume/recovery cleanup.

HQingXuan:
Hi @vic-lsh thanks for the review and the refactor. With the refactor you made in [764](https://github.com/uw-syfi/vibesys/pull/764), should I still keep this PR open or should I close it?

vic-lsh:
Closing: the dataflow_opt modality is replaced by the profile-guided outer loop in #764, which lands with this PR's commits squashed in. Thanks @HQingXuan for the original work.

PR #631, opened 2026-09-06
agent_runner: recover typed replies behind inline braces and raw newlines

kwanUm:
## Summary

`parse_typed_response_text` sliced the model text from the first `{` to the last `}`, so a reply preceded by narrative that itself contains braces (a knob value like `{"cudagraph_mode": "FULL_DECODE_ONLY"}` is common in implementer output) never parsed and the round fell back to the failure reply. It now tries every balanced top-level object in order.

Models that emit literal newlines or tabs inside JSON string values (Kimi K3 through OpenRouter does) failed with "Unterminated string"; each candidate is retried with those control characters escaped, leaving everything outside string values untouched.

Both shapes were seen repeatedly on real campaigns.

## Test plan

- [x] `tests/vibesys/test_agent_runner_parse.py` (new): plain/fenced still parse, reply behind inline braces, raw newline repair, escaped quotes and backslashes survive, nothing parseable returns None
- [x] `uv run pytest tests/vibesys/test_agent_runner_parse.py tests/vibesys/agents/test_agent_runners.py`: 83 passed; ruff clean

### DIFF
--- a/src/vibesys/agent_runner.py
+++ b/src/vibesys/agent_runner.py
@@ -216,22 +216,71 @@ def parse_typed_response_text(text: str, response_cls: type[T]) -> T | None:
     fenced_matches = re.findall(r"```(?:json)?\s*(\{.*?\})\s*```", text, flags=re.DOTALL)
     candidates.extend(match.strip() for match in fenced_matches if match.strip())
 
-    start = text.find("{")
-    end = text.rfind("}")
-    if start != -1 and end != -1 and end > start:
-        candidates.append(text[start : end + 1].strip())
+    # Every balanced top-level ``{...}`` object, in order. Narrative text often
+    # carries inline braces (``{"cudagraph_mode": "FULL"}``) before the real
+    # reply, so a single first-``{``-to-last-``}`` slice would fail to parse.
+    candidates.extend(_balanced_objects(text))
 
     seen: set[str] = set()
     for candidate in candidates:
         if not candidate or candidate in seen:
             continue
         seen.add(candidate)
-        try:
-            payload = json.loads(candidate)
-            return response_cls.model_validate(payload)
-        except (json.JSONDecodeError, ValidationError, TypeError):
-            continue
+        for attempt in (candidate, _escape_control_chars_in_strings(candidate)):
+            try:
+                payload = json.loads(attempt)
+                return response_cls.model_validate(payload)
+            except (json.JSONDecodeError, ValidationError, TypeError):
+                continue
     return None
+
+
+def _balanced_objects(text: str) -> list[str]:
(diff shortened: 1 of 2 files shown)


vic-lsh:
Review at `48aad08a`. Thanks for this; both failure shapes are real and the new tests demonstrate them (3 of 5 fail at the merge base, all pass here).

**[P1] Brace scan counts braces inside string values** ([agent_runner.py:240-254](https://github.com/uw-syfi/vibesys/blob/48aad08afb33bfb3e5f7179a6c828945e3154b93/src/vibesys/agent_runner.py#L240-L254))

`_balanced_objects` tracks depth on every `{` and `}`, including ones inside JSON strings. Implementer and judge replies routinely quote code in `summary` or `analysis`, so a value like `"added fn main() {"` leaves the scan unbalanced and the reply falls to the failure response. Two probes that parse on `main` and return `None` here:

```
Done. Final:\n{"analysis": "added fn main() {", "verdict": "pass"}
Done. Final:\n{"analysis": "closing } brace", "verdict": "pass"}
```

Skip braces while inside a string (the same in-string and escape tracking the control-character pass already does), or keep the old first-to-last slice as a final fallback candidate. Please add a regression test with a brace inside a string value.

**[P2] `_parse_profiler_response_text` still uses the old slice** ([agent_runner.py:520-545](https://github.com/uw-syfi/vibesys/blob/48aad08afb33bfb3e5f7179a6c828945e3154b93/src/vibesys/agent_runner.py#L520-L545))

It is a copy of the pre-PR logic, so profiler rounds keep both bugs this PR fixes. Delegating to `parse_typed_response_text(text, ProfilerResponse)` fixes it and removes the duplicate.

**[P3] Control-character repair duplicates `json.loads(strict=False)`** ([agent_runner.py:257-284](https://github.com/uw-syfi/vibesys/blob/48aad08afb33bfb3e5f7179a6c828945e3154b93/src/vibesys/agent_runner.py#L257-L284))

`json.loads('{"a": "x\ny"}', strict=False)` accepts raw newlines and tabs inside strings and touches nothing outside them. Passing `strict=False` on the existing call removes the second-attempt loop and about thirty lines. Your implementation is correct as written; this is a simplification.

**[P3] PR body**: please use the repo template (`Problem`, `Solution`, `Verification`) from `.github/pull_request_template.md`.

Checked and fine: the narrative knob dict is tried first but fails validation because every response model has required fields, so the scan continues. Scan cost is linear (2 MB with 500k objects in 0.34 s).

Checks run locally at the head since fork CI has not run: `uv run pytest tests/vibesys/test_agent_runner_parse.py tests/vibesys/agents/test_agent_runners.py` (84 passed), `ruff check` and `ruff format --check` clean.

vic-lsh:
Closing for now due to inactivity

PR #634, opened 2026-09-08
fix(tui): dim background with a scrim while a modal is open

### LINKED ISSUE #566: TUI: a modal does not dim the background, so content behind it competes for the eye
### Observed behavior

A modal draws over live content with no scrim, so the transcript behind and beside it stays fully legible. With `/help` open, the operator has two competing texts on screen and nothing indicates which one accepts keys.

### Expected behavior

While a modal is open it is the only readable thing on screen, and the background recedes without moving.

### Reproduction

`clients/tui/dev/mock-ui.sh --tmux 150x40 --speed 0`, then `/help`.

### Acceptance criteria

- [ ] The entire background dims while a modal is open, not only the region behind the box.
- [ ] The modal's own surface stays at full contrast, so the dim reads as depth.
- [ ] Closing restores exactly: nothing about the underlying layout moves.
- [ ] Verified in all eight themes, including the two high-contrast ones where a dim has the least headroom.

### Affected subsystem

`clients/tui/src/ui/overlay.ts`, `ui/chat-overlay.ts`, `ui/app.ts`.

### Impact

A legibility failure rather than a polish issue: the operator cannot tell where a keystroke will land.

Parent: #284. A full-screen view was considered and rejected: it discards the operator's place in the run, which is the reason the modal exists rather than a route.

zatchbell1311-wq:
## Problem

Closes #566

When a modal is open (`/help`, `/design`, or the experiment chat), the background content stays fully visible and legible. The operator has two competing surfaces on screen with no visual indication of which one accepts keystrokes.

## Solution

Add a full-screen scrim `BoxRenderable` at `zIndex: 19` — below the chat modal (20) and the command overlay (25), above everything else. The scrim uses `theme.canvas` at fixed `opacity: 0.7` so the background recedes without moving or reflowing. It is shown whenever `state.overlay !== null || state.chatOpen || state.themePicker !== null` and hidden otherwise.

Closing restores exactly: the scrim is hidden on the same render pass that hides the modal, so nothing about the underlying layout moves.

## Architecture

`OverlayView` owns the scrim alongside the overlay box. `app.ts` mounts `overlay.scrim` on root ahead of `overlay.output` so z-order is respected, and calls `overlay.renderScrim(...)` in the render loop.

The scrim color derives from `theme.canvas`. Note: fixed opacity does not hold WCAG large-text contrast uniformly across all 8 themes — Solarized Dark is the binding case. Per-theme scrim strength computation (as explored in #639) is a follow-up.

`paneFallback` is intentionally excluded from the predicate: it renders a visualization pane through the overlay rather than a true blocking modal, and `paneBorders` only finds the fallback pane in that state — raising the scrim erases the Agents and Transcript frames rather than dimming them. #639 hit the same issue independently. Existing tests assert that background content remains visible in that state.

## Changes

- `ui/overlay.ts` — adds `readonly scrim: BoxRenderable` created once in the constructor; `renderScrim(visible: boolean)` toggles visibility; `applyTheme` updates `scrim.backgroundColor`
- `ui/app.ts` — mounts `overlay.scrim` on root before `overlay.output`; calls `overlay.renderScrim(state.overlay !== null || state.chatOpen || state.themePicker !== null)` in the render loop

## Verification

- All 15 CI checks pass
- `paneFallback` inclusion was attempted but caused 2 existing theming tests to fail; excluded with explanation above and in review comments

### DIFF
--- a/clients/tui/src/ui/app.ts
+++ b/clients/tui/src/ui/app.ts
@@ -48,12 +48,12 @@ export interface OpenTuiApp {
 /** Which of the client's editors currently holds the cursor. */
 type FocusTarget = 'command' | 'chat' | 'modal';
 
-const KEY_HELP = `←→: rounds/agents/transcript · ↑↓: within · [/]: round · F4: zoom · ${COMMAND_NAMES.todos} · ${COMMAND_NAMES.prompt} · Ctrl+L: live`;
-const SCOPED_KEY_HELP = `←→: rounds/agents/transcript · ↑↓: within · [/]: round · F4: zoom · ${COMMAND_NAMES.todos} · ${COMMAND_NAMES.prompt} · Esc: back`;
-const LOG_KEY_HELP = `↑↓ or scroll: select · Enter/click: open hypothesis · F4: zoom · ${COMMAND_NAMES['open-round']} --N`;
-const LOG_CHAT_KEY_HELP = `↑↓: select · Enter/click: hypothesis · Ctrl+W: chat · F4: zoom · ${COMMAND_NAMES['open-round']} --N`;
+const KEY_HELP = `â†â†’: rounds/agents/transcript Â· â†‘â†“: within Â· [/]: round Â· F4: zoom Â· ${COMMAND_NAMES.todos} Â· ${COMMAND_NAMES.prompt} Â· Ctrl+L: live`;
+const SCOPED_KEY_HELP = `â†â†’: rounds/agents/transcript Â· â†‘â†“: within Â· [/]: round Â· F4: zoom Â· ${COMMAND_NAMES.todos} Â· ${COMMAND_NAMES.prompt} Â· Esc: back`;
+const LOG_KEY_HELP = `â†‘â†“ or scroll: select Â· Enter/click: open hypothesis Â· F4: zoom Â· ${COMMAND_NAMES['open-round']} --N`;
+const LOG_CHAT_KEY_HELP = `â†‘â†“: select Â· Enter/click: hypothesis Â· Ctrl+W: chat Â· F4: zoom Â· ${COMMAND_NAMES['open-round']} --N`;
 const HYPOTHESIS_KEY_HELP =
-  '↑↓: select round · Enter/click: trajectory · PgUp/PgDn: scroll · Esc: hypotheses';
+  'â†‘â†“: select round Â· Enter/click: trajectory Â· PgUp/PgDn: scroll Â· Esc: hypotheses';
 /** Bezel, one content row, bezel. See the header frame below. */
 const HEADER_FRAME_HEIGHT = 3;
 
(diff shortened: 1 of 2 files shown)


AyanBinRafaih:
Review finding at `ba9567ac`:

**[P2] Include fallback visualizations and the theme picker in the scrim predicate** ([app.ts:514](https://github.com/uw-syfi/vibesys/blob/ba9567ac5e46dc94fa4ca171b09155e0a57c5bec/clients/tui/src/ui/app.ts#L514)).

At 90 columns, `/design` opens through `paneFallback`: `openPane` clears `state.overlay`, and `overlay.render(state, paneFallback)` shows the modal. This new predicate still evaluates to false, so its background remains undimmed. The theme picker is another visible modal omitted by the same condition. Two renderer reproductions confirmed that the design overlay/theme picker was visible while `overlay-scrim.visible` stayed false. Please derive scrim visibility from all actual modal surfaces, including `paneFallback` and `state.themePicker`, and cover those paths.

Validation: the existing app suite passed 160 tests and TypeScript checking passed; both additional modal/scrim reproductions failed as described.

zatchbell1311-wq:
Thanks for catching that. I added paneFallback to the predicate but it caused two existing tests to fail:

- `theming > leaves every round pane neutral when a fallback modal takes the keys`
- `theming > shows the narrow fallback as the focused visualization, not a Command box`

The issue is that paneFallback renders a visualization pane through the overlay rather than a true blocking modal — content behind it is still intentionally visible. Including it in the scrim predicate hides that context, which is what the tests verify. I've kept state.themePicker in the predicate since that is a true modal.

Happy to revisit if there's a better way to handle the paneFallback case.

Nano-AI:
Cross-reference: #639 targets #566 as well, from a branch that predates this PR. Flagging it so a reviewer landing on both knows they are the same issue. No changes requested here.

This PR targets current main; #639 needed a substantial rebase to get there, so this is likely the one to take.

The piece of #639 that may be worth lifting either way is the per-theme scrim strength in `ui/theme.ts`. The eight themes do not start from the same contrast, so a single fixed blend cannot hold body text at the WCAG large-text floor across all of them. Solving strength per theme lands them between 2.94 and 2.98 from an unscrimmed 5.61 to 21.00, and the binding theme is Solarized Dark rather than either high-contrast palette.

One finding from that rebase that may apply here too: raising the scrim for the narrow-terminal pane fallback erases the Agents and Transcript frames rather than dimming them (`paneBorders` then finds only the fallback pane on screen). #639 excludes the fallback from its predicate for that reason.

zatchbell1311-wq:
Thanks @Nano-AI for the heads up, and for the per-theme contrast analysis — the Solarized Dark binding case is a real gap in the fixed opacity approach. I've noted it as a follow-up in the PR description referencing #639.

The paneFallback exclusion reason matches what you found too — paneBorders only sees the fallback pane so the scrim erases Agents and Transcript rather than dimming them. Good to have independent confirmation on that one.

PR #636, opened 2026-09-09
fix(agent): disclose omitted pending frontier claims past the cap

### LINKED ISSUE #547: TUI: Stop dropping older unreviewed frontier claims past the eighth

Nano-AI:
## Problem

`_pareto_archive_summary` renders the block that warns the agent which measured
frontier points have not passed independent review. It sliced `pending[-8:]`
with no signal that anything was dropped, so on a long run the oldest
still-unreviewed claims silently vanished from the agent's context.

The warning exists precisely to stop an unreviewed commit being treated as a
trusted parent. Dropping the oldest rows without saying so defeats it, and the
issue is labeled "Blocks normal use" for that reason.

This output is not display: it goes to `issue_board.write_pareto_archive`, so it
is agent-facing context.

Closes #547.

> **Overlaps #621 by @AyanBinRafaih**, which fixes the same issue by removing the
> cap entirely (`for record in pending`). This PR is an alternative for
> comparison, not a competing claim on the issue. The difference worth deciding:
> #621 is one line and shows every claim; this keeps a bound on an agent-visible
> file that grows with the run, and discloses what the bound hid. If the team
> prefers #621, close this.

## Solution

Keep a display cap, name it `_PENDING_CLAIM_DISPLAY_LIMIT`, and emit a
disclosure line whenever it bites, giving the count of omitted claims and the
round range they span, with an explicit instruction not to treat them as trusted
parents just because they scrolled off.

### Architecture

No ownership change. The rendering stays entirely inside
`_pareto_archive_summary` in `loops/agent/loop.py`; its caller and the issue
board are untouched.

### Correctness properties

- Fewer pending claims than the limit produces byte-identical output to before.
- The disclosure line appears only when the cap actually elides rows.
- Singular and plural forms are handled for both the claim count and the round
  range, so a single omitted claim does not read as a range.
- The prompt block stays bounded regardless of run length.

### Testing

`PYTHONPATH=$PWD/src .venv/bin/python -m pytest
tests/vibesys/loops/agent/test_orchestrate.py --no-cov`: 148 passed.

PYTHONPATH is prepended because a worktree's symlinked `.venv` otherwise
resolves the package back to the main checkout and tests the wrong tree.

### DIFF
--- a/src/vibesys/loops/agent/loop.py
+++ b/src/vibesys/loops/agent/loop.py
@@ -341,6 +341,13 @@ def _format_metric_row(metrics: dict[str, float], objectives: Sequence[Objective
     )
 
 
+# How many pending frontier claims to render in full. The list can grow
+# unbounded over a long run, so it is still capped, but an omission is always
+# disclosed (see _pareto_archive_summary) instead of silently dropping the
+# oldest, still-unreviewed rows.
+_PENDING_CLAIM_DISPLAY_LIMIT = 8
+
+
 def _pareto_archive_summary(records: list[RoundRecord], space: MetricSpace) -> str:
     """Render trusted frontier parents and any measured points awaiting review."""
     objectives = space.objectives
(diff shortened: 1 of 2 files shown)


Nano-AI:
Closing: superseded by #621, which merged as `8a1312c`.

Both PRs answered #547, and #621 landed the same design this one proposed — keep a bounded list and disclose what was dropped, rather than removing the cap:

```python
omitted = pending[:-_PARETO_ARCHIVE_PENDING_CLAIM_LIMIT]
if omitted:
    f"- {len(omitted)} older untrusted claims omitted from this context "
    f"(rounds {omitted[0].round_number}-{omitted[-1].round_number}); do not "
    "treat any omitted claim as a trusted parent."
```

That is what this branch did, so there is nothing left for it to add. #547 is closed as completed.

One idea from this PR's discussion is **not** covered by what merged and is worth its own issue if anyone wants it: the surviving set is still ranked by recency. The record stores parent round and commit, so it could instead rank by how many rounds descend from a claim, keeping the ones most likely to be built on. That is the issue's own "deterministically prioritized set" and a larger change than either PR.

PR #638, opened 2026-09-09
fix(sandbox): reject a pre-existing writable evaluator runtime directory

Nano-AI:
## Problem

#622 namespaced the Modal evaluator's lock and lease per user and rejected a
runtime directory that is a symlink or owned by another uid. Two residual gaps
from #548 survive that, on a shared host where an attacker knows the invoking
uid and can create `vibesys-<uid>` first:

- `_ensure_runtime_dir` calls `mkdir(mode=0o700, exist_ok=True)`, which
  silently accepts a directory that already existed, and only then
  `chmod(0o700)`. A pre-created group- or world-writable directory may already
  hold files another user planted. The chmod closes access going forward but
  removes nothing.
- Neither the lock nor the lease is opened with `O_NOFOLLOW`.

Both are reachable today:

| Vector | Path on `main` | Effect |
| --- | --- | --- |
| Planted lease file | `_read_deployment_lease` does no ownership check | attacker-chosen `base_url` / `app_identifier` reach `_execute_colocated`, sending the candidate payload to their host |
| Symlink at lock path | `open(_LOCK_PATH, "w")` | truncating write to any file the invoking user can write |
| Symlink at lease path | `Path.read_text()` | reads a file outside the runtime directory as a lease |
| Symlink at lease `.tmp` path | `temporary.write_text(...)` | truncating write outside the runtime directory |

`Path.replace` on the lease path itself is `rename`, which replaces a symlink
rather than following it, so only the staging half of the lease write needed
changing.

## Solution

1. Reject a runtime directory whose mode has `st_mode & 0o022`, rather than
   coercing it. Rejecting beats coercing because the damage predates the
   chmod: the permission bits can be fixed, planted files cannot. The mask is
   `0o022` and not `0o077` because write permission is the planting vector,
   while a `0o755` directory is safe (and common under a shared
   `XDG_RUNTIME_DIR`) and its read bits are already closed by the existing
   `chmod(0o700)`, which is kept.
2. Open the lock, the lease read, and the lease staging write with
   `os.open(..., O_NOFOLLOW)` wrapped in `os.fdopen`. A planted symlink now
   fails with `ELOOP` instead of redirecting the read or write.

The directory is deliberately not unlinked and recreated: that would delete a
live lock held by another process of the same user.

### Architecture

No boundary change. Both hunks stay inside `src/vibesys/sandbox/modal_evaluator.py`;
`_ensure_runtime_dir` remains the single validation point that
`_exclusive_evaluation` and `_write_deployment_lease` call before touching the
directory.

### Correctness properties

- A runtime directory group- or world-writable is rejected with a `RuntimeError`
  naming the reason, whether or not this process created it.
- A `0o755` runtime directory is still accepted and tightened to `0o700`; the
  fix does not over-reject.
- A symlink at the lock, lease, or lease-staging path raises `OSError` with
  `errno.ELOOP`, and the file it points at is left byte-identical.
- The pre-existing symlink-shadowed-directory and foreign-uid checks from #622
  are unchanged.

### Testing

Five cases added to `tests/vibesys/sandbox/test_modal_evaluator.py`. Four of the
five fail against `main` and pass with this change; the `0o755` case passes on
both by design, as the guard against over-rejecting.

```
# against main (fix reverted, new tests present)
4 failed, 1 passed, 59 deselected

# with the fix
uv run pytest tests/vibesys/sandbox/test_modal_evaluator.py   ->  64 passed
uv run pytest tests/vibesys/sandbox/                          -> 169 passed
uv run ruff check / ruff format --check (changed files)       -> clean
./scripts/check_types.sh                                      -> All checks passed
```

Every external boundary stays mocked with `monkeypatch`; no network and no real
`modal` CLI.

Refs #548. #622 covered the rest of that issue.

### DIFF
--- a/src/vibesys/sandbox/modal_evaluator.py
+++ b/src/vibesys/sandbox/modal_evaluator.py
@@ -94,9 +94,10 @@ def _ensure_runtime_dir(path: Path) -> None:
 
     Raises ``RuntimeError`` naming the directory and its owning uid when the
     location cannot be used as a private per-user directory, for example a
-    plain file already occupies it or another user owns it, so a hostile or
-    stale runtime-directory entry fails with a clear message instead of a
-    deep ``PermissionError`` surfacing from ``flock`` or ``open``.
+    plain file already occupies it, another user owns it, or it already exists
+    writable by other users, so a hostile or stale runtime-directory entry
+    fails with a clear message instead of a deep ``PermissionError`` surfacing
+    from ``flock`` or ``open``.
     """
     runtime_dir = path.parent
     try:
@@ -110,6 +111,18 @@ def _ensure_runtime_dir(path: Path) -> None:
         raise RuntimeError(  # noqa: TRY003
             f"refusing to use {runtime_dir} as the evaluator runtime directory: "
             f"expected a directory owned by uid {os.getuid()}"
+        )
+    # ``mkdir(exist_ok=True)`` accepts a directory that already existed, and the
+    # ``chmod`` below only closes access from here on: it cannot remove files
+    # another user planted while the directory was writable by them. So reject a
+    # writable directory instead of coercing it. Test the write bits alone, not
+    # all of ``0o077``, because write permission is the planting vector while a
+    # readable ``0o755`` directory is both safe and common.
+    if metadata.st_mode & 0o022:
+        raise RuntimeError(  # noqa: TRY003
+            f"refusing to use {runtime_dir} as the evaluator runtime directory: "
+            "it is group- or world-writable, so another user may already have "
+            "planted files in it"
         )
     runtime_dir.chmod(0o700)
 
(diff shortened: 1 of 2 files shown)


Nano-AI:
Repointed. The original content of this PR is obsolete: #622 merged as 6aa5a3e and covers the per-user namespacing it did. The branch has been reset onto current `main` and now carries only the residual hardening from #548 that #622 left behind:

- a pre-existing group- or world-writable `vibesys-<uid>` is rejected rather than coerced by the trailing `chmod`, which cannot remove files an attacker already planted;
- the lock, the lease read, and the lease staging write get `O_NOFOLLOW`, so a planted symlink fails instead of redirecting the read or write outside the runtime directory.

Base is unchanged (`main`, no stack). Open for review, and CI is green.

PR #639, opened 2026-09-09
fix(tui): dim the background behind a modal

### LINKED ISSUE #566: TUI: a modal does not dim the background, so content behind it competes for the eye

Nano-AI:
> **Overlaps #634 by @zatchbell1311-wq**, which fixes #566 against current main.
> This PR is an alternative for comparison, not a competing claim. Its
> distinguishing piece is the per-theme scrim strength derivation in
> `ui/theme.ts`; if #634 is the one taken, that derivation is the part worth
> lifting.

## Problem

A modal draws over live content with no scrim, so the transcript behind and
beside it stays fully legible and nothing indicates which surface is taking
keys. With `/help` open the operator has two competing texts on screen.

Repro from #566: `clients/tui/dev/mock-ui.sh --tmux 150x40 --speed 0`, then
`/help`. The issue calls this "a legibility failure rather than a polish issue:
the operator cannot tell where a keystroke will land."

Refs #566.

## Solution

`ScrimView` is one absolutely positioned box covering the terminal at zIndex 15:
above every pane and the lists that rise out of them, below the chat modal (20),
the command overlay (25), and the theme picker (30). One scrim serves all three,
so two stacked modals never dim each other. The paint is translucent and the
renderer blends it per cell, so a background cell keeps its character and
changes only color. Nothing joins the flex layout, so nothing reflows.

`scrim(theme)` in `ui/theme.ts` is the single place the dim is computed. It
pulls toward the theme's own `canvas`, so neither light nor high-contrast
palettes get a tone they exclude. Strength is solved per theme rather than
fixed, because the themes do not start from the same contrast: measured
strengths run 0.374 to 0.655, and the binding theme is Solarized Dark, not the
high-contrast pair.

### Architecture

`ui/scrim.ts` owns the box and nothing else. `ui/theme.ts` owns the derivation.
`ui/app.ts` owns only the predicate for when it is raised.

### Correctness properties

- The scrim repaints cells and lays out nothing, so opening a modal moves no
  character outside the modal's rectangle. **This property no longer holds on
  current main**, which is the substance of the failures below.
- One scrim serves all three modals, so stacked modals never dim each other.
- Every theme keeps modal body text at its own contrast floor.

### Testing

`pnpm test:clients`: 704 pass, 0 fail.

Rebasing onto current main took real work, recorded here because it changed the
change:

- Conflicts resolved toward main's current APIs: `RoundStripView` to
  `RoundRailView`, the curated header (#571) now owning the Escape hint, and
  `overlay.render` taking `paneFallback` as its own argument.
- The per-theme `dims every cell` tests assert that opening a modal moves no
  character outside the modal. That held, but the test assumed the header was
  row 0; since #571 the header sits in its own housing, so its content row is 1.
  The test now reads the header's rows off `header-frame` rather than assuming.
- **The scrim no longer rises for the narrow-terminal pane fallback.** An
  earlier pass wired the fallback into the predicate, on the reasoning that it
  is a modal in everything but name. Measured against main, that erases the
  Agents and Transcript frames rather than dimming them: `paneBorders` finds
  only `▸ Performance` on screen with the fallback open and the scrim raised.
  That is a real defect in the interaction, not a test-method problem, so the
  predicate covers the three real modals only and the fallback is left for
  follow-up. The comment in `app.ts` says so.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -170,7 +170,11 @@ modal they used before panes existed. That modal is the same surface as the
 pane, so it keeps the pane's title and its focus marker rather than reading as a
 generic dialog. The layout re-flows on resize in either direction.
 
-`/help`, `/theme`, and errors stay modal.
+`/help`, `/theme`, and errors stay modal. While any of them is open, a scrim
+dims the entire screen behind it, so the modal is the only surface left at full
+contrast and the operator can tell where a keystroke will land. The scrim is a
+translucent paint applied over the finished frame rather than a layout node: the
+background keeps every character where it was, and closing restores it exactly.
 
 ### Experiment chat
 
@@ -270,6 +274,11 @@ the Markdown palette are derived from that core, and each derived foreground is
 pushed toward the nearest extreme until it clears the theme's `minContrast`
 against the surface it actually sits on. The `dark` theme additionally pins its
 derived values to the original literals so the baseline is byte-identical.
+The modal scrim is derived the same way: it pulls the background toward the
+theme's own `canvas`, and its strength is solved per theme so body text lands on
+WCAG's large-text floor whatever it started from. One fixed blend cannot do
+that, because a blend deep enough to recede Solarized Dark's 5.6:1 body text
+leaves High Contrast Dark's 21:1 fully readable.
 Status meaning never depends on color: agent phases carry a marker glyph and
 the spelled-out status, todos carry a per-status marker, and only the running
 round shows elapsed time.
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -2,6 +2,7 @@ import {afterEach, describe, expect, it} from 'bun:test';
 import {
   CliRenderEvents,
   InputRenderable,
+  type Renderable,
   rgbToHex,
   ScrollBoxRenderable,
   TextareaRenderable,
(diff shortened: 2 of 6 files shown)


Nano-AI:
Closing and reopening this against `main`.

Its base was `adi/card-role-stripe` (#640), which is now closed as superseded by #647. GitHub will not let a PR that belongs to a stack change its base, and `gh stack link` will not drop a member from an existing stack, so the only way to retarget was a fresh PR from the same branch.

Nothing about the change itself moved. The dimming work never depended on the stripe: the two only shared `app.test.ts`. The branch is rebased directly onto `main` and is unchanged otherwise.

Issue #516, opened 2026-09-01
TUI: render should agent-trace tool calls and results as readable summaries instead of raw JSON, shell noise, and provider stderr


AyanBinRafaih:
### Workstream

CLI/DX

### Problem

The round-view transcript is the primary surface for understanding what an agent did, and today it is dominated by machine-format noise. Observed live on a codex run: tool call previews print raw argument JSON verbatim, for example `→ file_change(changes=[{"path":"/tmp/vibesys-runs/queue-hunt/src/lib.rs","kind":"delete"}])`, and shell calls print the full wrapper with broken nested quoting, for example `→ execute(command="/bin/bash -lc "sed -n '1,240p' .vibesys/state/runs/20260831-...")` where the inner quotes terminate the display string and a stray `.")` wraps onto its own line. Tool results with a JSON payload are only re-indented (`JSON.stringify(value, null, 2)` at `clients/tui/src/ui/previews.ts:55` and `:97`), never summarized or field-selected, so an expanded card is a wall of raw JSON. Provider lifecycle chatter is rendered as full bordered transcript cards: `[codex stderr] Reading prompt from stdin...[codex thread 01a0... started][codex turn started]`, `[codex stderr] 2026-08-31T21:54:32Z ERROR codex_core::tools::router: error=apply_patch verification failed...`, plus `driver: agentshim, provider: codex, model: gpt-5.6-sol, ...` banners and bare `--- input ---` marker cards, one box each. In this session 11 of 24 tool results exceeded 2000 characters of raw text behind the 6-line/600-char collapse (`MAX_TOOL_OUTPUT_LINES`/`MAX_TOOL_OUTPUT_CHARACTERS`, previews.ts:3-6). There is no per-tool rendering: `previews.ts` has exactly one generic formatter for every tool, the result band has no word-wrap mode set (unlike `right-pane.ts:130`), and paired call/result events that never receive a result (codex `file_change`) render as duplicated bare cards. The backend already ships the structure needed to do better: `ToolResultData.payload` is a typed discriminated union (`command` with stdout/stderr/exit_code/duration, `json` with the parsed value) in `src/vibesys/server/events.py:229-262`, and calls carry `tool` plus typed `args`.

### Desired outcome

[...]

Issue #565, opened 2026-09-03
TUI: the transcript spends its rows on card chrome rather than separation


Nano-AI:
### Workstream

CLI/DX

### Problem

The transcript spends its vertical budget on per-card chrome rather than on separation between cards. Every entry is a rounded bordered box with its own padding, so a short status line costs three rows to say one thing, and a screen of them reads as a stack of containers rather than a sequence of events.

The card padding is also inconsistent with what surrounds it: cards carry internal padding while the panes they sit in do not, so the eye gets a lot of information with little organisation.

This is distinct from #516, which is about the *content* being noisy (raw JSON, shell wrappers, provider chatter). This is about the *density* of whatever content survives that work.

### Desired outcome

A reader can scan a transcript and see where one entry ends and the next begins, without each entry costing a frame's worth of rows.

### Constraints or design considerations

Two directions worth pricing before choosing:

- **Dividers instead of cards.** A rule between entries does the same separating job as a border on four sides, at a fraction of the vertical cost. It gives up the ability to nest content inside a visible container.
- **A collapsible accordion.** Each entry summarises to one row and expands on demand. Best density, but adds interaction state and a keybinding, and needs a rule for what a collapsed entry shows.

Either interacts with #516: summarising tool output may change how much room an entry needs, so settle that first or design them together. The role stripe explored on `adi/card-role-stripe` is also unresolved and touches the same surface.

### Acceptance criteria

- [ ] A transcript entry costs materially fewer rows than a bordered card.
- [ ] Entry boundaries stay legible without a full border.
- [ ] Padding is consistent between an entry and the pane containing it.
- [ ] Role and selection stay distinguishable, in all eight themes.

### Verification approach

Compare rows-per-entry before and after on a recorded run through the replay harness, at several terminal heights, in a light and a dark theme.

### Parent, dependencies, or related work

Parent: #284. Related: #516.

PR #640, opened 2026-09-09
feat(tui): give a transcript card a role stripe instead of a role fill

Nano-AI:
## Problem

A transcript card painted its conversation role as a full-width background fill
mixed from the canvas and the role accent, while selection was carried by the
card's border color. The fill is far heavier than a border: across a screen of
cards it reads as "this pane is highlighted" rather than "this card is role X",
and it outweighs the cursor drawn on top of it.

In the two high-contrast themes the fill also had almost no headroom to work
with: assistant in high-contrast-dark was `#001a1a` on a `#000000` canvas, which
is a role signal nobody can see.

No issue: this came out of the transcript-legibility work and is opened for
review on its own. It is adjacent to #516 (@AyanBinRafaih's #620), which
reshapes what a card *contains*; this changes how a card carries its role. They
touch the same surface, so whichever lands second will need a look at the other.

## Solution

The role moves to the card body's left edge as a half block, and the card paints
no fill at all.

```text
  before                        after
  ╭──────────────────────────╮  ╭──────────────────────────╮
  │ implementer · round-1    │  │▌implementer · round-1    │
  │ Short answer: yes. The   │  │▌Short answer: yes. The   │
  ╰──────────────────────────╯  ╰──────────────────────────╯
```

The stripe is the body's left border rather than a renderable or a text prefix.
A border reserves its column, and it is the column `paddingLeft` already held,
so no text moves. It costs the same cell whether or not a glyph is painted in
it, so a status line, which has no conversation role, reserves the column
unpainted and lines up with the cards around it instead of hanging a column to
their left. It also runs the body's full height, which a card holding a markdown
block does not know in advance.

A half block, not `│`: the card's rounded frame already owns the box-drawing
vocabulary and a second vertical inside it reads as a double frame. Not a full
block either, which is the same fill one column wide and competes with the
frame, and the frame has to stay the loudest mark on the card because it is what
carries the cursor.

### Architecture

Contained in `ui/conversation.ts`. `BorderCharacters` is a closed shape, so a
one-sided border is expressed by giving every unreachable member a space;
`stripeChars()` is that helper and the only new construct.

### Correctness properties

- A status line reserves the stripe column unpainted, so every card and status
  line shares one left edge.
- No text moves relative to the previous fill-based card: the stripe occupies
  the column `paddingLeft` used to hold.
- Selection remains the border color, and with the fill gone it is now the
  strongest mark on the card in every theme.
- The role is legible in the high-contrast themes, where the fill was not.

### Testing

`bun test clients/tui/src/ui/app.test.ts clients/tui/src/ui/conversation.test.ts`:
175 pass, 0 fail.

Rebased onto current main. Two commits that were already merged via #572 were
dropped in the rebase, so this branch is now the single stripe commit.

## Screenshots

Real terminal captures at 150x40, round 1 transcript, F4-zoomed. Colours audited
against each worktree's `theme.ts`.

**Dark**: each card carries its role as a `▌` on the body's left edge, card
interiors on plain canvas. Sampled: analysis stripe `#b193f8` (exact), success
stripe `#6dd895` (d=1), card body `canvas #0f172a` (exact, no role fill).

dark

**High-contrast-dark**, the theme where the old fill had no headroom: analysis
stripe `#eab4ff`, success stripe `#58ffac`, card bodies `canvas #000000`. The
role reads at full saturation on pure black, where the fill it replaces was
`#001a1a` and invisible.

high contrast dark

### DIFF
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -2478,7 +2478,9 @@ describe('theming', () => {
     expect(spanColors(testRenderer, 'VibeSys')?.fg).toBe(light.accent);
     const body = spanColors(testRenderer, 'themed body text');
     expect(body?.fg).toBe(light.conversation.assistant.content);
-    expect(body?.bg).toBe(light.conversation.assistant.background);
+    // A card carries its role on a left edge stripe rather than a fill, so its
+    // body sits on the theme's canvas.
+    expect(body?.bg).toBe(light.canvas);
     expect(spanColors(testRenderer, 'implementer')?.fg).toBe(light.conversation.assistant.label);
   });
 
@@ -2499,8 +2501,8 @@ describe('theming', () => {
     expect(spanColors(testRenderer, 'VibeSys')?.fg).toBe('#22d3ee');
     const body = spanColors(testRenderer, 'themed body text');
     expect(body?.fg).toBe('#e2e8f0');
-    // Assistant cards derive their fill from the canvas and the role accent.
-    expect(body?.bg).toBe('#0e283d');
+    // No card fill: the role is a stripe, so a card body is the canvas.
+    expect(body?.bg).toBe('#0f172a');
     expect(spanColors(testRenderer, 'implementer')?.fg).toBe('#5cb6cc');
   });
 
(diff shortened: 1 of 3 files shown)


Nano-AI:
Closing as superseded by #647.

Both PRs answer the same question: how a transcript card shows its role once the
role *fill* is gone. This one keeps the four-sided rounded border and moves the
signal to a left-edge stripe drawn against it. #647 removes the border and the
fill outright and moves the signal onto the top divider rule, which also takes a
one-line entry from 5 rows to 3.

They cannot both land. The stripe here is positioned relative to a border #647
deletes, and both branches add their own `clients/tui/src/ui/conversation.test.ts`,
so there is an add/add conflict on top of the layout one.

One thing this approach had that #647 does not: a left-edge stripe runs the full
height of an entry, so the role stays visible when a long entry is scrolled past
its own heading. #647 carries role on the divider and in the heading label, both
of which sit at the top. If that turns out to matter in use, the cheaper fix is a
stripe on top of #647's divider rather than reviving the bordered card, and it
should be filed against #565 rather than reopening this.

Stack #641 is restructured accordingly: #635 -> #639, with #639 rebased directly
onto #635. Its dimming work never depended on the stripe.

PR #643, opened 2026-09-09
fix(tui): raise elevatedSurface above canvas in the three inverted dark themes

### LINKED ISSUE #574: TUI: the surface ladder is inverted, so panes read as recesses rather than panels
### Workstream

CLI/DX

### Problem

The surface ladder is inverted: the "elevated" surface is darker than the canvas it sits on.

```
canvas          #0f172a   root background
elevatedSurface #020617   every pane
```

In editors an elevated surface is lighter than its ground, which is why a raised element reads as raised. Here every pane reads as a recess cut into the page rather than a panel on it, and anything seated on `elevatedSurface` inherits that. An operator described the result as "unnatural": the header looked like a darker band laid over a lighter application.

Two recent fixes made components *consistent* with this ladder (`headerFrame` and `chat-pane` now both name `elevatedSurface`). Neither addresses the ladder itself.

### Desired outcome

Surfaces read as a hierarchy: the thing in front is nearer, and depth matches the visual weight an operator expects from an editor.

This is a palette change across eight themes, not a token swap. Simply exchanging the two values would invert every contrast pair derived from them, so every foreground that currently clears its floor against one would need rechecking against the other.

A method worth following, suggested by the operator: colour one component at a time, capture the real terminal, measure the contrast, and assign roles by what the component is (focus, primary, essential, metadata) rather than by which token was nearest. `~/dev/vs-shot` captures actual terminal pixels for this; a re-rendered ANSI screenshot is not trustworthy for colour work.

Two related asks belong with this rather than alone:

- Move the experiment chat to the right side, matching editors with an assistant panel.
- Give the chat a distinct shade so it reads as a companion surface rather than a second primary pane.

Both the experiment view and the run view need doing, and they have different surface counts.

### Acceptance criteria

- [ ] A raised surface is lighter than the ground behind it, consistently.
- [ ] Every foreground still clears its theme's contrast floor, verified in all eight.
- [ ] The high-contrast pair remains legible; they have the least headroom.
- [ ] Verified against real terminal captures, not re-rendered ANSI.

### Verification approach

Per theme, capture the experiment view and a run view with `vs-shot`, sample the actual pixels, and tabulate each surface against the ground and each foreground against its surface.

Parent: #284. Related: #565, #566.

Nano-AI:
## Problem

Closes #574. In `dark`, `solarized-dark`, and `catppuccin-mocha`, `elevatedSurface`
shipped darker than `canvas`. Every pane (chat-pane, header, error-banner,
experiment-log, overlays, theme-picker, composer slash menu) is painted on it, so
panes read as recesses cut into the page rather than panels raised above it. The
four light themes were already correct; the two high-contrast themes pin both
tokens to the same extreme and have no headroom either way.

Measured from real terminal captures, canvas vs pane:

| theme | before | after |
| --- | --- | --- |
| dark | 1.129:1 **recessed** | 1.120:1 raised |
| solarized-dark | 1.14:1 **recessed** | 1.097:1 raised |
| catppuccin-mocha | 1.14:1 **recessed** | 1.036:1 raised |

**Before** — the pane sits below the canvas, so it reads as a recess.

before

**After** — the pane sits above the canvas and reads as a panel.

after

`solarized-dark` and `catppuccin-mocha` change identically; the ratios in the
table above are the evidence for those, rather than four more frames.

## Solution

Only `elevatedSurface` moves. `canvas` feeds `ensureContrast` and `mix` for the
text, accent, conversation, and markdown tokens, so moving it ripples across the
palette and through pinned literals in `theme.test.ts`.

How far each theme can rise is capped by `textSubtle`, painted raw with no
re-contrast in every consumer but the header, and clearing its 3:1 floor against
canvas by a thin margin (3.75:1 dark, 3.37:1 solarized-dark). Values were chosen
inside that bound, from each theme's own palette family, hue held and lightness
only: `dark` `#020617` -> `#182234`, `solarized-dark` `#001f27` -> `#0a323d`,
`catppuccin-mocha` `#11111b` -> `#212131`. The per-theme ceilings and why the
obvious next palette stop overshoots are documented at each call site.

### Correctness properties

- `elevatedSurface` luminance is never below `canvas`, in all eight themes.
- Every foreground drawn raw on a pane (`textPrimary`, `textMuted`, `textSubtle`,
  `border`, `borderFocus`) clears its floor against `elevatedSurface`, in all
  eight. Floors are the ones `theme.test.ts` actually asserts: `border` and
  `borderFocus` never pass through `ensureContrast`, so theirs is 1.7:1.
- `selectedSurface` stays distinct from `elevatedSurface`, in all eight.
  `theme-picker.ts:78` picks between exactly those two.

Three table-driven tests over `THEME_NAMES` pin these. All three regressed
silently during development because the suite only ever asserted against `canvas`.

### Testing

`pnpm check:ts`, `pnpm check:clients`, `pnpm build:clients` all pass.
`pnpm test:clients`: 687 tui, 84 core-state, 25 backend-client, 2 architecture,
0 failures. Screenshots above are real terminal pixels via `vs-snap`, converted
to sRGB and verified against `theme.ts` at `d=0`.

### Known limits

- `catppuccin-mocha`'s step measures 1.036:1, which reads flat; depth there is
  still carried by the border. Reaching `surface0` (`#313244`) requires moving
  `border` off `surface1`, which collapses it into `borderStrong` (`surface2`).
  Verified as feasible (textPrimary 8.69, textMuted 5.65, textSubtle 3.40,
  border 1.88) but left out of scope. For reference, `light` ships at 1.046:1.
- `catppuccin-mocha`'s `surface` (mantle) is still darker than its canvas, a
  separate pre-existing inconsistency.
- `dark` and `solarized-dark` now have `elevatedSurface` below their own
  `surface`, a trade the `textSubtle` floor forces.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -262,11 +262,17 @@ the input still runs on its own Enter. `/theme <name>` re-themes every view in
 place without opening the list.
 
 `ui/theme.ts` is the only module holding color literals. A theme declares
-semantic roles — `canvas`, `surface`, `elevatedSurface`, `selectedSurface`;
-`textPrimary`, `textMuted`, `textSubtle`, `textStrong`; `border`,
-`borderStrong`, `borderFocus`; `accent`, `info`; `success`, `warning`, `error`;
-per-role conversation card colors; and Markdown/code colors. Views ask for a
-role and never for a color.
+semantic roles — `canvas`, `surface`, `selectedSurface`; `textPrimary`,
+`textMuted`, `textSubtle`, `textStrong`; `border`, `borderStrong`,
+`borderFocus`; `accent`, `info`; `success`, `warning`, `error`; per-role
+conversation card colors; and Markdown/code colors. Views ask for a role and
+never for a color.
+
+`canvas` is the only background the UI paints: every pane, modal, the header,
+the error banner and the composer fill with it, and a box is told from the page
+by its border rather than by a shade of its own. The other two fills name a
+state and not a depth — `selectedSurface` is what the cursor is on, and
+`surface` backs a code block, which has no border to delimit it.
 
 Adding a theme means adding one `ThemeSpec`: a semantic core plus one accent
 per conversation role. Card fills, labels, body text, the tool-call band, and
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -170,7 +170,11 @@ modal they used before panes existed. That modal is the same surface as the
 pane, so it keeps the pane's title and its focus marker rather than reading as a
 generic dialog. The layout re-flows on resize in either direction.
 
-`/help`, `/theme`, and errors stay modal.
+`/help`, `/theme`, and errors stay modal. While any of them is open, a scrim
+dims the entire screen behind it, so the modal is the only surface left at full
+contrast and the operator can tell where a keystroke will land. The scrim is a
+translucent paint applied over the finished frame rather than a layout node: the
+background keeps every character where it was, and closing restores it exactly.
 
 ### Experiment chat
 
@@ -270,6 +274,11 @@ the Markdown palette are derived from that core, and each derived foreground is
 pushed toward the nearest extreme until it clears the theme's `minContrast`
 against the surface it actually sits on. The `dark` theme additionally pins its
 derived values to the original literals so the baseline is byte-identical.
+The modal scrim is derived the same way: it pulls the background toward the
+theme's own `canvas`, and its strength is solved per theme so body text lands on
+WCAG's large-text floor whatever it started from. One fixed blend cannot do
+that, because a blend deep enough to recede Solarized Dark's 5.6:1 body text
+leaves High Contrast Dark's 21:1 fully readable.
 Status meaning never depends on color: agent phases carry a marker glyph and
 the spelled-out status, todos carry a per-status marker, and only the running
 round shows elapsed time.
--- a/clients/tui/src/ui/agent-map.ts
+++ b/clients/tui/src/ui/agent-map.ts
@@ -79,6 +79,13 @@ const AGENTS_TITLE = 'Agents';
 
 export class AgentMapView {
   readonly output: BoxRenderable;
+  /**
+   * Everything this view draws, in a box that outlives what it draws. `#clear`
+   * destroys the graph on every repaint, and while this pane is zoomed the
+   * command column is a child of `output` too: without the separation the first
+   * repaint after a zoom destroyed the command input along with the graph.
+   */
+  readonly #content: BoxRenderable;
   #theme: Theme;
   #renderedState: SessionState | null = null;
   #renderedWidth = 0;
@@ -106,6 +113,15 @@ export class AgentMapView {
       title: paneTitle(AGENTS_TITLE, false),
       onMouseUp: () => this.controller.focusRound('agents'),
     });
+    this.#content = new BoxRenderable(renderer, {
+      id: 'agent-map-content',
+      width: '100%',
+      flexGrow: 1,
+      flexShrink: 1,
+      flexDirection: 'column',
+      onMouseUp: () => this.controller.focusRound('agents'),
+    });
+    this.output.add(this.#content);
   }
 
   applyTheme(theme: Theme): void {
@@ -161,7 +177,7 @@ export class AgentMapView {
         roundNumber === null
           ? null
           : (stripRounds(state).find(item => item.number === roundNumber) ?? null);
-      this.output.add(
+      this.#content.add(
         new TextRenderable(this.renderer, {
           content:
             round?.status === 'planned'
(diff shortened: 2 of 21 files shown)


AyanBinRafaih:
[P2] Keep Solarized Dark warning text readable on the raised pane surface

Head: `640c26008dddc1f7538be6e492ee615244fdb540`

[`theme.ts:419`](https://github.com/uw-syfi/vibesys/blob/640c26008dddc1f7538be6e492ee615244fdb540/clients/tui/src/ui/theme.ts#L419) raises Solarized Dark's pane background to `#0a323d`, but `warning` remains `#b58900`. The same warning was 5.34:1 against the base pane `#001f27`; it is 4.26:1 on the new pane, below the theme's declared 4.5 `minContrast` ([`theme.ts:433`](https://github.com/uw-syfi/vibesys/blob/640c26008dddc1f7538be6e492ee615244fdb540/clients/tui/src/ui/theme.ts#L433)). A nonfatal error banner draws both its message and hint directly in `theme.warning` on `elevatedSurface` ([`error-banner.ts:87-114`](https://github.com/uw-syfi/vibesys/blob/640c26008dddc1f7538be6e492ee615244fdb540/clients/tui/src/ui/error-banner.ts#L87-L114)), so Solarized Dark makes active diagnostics less readable after this change.

Reproduce from this head:

```sh
bun -e "import {contrastRatio,resolveTheme} from './clients/tui/src/ui/theme.ts'; const t=resolveTheme('solarized-dark'); console.log({warning:t.warning,elevatedSurface:t.elevatedSurface,ratio:contrastRatio(t.warning,t.elevatedSurface).toFixed(2)})"
```

Expected at least `4.50`; actual output is `{ warning: "#b58900", elevatedSurface: "#0a323d", ratio: "4.26" }`. `bun test clients/tui/src/ui/theme.test.ts clients/tui/src/ui/header.test.ts` still passes all 112 tests because the new test only checks `textPrimary`, `textMuted`, `textSubtle`, `border`, and `borderFocus` ([`theme.test.ts:112-128`](https://github.com/uw-syfi/vibesys/blob/640c26008dddc1f7538be6e492ee615244fdb540/clients/tui/src/ui/theme.test.ts#L112-L128)), not the semantic foreground actually used by the banner. Adjust the pane surface or warning foreground so this rendered message/hint pair still reaches the theme floor, and add a regression assertion for semantic foregrounds actually drawn on raised surfaces. This is a remaining violation of #574's acceptance criterion that every foreground retain its advertised contrast.

Nano-AI:
Addressed in `cacace3`.

The reported ratio reproduces exactly at `640c260`: `warning` `#b58900` on `elevatedSurface` `#0a323d` gives 4.26, under the 4.5 floor.

Tracing every semantic foreground actually drawn raw on `elevatedSurface`, rather than only the one reported, turned up four further violations, all confined to Solarized Dark:

| token | before | after |
| --- | --- | --- |
| error | 4.15 | 4.58 |
| info | 4.19 | 4.67 |
| success | 4.27 | 4.78 |
| accent | 4.33 | 4.56 |
| warning | 4.26 | 4.75 |

The other seven themes had none. `conversation.failure.content`, `conversation.analysis.label` and `textStrong` are also drawn raw on `elevatedSurface`, but already clear the floor in every theme.

The correction sits in the token builder rather than at the four consumer call sites (`error-banner.ts`, `overlay.ts`, `experiment-log.ts`, `theme-picker.ts`). `ensureStatusContrast` chains the existing canvas `ensureContrast` with a second pass against `elevatedSurface`, and `buildTheme` applies it to `accent`, `info`, `success`, `warning` and `error`. `header.ts` already re-derives its own tones against `headerBackground` per call site, so the pattern is established; applying it once in the builder covers every present and future consumer without touching any of them. Two files, 27 lines.

Since `elevatedSurface` is never below `canvas` in luminance after this PR, the second pass can only hold or widen the first gap, never close it, so the canvas guarantee is preserved rather than traded away. Confirmed across the full matrix: 40 of 40 theme/token pairs now clear 4.50 against `elevatedSurface`, and none regressed against `canvas`.

The regression assertion lands in `theme.test.ts`'s surface-elevation block and covers semantic foregrounds on raised surfaces, which the previous assertion did not, pinning only `textPrimary`, `textMuted`, `textSubtle`, `border` and `borderFocus`. It fails at `640c260` (4.265) and passes after.

`pnpm test:clients`: 688 pass, 0 fail.

One note on SHAs: the branch was force-pushed to strip session URLs from the commit messages. Trees are byte-identical; `640c260` is now `ef838ba`.

Nano-AI:
Closing and reopening this against `adi/566-modal-backdrop`.

The change was reworked (the surface ladder is now collapsed to one background rather than re-ordered), and it turned out to conflict with #646 on `chat-pane.ts`, `error-banner.ts`, `experiment-log.ts` and `app.test.ts` — both PRs rewrite where fills live. Rather than leave that for whichever merged second, this branch is rebased onto #651 so the two halves of the rule sit in one line of work.

A PR that belongs to a stack cannot change its base, and `gh stack link` will not merge two existing stacks or drop a member, so a fresh PR from the same branch was the only way to retarget. The branch and its content are unchanged by the reopen.

The P2 contrast finding above and its reply stay readable here; the fix (`ensureStatusContrast`) is carried in the reworked commit.

PR #644, opened 2026-09-09
feat(tui): highlight fenced code blocks when a grammar ships

### LINKED ISSUE #575: TUI: syntax-highlight fenced code blocks in the transcript
# TUI: syntax-highlight fenced code blocks in the transcript

### Workstream

CLI/DX

### Problem

After fenced blocks receive a readable surface, their contents still render as
plain text. This makes long code samples harder to scan than the same content in
an editor or other agent CLIs.

### Desired outcome

A fence with a supported language uses theme-aware syntax colours. Unlabelled
and unknown languages remain readable as plain text.

### Acceptance criteria

- [ ] C++, Rust, Go, Python, and TypeScript fences are highlighted.
- [ ] Unknown and unlabelled fences fall back to plain text.
- [ ] Token colours remain legible in all built-in themes.
- [ ] Streaming works without network access.
- [ ] Any new dependency is justified and included in the published package.

### Scope

Transcript fenced blocks only. Tool output and diffs remain under #516.

### Verification approach

Extend the markdown replay fixture with supported, unknown, and unlabelled
fences. Check one light and one dark theme while streaming.

Parent: #284. Depends on PR #545. Hold this issue until the markdown renderer
lands and the dependency cost has been measured.

Nano-AI:
## Problem

Closes #575, partially. Fenced code blocks in the transcript rendered as flat
single-colour text, so a long code sample was harder to scan than the same
content anywhere else.

The cause was not a missing feature. `@opentui/core` already tokenizes the fence,
already resolves the info string through `infoStringToFiletype`, and already
ships tree-sitter grammars. VibeSys then discarded the result: `drawOnCodeSurface`
set a flat `block.fg` over the whole block and forced `drawUnstyledText = true`,
which suppresses the styled chunks. That was correct reasoning for a language with
no grammar (the block would otherwise sit blank while a wait resolves to plain
text anyway) and wrong for one that has a grammar.

The light theme and the no-grammar fallback path behave the same way; both are
covered in Verification rather than as further frames, since the delta this PR
is asking to be reviewed is highlighting off versus on.

### Before / after

The `typescript` fence in the transcript (the ring-buffer `push<T>`), 150x40,
dark, cropped `--rows 3:19 --cols 0:80`. Base is #652. Re-shot 2026-09-10:
before at `880a251`, after at `a696cb8`.

Both frames use the **same** fixture, this branch's `markdown.jsonl`, passed to
both runs explicitly. The base's own fixture has only `rust` and `bash` fences,
so a before frame shot from it would carry no comparable fence at all. The two
frames are character-for-character identical text; only the rendering differs.

**Before** - the whole fence is one flat colour. Measured: exactly two
foreground colours in the crop, `textPrimary` 73.53% (all of the code) and
`borderFocus` 26.47% (the pane frame and its title). Note the flat colour is
`textPrimary`, not `markdown.code`: `drawOnCodeSurface` does assign
`block.fg`, but the `drawUnstyledText` path draws with the block's inherited
prose foreground.

before

**After** - ten foreground colours, one per family the branch maps: keyword
`#38bdf9`, string `#22c55d`, comment `#94a2b8`, number `#facc14`, function
`#87909c`, type `#24d3ee`, operator `#8c95a1`, variable `#e2e8f0`, punctuation
`#64738b`. `markdown.code` falls to 2.65%, because `baseHighlight` only catches
captures the style does not name and the widened map now names most of them.

after

## Solution

`drawOnCodeSurface` now checks `block.filetype`, which the renderer has already
resolved, against the set of filetypes `@opentui/core` bundles a grammar for. If
there is one, it keeps the code background and returns without touching `fg` or
`drawUnstyledText`, letting per-token colours through, and sets
`baseHighlight = 'markup.raw.block'` so captures the style does not name still
read as code rather than as prose. If there is not, the path is byte-for-byte
what it is today.

`markdown.keyword`, `string`, `comment` and `number` are derived generically in
`buildMarkdown` from existing spec fields through `ensureContrast`, so all eight
themes get them without hand-authored palettes.

**No new dependency.** The grammars are already in the tree.

### Scope, stated plainly

The issue names C++, Rust, Go, Python and TypeScript. `@opentui/core` bundles
grammars only for `javascript`, `typescript`, `zig`, `markdown` and
`markdown_inline` (plus the `javascriptreact` / `typescriptreact` aliases). So
this delivers TypeScript and JavaScript; **Rust, Go, Python and C++ still render
as plain text** and #575 should stay open for them.

That is the dependency cost the issue asked to measure first:

- Bundled `.wasm` sizes: javascript 402KB, markdown 412KB, markdown_inline 416KB,
  zig 676KB, typescript 1.38MB. `assets/` totals 3.3MB, carried today as
  `@opentui/core`'s transitive weight, not `@vibesys/tui`'s own.
- `@vibesys/tui`'s `files` allowlist is `["dist", "package.json", "README.md"]`,
  so vendoring four more grammars (~400KB-1.4MB each) would add directly to our
  published package.
- `@opentui/core` exposes `addFiletypeParser()` / `addDefaultParsers()` for local
  `.wasm`+`.scm`, and `DownloadUtils.downloadOrLoad()` for fetching with a local
  cache instead of vendoring.

### Correctness properties

- A fence whose language has a bundled grammar renders in multiple token colours.
- A fence with an unknown, unsupported, or absent language renders exactly as it
  does today: readable flat text on the code surface, never blank, including
  while streaming.
- All nine mapped families clear their theme's contrast floor against
  `markdown.codeBackground` in all eight themes, including both high-contrast
  themes. `punctuation` is held to the lower `SUBTLE_TEXT_MIN_CONTRAST` floor
  on purpose, the same one `textSubtle` uses: it is brackets and separators,
  not words.
- An unsupported filetype is now distinguishable from a broken grammar. Both
  used to render as the same silent plain text.

### Testing

`pnpm --dir clients/tui test`: 795 pass, 0 fail.

## Deliberately out of scope

**Vendoring more grammars.** C++, Rust, Go and Python, which #575 names and
which are what this repo's examples actually contain (122 Python, 17 Rust, 16
Go, 3 C++ files), still render as plain readable text. Vendoring was measured
and declined: 5.0M of wasm for those four, 13.8M for eleven, against 5.5M for
all ~190 highlight.js languages, and `tree-sitter-cpp`'s queries are a 939-byte
delta over C's that yields one highlight if concatenated wrong. Breadth is
filed separately, and `lowlight` through `CodeRenderable`'s `onHighlight` hook
is the recommendation there, not more vendoring.

**A full-width code surface.** `CodeRenderable` colours only the cells its text
buffer draws, so a short line's trailing columns stay bare canvas rather than
code surface, and the parent classes expose no rect-filling background. A
wrapping `BoxRenderable` was tried and reverted: the renderer computes a
block's bottom margin itself (`getInterBlockMargin`, plus a regex on the raw
token), and replacing the block it tracks drops that margin, leaving the fence
flush against the next paragraph. The pre-existing "keeps the margin the
renderer puts after a fenced block" test catches it. Filed separately.

**A tree-sitter predicate that is not evaluated.** Probing the after frame's
cells, every plain identifier lands in the `type` family, which is why cyan is
about 45% of the crop's ink and `variable` is nearly invisible.
`typescript/highlights.scm` guards its identifier-to-`@type` rule with
`(#lua-match? @type "^[A-Z]")`, so `ring` and `next` should resolve to
`variable`; only `property_identifier` and `variable.parameter` actually do.
`@opentui/core`'s tree-sitter client appears not to evaluate the predicate.
Pre-existing and cosmetic, not introduced here, but it caps how much the
widened map can actually show.

Base is #652.


Nano-AI:
Scope settled, and it is narrower than this branch originally aimed at.

**Vendoring more grammars: measured and declined.** The plan was to vendor C++, Rust, Go and Python. That work reached 5.0M of wasm in `clients/tui/assets/` plus a `vendor-grammars.sh` and was then dropped. The numbers: 5.0M for four languages, 13.8M for eleven, against 5.5M for all ~190 highlight.js languages. `tree-sitter-cpp`'s queries are a 939-byte delta over C's, so cpp alone yields 1 highlight where `["c.scm", "cpp.scm"]` yields 14, and every inheriting grammar needs checking the same way. tree-sitter's cost is per language and this repo's language set is open-ended, so the curve is the wrong shape.

Breadth is filed separately, and the recommendation there is `lowlight` through `CodeRenderable`'s `onHighlight` hook rather than more vendoring. `onHighlight` receives the tree-sitter result plus `{content, filetype, syntaxStyle}` and returns `SimpleHighlight[]`, which is `[startByte, endByte, scope]`, so `if (highlights.length) return highlights` keeps real-parser accuracy for the five bundled grammars and falls through for everything else. No fork, no replacement of opentui's code path.

**A full-width code surface: tried and reverted.** `CodeRenderable` colours only the cells its text buffer draws, so a short line's trailing columns stay bare canvas, and neither `TextBufferRenderable` nor `Renderable` exposes a rect-filling background. Wrapping the fence in a `BoxRenderable` is the obvious fix, and it breaks the renderer's margin tracking: the renderer computes a block's bottom margin itself (`getInterBlockMargin`, plus a regex on the raw token), so replacing the block it tracks leaves the fence flush against the following paragraph. The pre-existing `keeps the margin the renderer puts after a fenced block` test catches it. Filed separately rather than shipped as a workaround.

**What landed** in `a696cb8`: the discarded highlight signal is surfaced (`startHighlight()` read `result.highlights ?? []` and nothing else, so a missing parser and a broken grammar were the same silence), and the token map widened from four capture families to nine. `pnpm --dir clients/tui test`: 795 pass, 0 fail.

Frames re-shot against `a696cb8`. Both use this branch's fixture, passed explicitly, because the base's fixture has no comparable fence and the two frames would otherwise be showing different text.

Issue #567, opened 2026-09-03
TUI: the rounds strip stops working once a run is long


Nano-AI:
### Workstream

CLI/DX

### Problem

`[ r1 ]` is legible at one round and degrades from there. The horizontal chip strip overflows into `‹ 12` / `4 ›` markers whose counts go stale, and each chip carries almost no information: a round number and, for the active round, a ticking elapsed. No outcome, no hypothesis, no measured delta.

Navigation works today, so this improves a frequent workflow rather than unblocking one.

### Desired outcome

An operator can move between rounds and see the run's shape, at any round count.

### Constraints or design considerations

The question to settle first is whether the control's job is **navigation** or **overview**, because that decides the form:

- **Editor-style tabs**, the way an editor shows open buffers. Familiar, and `[ r1 ]` already hints at it. But tabs imply a small user-curated set, and rounds are machine-generated and unbounded.
- **Dropdown.** Scales to any count, costs one row at rest, but hides the trajectory, which is what an operator most wants to see across rounds.
- **Toggle or stepper.** Cheapest, matches `[`/`]` today, gives no overview.
- **Slider.** Suits a long ordered sequence and hints at position, but reads as continuous for what is a discrete list.

#518 and open PR #539 replace the strip with a vertical rail and left-to-right drill-down. That is a different answer to the same problem and must not be duplicated: resolve against it first. If the rail lands, this may reduce to a per-round label change inside it.

Any binding change here also collides with the keybinding standardization work.

### Acceptance criteria

- [ ] Rounds remain navigable at counts where the current strip overflows.
- [ ] Whatever is shown per round is accurate at all times, including overflow counts.
- [ ] Bindings follow `docs/contributing/tui-conventions.md`.

### Verification approach

Replay a long recorded run through the harness and check the control at several widths and round counts.

### Parent, dependencies, or related work

Parent: #284. Overlaps #518 and PR #539.

AyanBinRafaih:
Closing as resolved. The horizontal chip strip this issue describes no longer exists.

- #539 deleted `round-strip.ts` and replaced it with a rail that recomputes its window and the `‹ n` / `n ›` overflow counts from live state on every render, which removes the stale-count complaint.
- #645 ("Addresses #567") added the missing per-round outcome as a failed-judge verdict glyph.
- #686 replaced the rail with `clients/tui/src/ui/round-tabs.ts`, carrying the glyph forward and giving every tab an outcome glyph plus a live metric.

[...]

PR #645, opened 2026-09-09
fix(tui): show a failed judge verdict in the round rail

Nano-AI:
## Problem

Addresses #567. A rail row said how a round ended, never what the judge decided
about it, so a round whose `judge_verdict` was `fail` wore the same solid check
and the same `done` as one that passed.

Both frames below replay the same run with a `fail` verdict on round 1:

`▸r1 ✓ done 3s` becomes `▸r1 ✗ done 3s`.

Two of the three complaints in #567 no longer apply and are **not** addressed here:

- The stale `‹ 12` / `4 ›` overflow counts were fixed by #539. `railWindow`
  recomputes them from the live window every render, so they cannot go stale.
- Navigation at high round counts works.

### Before / after

The round rail, 150x40, dark, on a run whose round 1 carries
`judge_verdict: "fail"`. Base is #644.

| Before (this PR's base) | After |
| --- | --- |
| before | after |

## Solution

`roundMetric` already calls `hypothesisRoundFor(state, round.number)` for
`perf_delta_pct`, and the `HypothesisRound` it returns carries `judge_verdict`,
so this needs no new plumbing.

The verdict **replaces** the status glyph rather than being appended. A row
carrying both a check and a cross states two contradictory things, and the status
word already says the round finished. More importantly, compact width (<= 13
columns) keeps only marker, number and glyph, so an appended mark would vanish on
exactly the narrow rail where an operator can least afford to lose the verdict.
The swap costs no columns, so the longest label is unchanged and this file's
no-truncation-by-construction property still holds.

This extends the ordering `profileSkipped` already established: how a round was
judged outranks how it measured, which outranks that it finished. Only `fail`
takes the cross; `pass` matches what the check already implies, and `deferred` or
an unjudged round has no claim to contradict it.

### Also fixed, and no longer in this diff

An earlier revision of this branch fixed `#syncElapsedTimer` rebuilding the
active round's label every second with `compact` hardcoded `false`, so on a
compact rail the running round was overwritten with the full-width string once
per second until the next full render.

**#624 fixed that first**, by threading `compact` through `#runningRound`, and
this branch now carries #624's version verbatim
(`const {round, state, text, compact} = this.#runningRound;`). Nothing of it
remains in this PR's own diff. Kept here only so the claim is not read as still
outstanding. Same class of bug as #551, whose PR #637 is closed as superseded by
#624 for the same reason.

### Correctness properties

- A completed round whose judge verdict is `fail` shows `✗`; `pass`, `deferred`
  and unjudged rounds are unchanged.
- The verdict is carried by a glyph, not a colour, so it survives the two
  high-contrast themes (`tui-conventions.md`, WCAG 1.4.1).
- The label's column count is unchanged, at full and compact width.
- The elapsed timer and `render()` derive `compact` from the same width.

### Testing

`pnpm check:ts`, `pnpm check:clients`, `pnpm build:clients` pass.
`bun test round-rail.test.ts`: 29 pass, 0 fail.
`pnpm test:clients`: 701 pass, 0 fail.

No recorded fixture currently contains a failing verdict (only `pass`), so the
screenshots above were captured with a `fail` verdict injected locally into
`markdown.experiments.json` and reverted afterwards. No fixture change is in this
diff. Worth considering separately whether a recorded fixture should cover this
state, since nothing exercises it visually today.

Stacked on #644.


Nano-AI:
Force-pushed `840c82a` -> `70acaee`: a rebase onto #644's new tip (`a696cb8`), nothing else. `git range-diff 82284bc..840c82a a696cb8..70acaee` reports the single commit as unchanged (`=`), so this PR's own diff has not moved.

Issue #642, opened 2026-09-09
TUI: a rounded box fills its corner cells, so the fill's square corners sit outside the border arc


Nano-AI:
## Observed behavior

A box that draws a rounded border and also sets a `backgroundColor` paints that
background across its full rectangle, including the four corner cells. The
corner cell then renders the arc glyph (`╭`, `╮`, `╰`, `╯`) in the border colour
*on top of* the box's own fill, so the fill occupies the whole corner cell and
reads as a square corner with an arc drawn inside it.

The visible result is that the fill and the border disagree at every corner: the
box is a rectangle of fill with a rounded outline sitting inside it, rather than
a rounded shape. The filled region is the union of the rectangle and the rounded
frame, where it should be their intersection.

This is visible in a real terminal capture already attached to #637: the
selected `implementer` card in the agent map carries `selectedSurface #1e293b`
behind a rounded `borderFocus #22d3ee` frame, and the grey fill is plainly
square at each corner while the cyan arc curves inside it.

The same pairing of a rounded border with a distinct fill occurs in:

- `clients/tui/src/ui/agent-map.ts:281,287` and `:348,352` (selected agent card)
- `clients/tui/src/ui/chat-composer.ts:168,170`
- `clients/tui/src/ui/chat-overlay.ts:74,76`
- `clients/tui/src/ui/command-input.ts:81,83`
- `clients/tui/src/ui/error-banner.ts:38,42`
- `clients/tui/src/ui/conversation.ts:337,341`

It is most obvious where the fill contrasts most with what is behind it, so the
selected agent card and the command input show it worst.

## Expected behavior

A corner cell of a rounded box takes the background of the surface *behind* the
box, not the box's own fill, so the arc glyph reads against the surrounding
canvas and the corner actually looks rounded. The fill should stop at the
rounded frame rather than extending into the corner cells the frame does not
cover.

## Reproduction

1. `clients/tui/dev/mock-ui.sh --tmux 150x40 --speed 0`
2. `/open-round --1`, then move focus into the Agents pane and select an agent
   so its card is drawn with `selectedSurface`.
3. Look at the four corners of the selected card. The grey fill is square; the
   cyan arc curves inside it.

Also reproducible without a run by looking at the capture on #637.

## Affected subsystem

CLI and packaging

## Environment

Commit: 12e807a
OS: macOS 15 (Darwin 25.4.0), also present in Linux CI captures
Hardware: arm64
Backend: not applicable, this is frontend rendering
[...]

PR #647, opened 2026-09-09
feat(tui): replace transcript card borders with a top-edge divider

### LINKED ISSUE #565: TUI: the transcript spends its rows on card chrome rather than separation

Nano-AI:
## Problem

Closes #565.

Every transcript entry is a rounded bordered box with its own `marginTop` and
left/right padding, so a one-line status costs 5 rows to say one thing and a
screenful reads as a stack of containers rather than a sequence of events. The
card's padding also disagreed with the pane around it: cards inset their content,
the panes containing them do not, so an entry sat one layer deeper than a
non-card sibling in the same pane.

The same 16 rows of transcript, 150x40, dark theme, both starting on an entry
boundary and ending on the same pane floor. Three filled four-sided cards before,
five rule-separated entries after.

**Before (`main`)**

**After**

Note the heading in the after frame: the role holds the left edge and the run id
is pushed right, which is the second commit on this branch.

Role stays legible on the rule rather than the fill: `conversation.analysis.border`
for judge, success green for the benchmark. Card background across the crop goes
from 48.72% success fill plus 24.36% analysis fill to 100% canvas; pinned at
`22,70`, `#132f30` to `#0f172a`. Across the full pane the same run shows 6
entries before and 10 after.

### Before / after

150x40, dark, full window. Base is #646.

**Before** — `judge · round-1-retry-1-judge` is restated on every one of six
consecutive entries, and each `Benchmark` / `round-1 · PASS` card draws a
four-sided border.

before

**After** — one heading for the whole judge run, role pinned left and run id
right, with fresh chrome only where the speaker changes. Cards are separated by
a top rule instead of a border.

after

## Solution

The issue prices two directions and asks that one be chosen. This takes
**dividers**, not the collapsible accordion: a rule between entries separates as
well as a border on four sides at a fraction of the row cost, and the accordion
would add interaction state, a keybinding, and a rule for what a collapsed entry
shows, none of which this issue asks for.

Each entry is now `border: ['top']` / `borderStyle: 'single'` with no margin and
no card padding, relying on the pane's existing inset. A one-line entry costs 3
rows instead of 5. The `'status'` kind's old borderless carve-out is gone, so the
transcript has one boundary grammar rather than two.

Selection is unaffected: the heading's `▸ ` marker and `textStrong` colour carry
it independently of border colour, per `tui-conventions.md`'s rule that focus is
never carried by colour alone.

### Scope note

The resting divider colour now runs through `ensureContrast` at the same 3:1
floor `textSubtle` uses. `roleAccents` skips that check, and 5 of 64 role/theme
combinations sit under 3:1. That is a pre-existing gap, but a four-sided border
could lean on its own area to stay visible at marginal contrast and a one-row
rule cannot, which is what makes this change the point to close it.

Rebased onto #646, which now enforces that a box may have a rounded border or a
fill, never both. This card ends up on the compliant side of that rule for a
different reason: it has no rounded border left to reconcile, because #565
replaces the four-sided frame with a single top-edge rule. The `cardBehindColor`
option an earlier version of #646 threaded through here is gone with it.

**Conflicts with #640 as written.** #640 keeps the four-sided border and replaces
the fill with a left-edge role stripe positioned against that border. This removes
the border it measures from. The two are competing answers to the same question
and one has to give; flagging rather than resolving it unilaterally.

### Adjacent scope: the heading row

A second commit pins the role to the left edge and the run id to the right. An
entry heading rendered one string, `agentKind · roundLabel`, so a column read
`judge · round-1-retry-1-judge` over and over with both starting at the left
edge, pushing the eye a variable distance rightward to find where the detail
ended. The heading box was already `flexDirection: 'row'` with
`justifyContent: 'space-between'` and only ever had one child, so this needs no
new layout: it reads the `agentKind` and `roundLabel` fields already on
`ConversationEntry` rather than splitting the joined label back apart. An entry
without that pair, such as a `result` labelled `round-1 · PASS`, is unchanged.

This is horizontal scannability rather than row cost, so it is adjacent to #565
rather than part of it. It is folded in here because it rewrites the same heading
row this PR already touches, and a separate branch would only conflict with this
one on `conversation.ts`.

### Second commit: chrome once per speaker run

An entry draws its divider and heading **only when the speaker changes**;
consecutive entries from the same agent render as content lines alone. Speaker
is `agentKind` + `roundLabel` when both are present, else `label ?? kind`,
derived in one place so the heading and the grouping cannot disagree.

**Measured on the `bad-cpp-round1` fixture at 148 columns** (131 entries, 8
speaker runs): **1142 rows to 898, a 21% cut.** A run of N same-speaker entries
costs `2 + N` rows instead of `3N`.

Two things a reviewer should look at:

- **Selection moved out of the heading**, because an entry inside a run has
  none. It is now a left-edge rule in `borderFocus`, with unselected entries
  reserving that column via `paddingLeft: 1`, so nothing shifts as the cursor
  travels. The alternative — give a selected run member its own heading —
  inserts a row under the cursor, which `tui-conventions.md` forbids.
- **A second surface changed.** The chat pane reuses `ConversationView`, so
  consecutive chat messages with the same label now share a heading too. Same
  rule, and its tests pass, but it is a visible change somewhere this PR does
  not otherwise touch.

Rendering here is incremental, so this was the hard part: append builds each new
card against its predecessor; prepend re-renders the entry that used to be first
iff the newly revealed entry above it is the same speaker; replace rebuilds the
changed card and, if its speaker changed, the single card below it. No path
rebuilds the transcript. Three tests assert that append, prepend and replace
each produce the same tree as a fresh render — mutation-checked, so breaking any
one path fails exactly one test.

### Third commit: source tags and run ids take the card's label colour

Review feedback: the `[git-tracking]` / `[framework-validation]` prefixes blended into
the event stream, and the run id (`round-1-retry-1-judge`) sat in grey instead of the
card's own colour.

- The run id now uses the same expression as the role on the left of the heading
  (`palette.label`, `textStrong` when selected), so a heading reads as one colour.
- A bracketed tag at the start of a line (`^\[[A-Za-z0-9][\w-]*\]`) renders in
  `palette.label`; the rest of the line keeps `palette.content`. Brackets anywhere
  else are untouched. Only the plain-text path is affected (diagnostic entries);
  Markdown entries are not.
- The content stays one wrapping `TextRenderable` fed a `StyledText`, the first use
  of `StyledText` in the client. The header's row-of-spans pattern cannot word-wrap.

| Before | After |
| --- | --- |
| before | after |

### Correctness properties

- An entry costs 3 rows plus its content, uniformly across kinds.
- An entry adds no horizontal inset beyond its containing pane's.
- The divider clears 3:1 against the canvas in all 8 themes, and no two roles
  collapse onto one colour.

### Testing

`pnpm check:ts`, `pnpm check:ts-architecture`, `pnpm check:clients` all clean.
`pnpm test:clients`: 689 TUI pass, 0 fail (684 on `main`, +1 from #642, +4 here).
After the third commit: 762 pass, 0 fail, 20 of them new in `conversation.test.ts`
(tagged, untagged, mid-line bracket, and run-id colour across all themes).

New `conversation.test.ts` pins the 3-row cost through the renderer (fails at the
merge base at 5 rows / 15 for three entries) and the two theme properties as
computations over `theme.ts`. Replay harness at 150x40 shows 6 fully visible
entries before and 10 after; at 150x18, 2 before and 3 after. Light theme
identical, geometry does not depend on it.

### Not fixed here

In 5 theme/role pairs the role accent already equals `borderFocus`, so selecting
that role's entry produces no colour change. Present on `main`, untouched by this
diff, and a `theme.ts`-level issue rather than a density one.

### DIFF
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -72,7 +72,15 @@ import type {ClipboardCopyResult, SelectionClipboard} from './clipboard.js';
 import {renderDesignSummary} from './design-log.js';
 import {paneTitle} from './focus.js';
 import {headerBackground} from './header.js';
-import {contrastRatio, listThemes, resolveTheme, THEME_NAMES, type ThemeName} from './theme.js';
+import {
+  contrastRatio,
+  ensureContrast,
+  listThemes,
+  resolveTheme,
+  SUBTLE_TEXT_MIN_CONTRAST,
+  THEME_NAMES,
+  type ThemeName,
+} from './theme.js';
 
 const cleanup: Array<() => void> = [];
 
@@ -1798,10 +1806,14 @@ describe('OpenTUI presentation', () => {
     const frame = await testRenderer.waitForFrame(value => value.includes('2 passed'));
     expect(frame).toContain('→ Bash pytest');
     expect(frame).toContain('← 2 passed');
-    // Header housing, agents pane, transcript frame, and the card's call and
-    // result regions: the rail is absent because this fixture has no rounds.
-    // A focused pane draws a heavy corner, so the count is over both styles.
-    expect(frame.match(/[╭┏]/g)).toHaveLength(5);
+    // Header housing, agents pane, transcript frame, and the command box: the
+    // rail is absent because this fixture has no rounds. A focused pane draws a
+    // heavy corner, so the count is over both styles. The transcript card used
+    // to contribute a fifth: #565 replaced its four-sided border with a
+    // top-edge rule, which draws no corner glyph at all. The call and result
+    // regions #620 added are bands inside the card, not bordered boxes, so they
+    // never contributed one.
+    expect(frame.match(/[╭┏]/g)).toHaveLength(4);
   });
 
   it('renders a typed command payload with labeled stderr and exit code', async () => {
@@ -1975,10 +1987,12 @@ describe('OpenTUI presentation', () => {
     const frame = await testRenderer.waitForFrame(value => value.includes('[codex error]'));
     expect(frame).toContain('[codex turn started]');
     expect(frame).toContain('driver: agentshim');
-    // Header housing, agents pane, transcript frame, command bar, and the one
-    // card the error diagnostic still earns. The two quiet diagnostics draw no
-    // border at all, so they add nothing to the count.
-    expect(frame.match(/[╭┏]/g)).toHaveLength(5);
+    // Header housing, agents pane, transcript frame, and the command bar. The
+    // two quiet diagnostics draw no border at all, so they add nothing to the
+    // count, and the one card the error diagnostic still earns no longer adds
+    // one either: #565 turned that four-sided border into a top-edge rule,
+    // which has no corner glyph.
+    expect(frame.match(/[╭┏]/g)).toHaveLength(4);
   });
 
   it('renders an unanswered tool call once, with no response band', async () => {
@@ -1997,8 +2011,9 @@ describe('OpenTUI presentation', () => {
     // The response band is the only thing that draws a left arrow followed by
     // text; the footer's key hint is the glyph pair, not this.
     expect(frame).not.toContain('← ');
-    // One card: the header, the two panes, the command bar, and this turn.
-    expect(frame.match(/[╭┏]/g)).toHaveLength(5);
+    // The header, the two panes, and the command bar. This turn's card no
+    // longer adds a corner: #565 made it a top-edge rule instead of a border.
+    expect(frame.match(/[╭┏]/g)).toHaveLength(4);
   });
 
   it('collapses, prettifies, and expands long JSON tool responses', async () => {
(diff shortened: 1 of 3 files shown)


Nano-AI:
#640 is closed as superseded by this PR, so the "conflicts with #640 as written, flagging rather than resolving it unilaterally" note in the body is settled rather than open.

Both answered the same question, how a transcript card shows role once the fill is gone. #640 kept the four-sided border and added a left-edge stripe against it; this one deletes the border, moves role onto the divider, and cuts an entry from 5 rows to 3. The stripe was positioned against a border this PR removes, and both branches added their own `conversation.test.ts`, so they could not both land.

Worth recording: the stripe's one advantage was running the full entry height, so role stayed visible when a long entry scrolled past its heading. If that turns out to matter in use, put a stripe on this PR's divider rather than reviving the bordered card.

PR #651, opened 2026-09-09
fix(tui): solve the modal scrim strength per theme instead of a fixed opacity

Nano-AI:
> **Scope changed under review.** #634 merged as `3d2afe7` while this branch was
> open and now owns the scrim itself: `OverlayView` paints a full-screen box at
> zIndex 19, raised on the same predicate this branch had derived. The
> `ScrimView` module and the `app.ts` wiring this PR originally carried are
> redundant against it and are dropped rather than added a second time. What
> survives is the part #634 does not have: the per-theme strength derivation.
>
> Provenance: reopened from #639, position 4/5 of stack #649. Its old base #640
> is closed as superseded by #647; GitHub will not let a stacked PR change its
> base and `gh stack link` will not drop a member, so retargeting needed a fresh
> PR from the same branch. Base is #648.

## Problem

#634 paints `theme.canvas` at a fixed `opacity: 0.7` for all eight themes, and
the themes do not start from the same contrast. One blend deep enough to recede
Solarized Dark's 5.6:1 body text leaves High Contrast Dark's 21:1 fully
readable, so a single literal cannot serve both ends of the set.

Refs #566.

## Solution

`scrim(theme)` in `ui/theme.ts` solves the strength per theme for a fixed
residual against that theme's own `canvas`. Measured strengths run 0.374 to
0.655, and the binding theme is Solarized Dark, not the high-contrast pair.
`OverlayView` takes its background and opacity from the solver rather than from
the literal, so the dim has one place it is computed.

Pulling toward the theme's own `canvas` also means neither the light nor the
high-contrast palettes get a tone they exclude.

### Architecture

`ui/theme.ts` owns the derivation; `ui/overlay.ts` consumes it. No new module,
and `app.ts` is untouched: the predicate that decides when the scrim is raised
already lives there from #634.

### Before / after

`/help` open over a zoomed round-1 transcript, 150x40, dark, full window.
Re-shot 2026-09-10. Before is this PR's base #648 at `bd16a38`, showing #634's
fixed 0.7; after is this branch at `9d04ec7`, showing the derived strength. The
scrim strength is the only difference between the frames.

**Before**

before

**After**

after

Layout is character-for-character identical; only colour moves, which is the
point. The scrim is a translucent paint blended per cell, so opening a modal
shifts nothing. Sampled out of the written PNGs:

| cell | role | before, fixed 0.7 | after, derived |
| --- | --- | --- | --- |
| `1,2` ink | header accent `#22d3ee` | `#154f64` | `#165b71` |
| `1,24` ink | `textPrimary` `#e2e8f0` | `#4e5565` | `#5b6271` |
| `19,2` fill | analysis card `#202146` | `#141a33` | `#151b34` |
| `19,3` ink | analysis label `#b193f8` | `#3e3c67` | `#494474` |

Solving the blend toward `theme.canvas` from `textPrimary` gives alpha **0.701**
before and **0.640** after, against the 0.6375 the solver derives for dark.

**A note for anyone re-shooting this:** vs-snap's colour audit cannot pass on
this change. Every dimmed cell is a blend, so by construction it is not a
`theme.ts` colour and the audit reports it OFF-THEME against the nearest named
one (`! dominant foreground #5b6271 is not a theme colour`). Verify by reading
the PNGs and sampling named cells, not by the audit.

### Correctness properties

- The scrim repaints cells and lays out nothing, so opening a modal moves no
  character outside the modal's rectangle.
- One scrim serves all three modals, so stacked modals never dim each other.
- Every theme keeps modal body text at its own contrast floor.

### Testing

`app.test.ts` asserts, for each of the eight themes, that every cell outside the
modal's rectangle is its own colour blended toward that theme's scrim, that
dimmed cells appear above, below, left and right of the box, that no character
outside the box moves, and that the modal's body keeps `textPrimary` on
`elevatedSurface`. Held against #634's fixed 0.7, **four of the eight fail**
those assertions: dark, light, solarized-dark and solarized-light.
`theme.test.ts` pins the derivation itself.

`pnpm test:clients`: 0 fail; the TUI suite is 775 pass across 30 files.

### Known follow-up, inherited not introduced

The scrim does not rise for the narrow-terminal pane fallback: the predicate at
`app.ts:543` covers the three real modals only. An earlier revision of this
branch wired the fallback in, on the reasoning that it is a modal in everything
but name, and measuring it showed that erases the Agents and Transcript frames
rather than dimming them (`paneBorders` finds only `▸ Performance` with the
fallback open and the scrim raised). That predicate now belongs to #634, so this
is recorded as a real defect in the interaction rather than something this PR
changes.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -178,7 +178,11 @@ modal they used before panes existed. That modal is the same surface as the
 pane, so it keeps the pane's title and its focus marker rather than reading as a
 generic dialog. The layout re-flows on resize in either direction.
 
-`/help`, `/theme`, and errors stay modal.
+`/help`, `/theme`, and errors stay modal. While any of them is open, a scrim
+dims the entire screen behind it, so the modal is the only surface left at full
+contrast and the operator can tell where a keystroke will land. The scrim is a
+translucent paint on an absolutely positioned box that joins no flex row: the
+background keeps every character where it was, and closing restores it exactly.
 
 ### Experiment chat
 
@@ -278,6 +282,11 @@ the Markdown palette are derived from that core, and each derived foreground is
 pushed toward the nearest extreme until it clears the theme's `minContrast`
 against the surface it actually sits on. The `dark` theme additionally pins its
 derived values to the original literals so the baseline is byte-identical.
+The modal scrim is derived the same way: it pulls the background toward the
+theme's own `canvas`, and its strength is solved per theme so body text lands on
+WCAG's large-text floor whatever it started from. One fixed blend cannot do
+that, because a blend deep enough to recede Solarized Dark's 5.6:1 body text
+leaves High Contrast Dark's 21:1 fully readable.
 Status meaning never depends on color: agent phases carry a marker glyph and
 the spelled-out status, todos carry a per-status marker, and only the running
 round shows elapsed time.
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -77,8 +77,10 @@ import {
   contrastRatio,
   ensureContrast,
   listThemes,
+  mix,
   resolveTheme,
   SUBTLE_TEXT_MIN_CONTRAST,
+  scrim,
   THEME_NAMES,
   type ThemeName,
 } from './theme.js';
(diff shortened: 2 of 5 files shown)


Nano-AI:
Repointed and rescoped. #634 merged as `3d2afe7` while this was open and now owns the scrim itself: `OverlayView` paints a full-screen box at zIndex 19, raised on the same predicate this branch derived. The `ScrimView` module and the `app.ts` wiring this PR originally added are redundant against it and have been dropped rather than added a second time.

What survives is the part #634 does not carry. It paints `theme.canvas` at a fixed `opacity: 0.7` for all eight themes, and the themes do not start from the same contrast: one blend deep enough to recede Solarized Dark's 5.6:1 body text leaves High Contrast Dark's 21:1 fully readable. `scrim(theme)` in `ui/theme.ts` solves the strength per theme for a fixed residual against that theme's own canvas, and `OverlayView` now takes its background and opacity from the solver rather than from the literal, so the dim has one place it is computed.

Held against the fixed 0.7, four of the eight themes fail this branch's own assertions (dark, light, solarized-dark, solarized-light).

The title is updated to match the surviving scope. The body still describes the original `ScrimView` shape and is being rewritten along with fresh frames; the commit message is current. Base is unchanged: #648, position 4/5 of stack #649.

PR #652, opened 2026-09-10
fix(tui): paint one background instead of a surface ladder

### LINKED ISSUE #574: TUI: the surface ladder is inverted, so panes read as recesses rather than panels

Nano-AI:
## Problem

Closes #574.

The theme carried two backgrounds: `canvas` for the root frame and
`elevatedSurface` for every pane, modal and the header. In the dark themes the
"elevated" surface was the *darker* of the two, so panes read as recesses cut
into the page rather than panels on it, which is what #574 reported.

Two symptoms followed from the same pair, and the second is visible without
comparing panes at all:

- The header calls `headerBackground()`, which returned `elevatedSurface`, so it
  read as a different colour from the frame around it.
- The key-help line sets no background of its own, so it fell through to the
  root's `canvas`. On `main`, against panes painted the darker surface, that left
  a **pale rule across the bottom of the screen**.

**What this PR's own frames show, and what they do not.** By the time this
change applies, #646 moves the pane fills inside their borders rather than
dropping them, so the panes still paint `elevatedSurface` in the base: measured
84.68% of the crop on the 2026-09-10 frames, not the 450 of 6000 cells an
earlier revision of this paragraph claimed. The before frame therefore shows the
pale bottom rule and the pale margins the Problem section describes, rather than
the header alone disagreeing with its surroundings. Both symptoms come from the
same pair of tokens; this removes the pair.

### Before / after

Full window, 150x40, dark. Base is #651.

**Before** — the three header rows are `#020617` while the panes, the frame and the key-help
row are all `#0f172a`, so the header reads as a different colour from everything
around it.

before

**After** — all four sampled points are the single background `#020617`, asserted at cells
`1,60`, `20,10` and `39,60`.

after

## Solution

The ladder is removed rather than re-ordered. `elevatedSurface` is deleted and
`canvas` moves down to the value the panes and header already painted, so there
is exactly one background and nothing left to fall through to that disagrees
with its neighbour.

| theme | canvas before | canvas after |
| --- | --- | --- |
| dark | `#0f172a` | `#020617` |
| light | `#f8fafc` | `#ffffff` |
| solarized-dark | `#002b36` | `#001f27` |
| solarized-light | `#fdf6e3` | `#fffbf0` |
| catppuccin-mocha | `#1e1e2e` | `#11111b` |

The cells the panes and header paint do not change colour; the root comes to
meet them. `selectedSurface` is untouched: selection is state, not a decorative
tint, and it is the one distinction worth keeping.

This supersedes the first attempt on this branch, which raised `elevatedSurface`
above `canvas` and added `ensureStatusContrast` so semantic foregrounds still
cleared 4.5:1 against the raised surface. With one background that second pass
has nothing to guard, but the contrast guarantee it protected is unchanged:
every semantic foreground still clears its theme's `minContrast` against the
single surface, and the regression test a reviewer asked for is kept.

### Stack position

Rebased onto #651 and reopened from #643 (see the note there). It conflicted
with #646 on `chat-pane.ts`, `error-banner.ts`, `experiment-log.ts` and
`app.test.ts`: both PRs decide where fills live. They are two halves of one rule
— #646 says a box may have a rounded border **or** a fill, this says there is
only one fill colour — so they now sit in one line of work instead of colliding
at merge. At the seam, #646 wins on the three rounded panes (their fill is
removed outright, so there is nothing left to recolour) and this PR applies at
the sites #646 kept a fill: the modals, the header and the composer.

### Correctness properties

- Exactly one background token exists; nothing renders a second surface.
- The key-help line and the header carry the same background as the panes.
- Every semantic foreground still clears its theme's `minContrast`.
- Selection remains visually distinct via `selectedSurface`.

### Testing

`pnpm check:ts`, `pnpm check:ts-architecture`, `pnpm check:clients` all clean.
`pnpm test:clients`: **722 pass, 0 fail** at this commit; 737 at the top of the
stack.

New assertion: *"gives the key-help line the same background as the chrome above
it"*, checked against the header's own cells rather than a literal, so the
property survives a future palette move rather than needing the number updated.

The scrim assertion in #651 named `theme.elevatedSurface`; it now names
`theme.canvas`. That is the deleted token, not a behaviour change — the modal is
still the one surface the scrim does not paint over.

### Design note

The modal interior is now `theme.canvas` and the scrim tints toward
`theme.canvas`, so separation rests on the modal's single-line border plus the
scrim being translucent. The contrast floor is still asserted, so this is a
visual-separation tradeoff, not a legibility one. Taken deliberately; if it
reads badly in use the fix belongs on the modal edge, not in a revived surface
ladder.

### DIFF
--- a/clients/tui/README.md
+++ b/clients/tui/README.md
@@ -270,11 +270,17 @@ the input still runs on its own Enter. `/theme <name>` re-themes every view in
 place without opening the list.
 
 `ui/theme.ts` is the only module holding color literals. A theme declares
-semantic roles — `canvas`, `surface`, `elevatedSurface`, `selectedSurface`;
-`textPrimary`, `textMuted`, `textSubtle`, `textStrong`; `border`,
-`borderStrong`, `borderFocus`; `accent`, `info`; `success`, `warning`, `error`;
-per-role conversation card colors; and Markdown/code colors. Views ask for a
-role and never for a color.
+semantic roles — `canvas`, `surface`, `selectedSurface`; `textPrimary`,
+`textMuted`, `textSubtle`, `textStrong`; `border`, `borderStrong`,
+`borderFocus`; `accent`, `info`; `success`, `warning`, `error`; per-role
+conversation card colors; and Markdown/code colors. Views ask for a role and
+never for a color.
+
+`canvas` is the only background the UI paints: every pane, modal, the header,
+the error banner and the composer fill with it, and a box is told from the page
+by its border rather than by a shade of its own. The other two fills name a
+state and not a depth — `selectedSurface` is what the cursor is on, and
+`surface` backs a code block, which has no border to delimit it.
 
 Adding a theme means adding one `ThemeSpec`: a semantic core plus one accent
 per conversation role. Card fills, labels, body text, the tool-call band, and
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -2755,7 +2755,7 @@ describe('theming', () => {
     expect(spanColors(testRenderer, 'implementer')?.fg).toBe(light.conversation.assistant.label);
   });
 
-  it('keeps the dark baseline identical to the pre-theme palette', async () => {
+  it('keeps the dark baseline identical to the pre-theme palette, background aside', async () => {
     const testRenderer = await createTestRenderer({width: 90, height: 20});
     const controller = new FakeController({
       ...initialSessionState(),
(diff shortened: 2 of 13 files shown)


Nano-AI:
One design question this PR should have raised in the body and did not, stated plainly so it is reviewed rather than discovered.

After the collapse the modal interior is `theme.canvas`, and the scrim tints toward `theme.canvas`, so the modal is an opaque patch of the scrim's own colour. Separation now rests on its single-line border plus the scrim being translucent. The contrast floor is still asserted, so this is legibility-safe: it is purely a question of visual separation, not of readability.

Shipping it that way deliberately. If it reads badly in use, the fix is a treatment on the modal edge, not a return to the surface ladder. Two earlier attempts at a ladder were already rejected and should not be revived: raising `elevatedSurface` above `canvas`, and blending a rounded corner cell to the midpoint of fill and canvas.

Also carrying forward from the closed predecessor: the `[P2]` Solarized Dark `warning` contrast finding @AyanBinRafaih raised on #643 was fixed there along with four unreported siblings (`error`, `info`, `success`, `accent`). That thread lives on #643, so it is easy to miss here. Note that the dedicated `ensureStatusContrast` chaining is gone on this branch, not regressed: it existed to run a second pass against `elevatedSurface`, and with one background there is no second surface to contrast against. The single `ensureContrast` pass against `canvas` plus the per-theme `minContrast` property is what holds the floor now.

PR #619, opened 2026-09-05
TUI: close out dead rounds and keep the agent graph stable across kill and resume

AyanBinRafaih:
## Problem

Fixes #514. A killed run left open phases and timings in the TUI, and a resumed implementer could reuse the killed attempt's label. The initial fix closes run-scoped terminal events and handles a hard-kill when the next `run_started` arrives.

Tail subscriptions exposed a remaining case: the server sends run-level spine events before the history floor but omits ordinary execution starts. During later backfill, a fresh prefix fold did not see the resume or terminal boundary, so it could merge an old active execution into the resumed execution. Multi-chunk backfill could also close the old attempt at an earlier chunk's timestamp instead of the actual last pre-resume event.

## Solution

`core-state` retains the run-lifetime spine events (`run_started`, `run_finished`, `run_failed`, and `run_interrupted`) and the closest non-chat predecessor for each resumed-start boundary. Prefix reduction folds the history chunk and separately reconstructs its run-map projection with retained boundaries interleaved by sequence. Only the run-map fields from that reconstruction enter the merge. This preserves the old attempt's closeout status, phase identity, interval, and timestamp without duplicating terminal transcript or diagnostic facts already delivered in the tail.

Prefix round merging retains a run-boundary endpoint across later resumed attempts, while an explicit `round_finished` event remains authoritative.

The existing loop marker and TUI graph changes remain scoped as before: an implementer start marker makes resumed attempt labels distinct, and the graph respects its row budget.

### Architecture

```mermaid
flowchart LR
  S[server tail: spine + suffix] --> C[core-state tail fold]
  C --> B[retained lifetime boundaries]
  H[history backfill] --> P[prefix data fold]
  B --> R[run-map replay in sequence]
  P --> M[prefix merge]
  R --> M
  M --> U[TUI phases, rounds, timings]
```

The backend remains the source of event semantics and execution liveness. `core-state` owns replay and prefix merge behavior; the TUI consumes the resulting semantic state.

### Correctness properties

- Run-scoped terminal events close every open phase, round, and active timing with their semantic status.
- A hard-killed lifetime closes at its last non-chat event when a later `run_started` is folded, so downtime is excluded.
- Full replay and actual spine-bootstrap plus multi-chunk backfill agree for every replay-derived field, including distinct old/resumed canonical execution IDs, phase status, round intervals, and inferred closeout timestamps. `activeExecutions` remains intentionally checkpoint-owned.
[...]

Issue #553, opened 2026-09-02
TUI: Mark the selected row and node with a glyph, not background color alone


AyanBinRafaih:
### Observed behavior

In the experiment log, a selected-but-inactive hypothesis row and every unowned round row differ from their neighbors only by a background-color swap; the leading `▸` marker encodes whether the entry is active, never whether it is selected. In the agent graph, selecting a node changes only border color, background, and text color; its only glyph encodes run status, not selection. On a low-contrast or color-dropping terminal, the operator cannot tell which row or node Enter will act on.

### Expected behavior

Selection should carry a textual or positional marker (a caret such as `▸`) independent of the active/status accent, so the selected element is legible without color, matching the round rail and theme picker which already do this. The background colour differentiation should still be maintained.

### Reproduction

1. Open the Experiments log and arrow the cursor onto any non-active hypothesis or unowned round; the selected row is distinguished by background only.
2. Tab to select an agent node in the graph; the selection is color-only.

### Affected subsystem

CLI and packaging

### Environment

Commit/branch: ayan/518-vertical-rounds-rail (working tree)
OS: Linux 6.1 (Debian), TUI toolchain node>=20 + bun + pnpm, Python via uv (.venv 3.14)
Backend: local Unix-domain-socket supervision service, codex/claude CLI providers
Profiler: none for TUI repros; real profiler only for loop-metric repros
Verification harness: clients/tui runtui.sh (codex provider, queue-rs example, max 2 rounds)

### Impact

Blocks normal use

### Relevant logs

```shell
`clients/tui/src/ui/experiment-log.ts:597` sets `const marker = entry.active === true ? '▸' : ' ';` (tied to active, not selection), and the row selection style at lines 375/391 is a background swap (`selectedSurface`, never run through `ensureContrast`). `clients/tui/src/ui/agent-map.ts:255-268` styles selection with border/background/text color while `STATUS_MARKER` encodes only `phase.status`.
```

### Related issues or pull requests

#284

### Pre-submission checks

- [x] I searched for an existing issue describing this problem.
- [x] I removed sensitive information from this report.

Issue #514, opened 2026-09-01
TUI: round agent graph renders a different, growing topology after a run is killed and reopened, and never closes out dead rounds


AyanBinRafaih:
### Observed behavior

Killing a run before completion and reopening it changes the round's agent graph instead of preserving it, and several pieces of interrupted state are never closed out. Reproduced live: a 2-round codex run on queue-rs was killed (SIGTERM via launcher teardown) while the round-1 implementer (attempt 3, label `round-1-retry-2-implementer`) was active. Before the kill the Agents pane showed "6 agents · 1 active · 4 done · 1 waiting" with two orchestrator nodes fanning into two stacked implementer nodes, then judge, then a pending profiler. After relaunching with `--resume`, the same round rendered "7 agents · 1 active · 4 done · 1 interrupted · 1 waiting": a third stacked implementer node marked `! interrupted` appeared, a fourth implementer node (the fresh attempt spawned by the resumed loop) appeared with the exact same round label `round-1-retry-2-implementer` as the interrupted one, and the complete-bipartite edge routing between the orchestrator column and the now three-to-four-node implementer column degenerated into an unreadable `─┬┬─▶`/`├┼─▶`/`└┴─▶` hairball. The rounds strip chip kept ticking (`[ r1 16m 20s ]` and counting) across the kill, counting wall-clock time during which no process existed. The pending profiler node from the killed process was never marked interrupted, and the round itself was never closed: the durable event log ends with `run_interrupted` and `run_failed` events that carry `agent_kind: null` and `round_label: null`, which the TUI phase reducer drops on the floor.

[...]

PR #654, opened 2026-09-10
feat(tui): list a round's agents in the rounds rail instead of the graph pane

### LINKED ISSUE #653: TUI: list a round's agents in the round rail instead of the agent graph pane
### Workstream

CLI/DX

### Problem

The round view spends most of its width on an agent graph whose edges carry no information.

`GRAPH_SHARE = 0.55` (`clients/tui/src/ui/agent-map.ts:44`) targets 55% of the terminal for the Agents pane, floored by the graph's own geometry: `graphPaneBounds` (`agent-graph.ts:68-75`) asks for `stages * NODE_WIDTH_MIN + (stages - 1) * GUTTER + PANE_CHROME`, and `GUTTER = 5` spends 15 columns on connectors alone in a four-stage round. Computed from that code, a four-stage round at a 150-column terminal gives the graph 75 columns and leaves the transcript 47, five columns above its `TRANSCRIPT_MIN = 42` floor:

| Terminal | Rail | Agents pane | Transcript |
| --- | --- | --- | --- |
| 120 | 28 | 30 | 62 |
| 150 | 28 | 75 | 47 |
| 200 | 28 | 91 | 81 |

What those columns buy is one column per agent kind, plus a complete bipartite join between adjacent columns. `routeEdges` (`agent-graph.ts:161-170`) is:

```ts
for (const source of column) {
  ...
  for (const target of next) {
```

Every agent in a stage is connected to every agent in the next stage. The edge set is a function of stage adjacency alone, and `AgentPhase` (`clients/core-state/src/run-map.ts:26-38`) has no predecessor, parent, or dependency field for it to be a function of anything else. `edgeTone` (`agent-graph.ts:283-296`) reads only the two endpoints' own statuses. The connectors therefore assert an ordering that the left-to-right columns already show, and 15 of those 75 columns are spent drawing it.

The vertical cost has the same shape: `NODE_HEIGHT = 5` plus `ROW_GAP = 2` is 7 rows per stacked agent, so a stage with three attempts is 19 rows.

Separately, the hierarchy #518 built stops one rung short. Its stated hierarchy is `run > hypothesis > round > agents/transcript`, and it shipped the vertical rail for `round`. The `agents` rung is still a separate pane rather than a level of that rail.

### Desired outcome

Opening a round in the rail lists that round's agents beneath it, in place, and closing it returns the rail to rounds only. Each agent row carries what an operator reads while watching a round: kind, status, and elapsed or duration. The Agents pane stops claiming a share of the width, and the transcript takes those columns back.

The graph stays reachable behind an opt-in, so the two can be compared before anything is deleted.

### Acceptance criteria

- [ ] A round in the rail expands to list its agents and collapses again, with the binding shown on the key-help line.
- [ ] Agent rows show kind, status, and elapsed or duration, and truncate rather than wrap at the rail's width.
- [ ] Selecting an agent row filters the transcript to that agent, reusing the existing selection state rather than a parallel one.
- [ ] The rail's scroll window accounts for expanded rounds, so overflow counts stay correct while a round is open.
- [ ] The transcript gains the columns the Agents pane held, and the `TRANSCRIPT_MIN` floor still holds at every width and round shape.
- [ ] The graph stays reachable behind an opt-in; `agent-map.ts` and `agent-graph.ts` are not deleted.
- [ ] Relevant tests or validation pass.

### Scope and non-goals

In scope:

- `round-rail.ts` rendering and windowing, the expansion state and its binding, the round view's width budget in `app.ts`, and the key-help line.

Not in scope:

- Deleting `agent-graph.ts` or `agent-map.ts`. This issue makes the list the default and keeps the graph behind a flag. Removing the graph is a separate decision, taken once the two have been compared.
- The event-model bugs in #514 (see below).
- Adding per-agent dependency data to the event schema.

**This partly reverses #396.** #396 introduced the graph to solve "multiple agents per round". That problem is real and has not gone away: a round can have several agents of one kind. What has not held up is the claim that a graph is how to show it. #396's own acceptance criteria asked for edges as handovers, and for unrelated handovers never to merge into a line that reads as a connection. With no per-agent dependency in the model, the only edge set available is the complete join, which is precisely a line reading as a connection where none was recorded. A list shows several agents of one kind as several rows, which is what the data supports.

**A graph could show topology a list cannot.** That argument is real and is not disputed here. It does not apply yet, because the topology is not modelled. If per-agent dependencies are ever recorded, whether through `invocation_id` parentage or an explicit predecessor field, the calculus changes and this decision should be revisited. That is why the graph is kept behind a flag rather than deleted.

**Relationship to #514.** #514 reports the graph rendering a different, growing topology after a run is killed and reopened. Its rendering symptoms (the `-|-|->` edge hairball, nodes clipping off-pane as attempts accumulate, an unreadable stacked column) are properties of the graph and go away with it. Its root causes do not: `run_interrupted` carrying `agent_kind: null` and `round_label: null` so `applyPhaseEvent` drops it (`run-map.ts:164-165`), retries colliding on one label through `upsertPhase`, seeded-but-unstarted phases never closed, and elapsed timers ticking across a kill are all projection bugs. In a list they resurface as duplicate rows, statuses stuck on `active`, and timers still counting. This change moots #514's presentation half only. #514 should be re-scoped to the projection bugs rather than closed.

**#553** targets `agent-map.ts:255-268` node selection in one of its two halves. That half goes away with the graph; its row-selection half does not.

Rendering rules from `tui-conventions.md` still bind: a box has a rounded border or a background fill and never both, focus is marked by a non-colour channel, and a change that adds a binding names the rule that binding follows.

### Verification approach

- Unit test over the rail's window with a round expanded, across synthetic round sets, asserting overflow counts and that no row is emitted beyond the rows available.
- Rendering test through `createTestRenderer` for expand and collapse. The symptom is terminal geometry (rows, truncation, overlap), and `coding-best-practices.md:215-219` requires the test renderer for those rather than a formatter test.
- Recompute transcript width across terminal widths and round shapes, asserting the `TRANSCRIPT_MIN` floor still holds, extending the existing table in `agent-map.test.ts`.
- `pnpm check:ts && pnpm check:ts-architecture && pnpm check:clients && pnpm test:clients`.
- Drive `clients/tui/dev/mock-ui.sh --tmux 150x40` and capture the frame, with the opt-in on and off.

Parent: #284; continues #518; partly reverses #396; moots the presentation half of #514; affects #553

Nano-AI:
> **This is an experiment.** It is a spike meant to be looked at and judged, not
> a finished change. It is deliberately a draft, deliberately not stacked, and
> parts of it are unfinished (listed at the bottom). Closes nothing.

## Problem

Closes #653 if this direction is taken.

The round view spends most of its width on an agent graph whose edges carry no
information.

`GRAPH_SHARE = 0.55` targets 55% of the terminal for the Agents pane, floored by
the graph's geometry: `GUTTER = 5` between every pair of stages spends 15 columns
on connectors alone in a four-stage round. What those columns buy is a **complete
bipartite join**. `routeEdges` is `for (const source of column) { for (const
target of next) { ... } }`: every agent in a stage connects to every agent in the
next one. `AgentPhase` has no predecessor, parent, or dependency field, and
`edgeTone` reads only the two endpoints' own statuses, so the edge set is a
function of stage adjacency and nothing else. The connectors assert the ordering
the columns already show.

Side by side at 150 columns, on the same fixture, the same round:

```
graph on  (transcript 47 cols)          rail lists agents (transcript 116 cols)
╭─ Rounds ───────────╮╭─ Agents ────────────────────────────╮   ╭─ ▸ Rounds ─────────────────────╮
│ ▸r1 ✓ done 25m 35s ││ Round 1 flow · 25m 35s              │   │ ▸r1 ✓ done 25m 35s             │
│                    ││ ┌──────────┐   ┌──────────┐         │   │   ✓ orchestrator        4m 40s │
│                    ││ │✓ orchest…│──▶│✓ implem… │──▶ ...  │   │   ✓ implementer        13m 57s │
│                    ││ │completed │   │completed │         │   │   ✓ judge     completed 6m 57s │
│                    ││ └──────────┘   └──────────┘         │   │   ○ profiler           pending │
```

The graph spends 75 columns and still truncates `orchestra…` and `implement…`.
The rail shows both in full in 34.

### Before / after

Same fixture, same run, 150x40, dark theme, full window. Base is #645.

**Before** — the Agents pane holds four boxes in a straight line. `orchestrator`
and `implementer` are truncated to `orchestra…` and `implement…` at a width that
also leaves most of the pane empty. The transcript takes what is left: the
build-and-correctness-gate command wraps across four lines, and
`total_ms=562.9504` breaks mid-number into `562.` / `9504`.

before

**After** — the same four agents are five rows in the rail, full names, with
their timings, and `profiler` still reading `pending`. The transcript gets the
reclaimed columns: the same command now fits in two lines and the number is
intact.

after

The connectors visible in the before frame are the ones this PR argues carry no
information: `routeEdges` joins every node in a stage to every node in the next
unconditionally, and `AgentPhase` records no agent-to-agent dependency, so they
restate the left-to-right ordering the columns already show.

## Solution

A round in the rounds rail expands with `Enter` to list its agents beneath it,
and collapses again. This is the `agents` rung of the `run > hypothesis > round >
agents/transcript` hierarchy #518 built, drawn in the rail that already carries
the rest of it, rather than a new invention.

**What it gives up.** A graph can show topology a list cannot. That argument is
real. It does not apply here because the topology is not modelled: with no
per-agent dependency in the event schema, the only edge set available is the
complete join. If per-agent dependencies are ever recorded, the calculus changes.
That is why `agent-map.ts` and `agent-graph.ts` are untouched and the graph stays
reachable behind `VIBESYS_AGENT_GRAPH=1`, following the `VIBESYS_BOOT_TRACE`
precedent for an opt-in read at this layer. Nothing is deleted.

This also **partly reverses #396**, which introduced the graph for "multiple
agents per round". That problem is real and remains; what has not held up is that
a graph is how to show it. Several agents of one kind become several rows.

**#514 is closed, and nothing here re-scopes it.** An earlier revision of this
body argued that this PR moots only #514's presentation half, and that its
projection-layer root causes (`run_interrupted` carrying `agent_kind: null` so
`applyPhaseEvent` drops it, retries colliding on one label through `upsertPhase`,
seeded-but-unstarted phases never closed, timers ticking across a kill) would
resurface here as duplicate rows and statuses stuck on `active`. That is no
longer true: #619 merged and fixed the projection layer as well, closing #514
outright, so there is nothing left for it to be re-scoped to. **#553**'s
node-selection half dies with the graph; its row-selection half does not.

### Architecture

- `expandedRounds: readonly number[]` on `SessionState`, with `toggleRoundAgents`
  next to the existing `toggleTodos`. Selection is untouched: expanding says
  "show me what ran here", not "take me there".
- `railWindow` grew a `rowsFor(round)` callback in place of its `rowHeight`
  scalar, so the scroll window budgets rows rather than rounds and the overflow
  counts stay correct with a round open.
- Agent rows reuse the existing machinery rather than adding a parallel one:
  `phasesForRound` for the data, `selectedAgentKind` / `selectAgent` for
  selection and the transcript filter, and `STATUS_MARKER` / `statusColor`
  exported from `agent-map.ts` so the list and the graph speak one vocabulary.
- The single elapsed-clock timer became a list of ticking rows, so an expanded
  round's active agent ticks through the same mechanism instead of a second one.

**Keybinding: `Enter`, and it needs no new rule.** The hypothesis view already
gives `Enter` the meaning "drill into the selected row" (it opens the selected
round); the rail's agents are the next rung down. It was free at rail focus
because the round view's existing `Enter` handler is guarded to `roundFocus ===
'transcript'`. Like `[`, `]` and the arrows, it is gated on an empty command
input. `tui-conventions.md` gains the rule, as that doc requires of any change
that adds a binding.

**Width reclaimed, measured.** `TRANSCRIPT_MIN` (42) and `MIN_SPLIT_WIDTH` (100)
are unchanged; `GRAPH_SHARE` is now only consulted while the graph is on. The
rail is 34 wide when it lists agents and keeps its old 28 when the graph is on,
so opting into the graph does not also cost it the room it needs to draw. Because
there is no agents pane to budget for, the rail also survives to much narrower
terminals (from 106 columns down to 76).

| Terminal | Transcript before | Transcript now | Gain |
| --- | --- | --- | --- |
| 120 (4 stages) | 62 | 86 | +24 |
| 150 (4 stages) | 47 | 116 | **+69** |
| 200 (4 stages) | 81 | 166 | +85 |
| 240 (4 stages) | 121 | 206 | +85 |

Honest about the narrow end: between 89 and 105 columns the rail is now on screen
where it used to be hidden, so it costs the transcript the columns it takes. The
`TRANSCRIPT_MIN` floor holds at every width, and a test asserts it.

### Correctness properties

- The transcript never drops below `TRANSCRIPT_MIN` at any terminal width or
  round shape, in either mode.
- The rail never emits more rows than it has, at any height, with any number of
  rounds expanded; overflow indicators are drawn only while a row is free.
- Collapsing a round returns the rail to exactly its previous contents.
- An expanded round lists only its own agents, in the order they ran.
- Status is carried by a glyph, never by colour alone.
- `agent-map.ts` and `agent-graph.ts` are behaviourally unchanged; with
  `VIBESYS_AGENT_GRAPH=1` the round view is what it was on the base branch.

### Testing

| Check | Result |
| --- | --- |
| `pnpm check:ts` | clean, 87 files |
| `pnpm check:ts-architecture` | clean, 92 modules / 346 dependencies, no violations |
| `pnpm check:clients` | clean, 0 errors |
| `pnpm test:clients` | **748 pass, 0 fail**, 29 files (737 on the base branch, so +11) |

The 11 new tests are expand/collapse and agent-row rendering through
`createTestRenderer` in `round-rail.test.ts`, plus the reclaimed-columns
measurement in `agent-map.test.ts`, which pins 47 -> 116 at 150 columns against
the base branch's geometry rather than restating the new constants.

The graph's own tests in `app.test.ts` are unchanged and now run with the flag
on, scoped by `beforeAll`/`afterAll` because bun runs a suite in one process.

Driven in the harness at the size the reviewer asked for:

```
clients/tui/dev/mock-ui.sh --tmux 150x40 --speed 0
```

Expand, collapse and re-expand all verified against real frames, plus the same
run with `VIBESYS_AGENT_GRAPH=1` to confirm the graph still draws at 150.

## What a reviewer should look at to decide

1. **Is the bipartite claim right?** `routeEdges` in `agent-graph.ts`, and
   `AgentPhase` in `clients/core-state/src/run-map.ts`. If per-agent dependencies
   are coming, this change is premature and the flag should default the other way.
2. **Is a list enough for a round with a dozen agents?** The fixture has four.
   The rail scrolls, but a twelve-agent round pushes other rounds out of the
   window, and that is the case this spike has not been driven against.
3. **Is `Enter` the right key**, given `Esc` already means "back one level".
4. **The narrow-width trade** at 89 to 105 columns, above.

## Unfinished (it is a spike)

- `Tab` still cycles agent selection and sets `roundFocus: 'agents'`, which with
  the graph off is a pane nobody can see. The arrow-key pane order drops
  `'agents'` correctly, but `Tab` was left alone: fixing it properly means moving
  the focus decision into `session-model.ts`, which cannot read a `ui/` flag
  without inverting the dependency direction.
- Agent rows show the status word only when the row has room for it, so a wide
  kind like `orchestrator` shows glyph plus duration and `judge` shows the word
  too. Legible, but ragged.
- No mouse test for clicking an agent row; the handler is wired but only
  exercised by hand.
- The compact (13 column) rail truncates agent rows hard. Not designed for.
- `agent-map.ts` keeps its stacked fallback, which is now nearly dead code while
  the flag is off.
- `ROUND_KEYS` in `app.ts` is a module-level const, so it reads the flag at
  import time. Correct in production, where the environment precedes the process,
  but it means a test that flips the flag later still sees the old help string.


Nano-AI:
**Correction: #514 is closed, and the caveat above is obsolete.**

This was written while #514 was open. It has since been closed as completed by **#619** (`f3bac91`, "TUI: close out dead rounds and keep the agent graph stable across kill and resume").

The claim here was that #619 would fix only #514's *rendering* half, and that its projection root causes would resurface in a list as duplicate rows and statuses stuck on `active`. That is wrong. #619 fixed the projection layer too: it added `closeOpenRunState` and `closeAbandonedRunState` in `clients/core-state/src/run-map.ts`, handled `run_interrupted` despite its null `agent_kind`/`round_label`, and introduced `interrupted` and abandoned statuses. A list built on that projection inherits the fix rather than re-exposing the bug, so **there is nothing here for #514 to be re-scoped to.**

**And the honest other half:** #619 spent roughly 94 lines improving the graph itself (`agent-graph.ts` +40, `agent-map.ts` +54). This proposal retires a component that was repaired days ago. That cuts both ways and should be weighed rather than buried — it is effort that would be discarded, and it is also effort spent maintaining a view whose edges carry no information.

The case for this change does not rest on #514 and is unaffected: `GRAPH_SHARE = 0.55` of terminal width, `GUTTER = 5` columns of connectors per stage boundary, and edges that are a complete bipartite join over stage adjacency because `AgentPhase` records no dependency between agents.

Nano-AI:
Spinner added. The queued "grey agent rows" half of this item turned out to already work.

`#renderAgent` reuses `agent-map.ts`'s existing `STATUS_MARKER` and `statusColor()`, and pending/default already resolve to `theme.textSubtle`, so non-active rows were muted by code that predates this branch. Nothing was added for that.

What was missing was motion: an active agent wore the same static `●` from the moment it started to the moment it finished, indistinguishable from a row that had stalled. `agentMarker(status, frame)` in `round-rail.ts` now swaps in a frame from `activity-bar.ts`'s `SPINNER_FRAMES` when status is `active`; every other status keeps its static glyph.

**One tradeoff worth a reviewer's attention.** The rail's existing `#elapsedTimer` ran at a hardcoded `1000ms`. It now runs at `SPINNER_INTERVAL_MS` (120ms, the constant every other in-flight indicator in the app already animates on), so a single timer drives both the elapsed-time text and the spinner rather than a second interval being added. The cost is that the elapsed text re-renders 8x more often, and only while `#ticking` is non-empty: the timer does not start when no agent is active and is cleared once they finish.

The animation stays local to `round-rail.ts` rather than moving into `STATUS_MARKER`. `agent-map.ts`'s graph nodes and its stacked fallback render the marker once and never re-render on a tick, so animating the shared constant would freeze each node on whatever frame was current at last render instead of spinning.

`pnpm test:clients`: 936 pass, 0 fail. The before/after pair above predates the spinner and is being reshot.

Nano-AI:
Correction to the note above: the before/after pair does not need re-shooting, and it does not show the spinner.

The replay fixture runs to completion, so every agent in that frame is `✓ completed` or `○ pending` and none is `active`. The spinner only replaces the `active` glyph, so it changes nothing in that still. The frame does incidentally confirm the grey-rows half: `○ profiler pending` renders in `theme.textSubtle` already.

A still cannot show animation in any case. The motion is asserted in `round-rail.test.ts` instead: successive frames differ, the sequence wraps at `SPINNER_FRAMES.length`, every non-active status is frame-invariant, and every frame is exactly one display cell wide.

Nano-AI:
Force-pushed `72b3de3` -> `7a7a7fc`: a rebase, after #644 gained a commit and #645 was rebased onto it. `git range-diff 840c82a..72b3de3 70acaee..7a7a7fc` reports both commits unchanged (`=`), so neither the rail listing nor the spinner has moved.

Nano-AI:
Superseded by #686.

Review kept the agent graph (operators see which agents their workflow runs, and it is the shape multi-agent fan-out needs), so #686 removes the round rail instead: rounds become a one-row tab bar across the top, and the rail's column goes to the graph (40%) and the transcript.

Converting this to a draft to keep the discussion; it can be closed whenever convenient.

Issue #613, opened 2026-09-04
TUI: Index core-state phases so live event folding does not grow with run length


AyanBinRafaih:
### Workstream

Core/runtime

### Problem

The run-map reducer repeatedly scans and copies the complete accumulated round and phase arrays for each event. `applyRoundEvent()` searches `rounds`, then `replaceRound()` searches and maps the array again. `applyPhaseEvent()` repeatedly seeds roles through `ensurePhase()`, which scans the complete phase list for each expected role. `upsertPhase()` performs multiple `findIndex` passes and then maps the complete list to replace one item. Timing updates also search the accumulated phase list.

The number of phases grows with rounds. Event folding therefore approaches events multiplied by phases, and every live event late in a run performs work proportional to the entire run history.

[PR #468](https://github.com/uw-syfi/vibesys/pull/468) measured this residual at approximately 54 seconds for a full replay of roughly 72,000 events. Its analysis notes that a live event at round 280 copies an approximately 1,120-entry phase array. Tail bootstrap hides much of this during startup, but the cost remains on live delivery late in long runs, scroll-back prefix folding, old-server full-replay fallback, tests, and replay tooling.

### Desired outcome

Run-map folding should update rounds and phases through stable indexes with near-constant lookup cost per event. Published `CoreState` must retain its current immutable external contract and deterministic ordering. Full replay, batched replay, tail bootstrap, and prefix backfill must produce equivalent state.

### Acceptance criteria

- [ ] Round lookup does not linearly scan the full rounds array for every event.
- [ ] Phase lookup uses a stable key covering round, role, and execution identity.
- [ ] Expected-role seeding does not scan the full phase history once per role.
- [ ] Updating one phase does not map the complete phase history solely to replace one element.
- [ ] Public `rounds` and `phases` arrays retain deterministic current ordering.
- [ ] Full replay, single-event folding, batched folding, rebootstrap, and prefix backfill remain equivalent.
- [ ] Benchmarks cover at least 15,000, 36,000, and 72,000 events and report per-event fold time.
- [ ] Live per-event fold cost at a late round does not grow linearly with earlier phases.
- [ ] Relevant client tests and architecture checks pass.

### Scope and non-goals

In scope:

- internal round and phase indexing;
- immutable publication strategy;
- replay and live-fold benchmarks.

Not in scope:

- changing round or phase semantics;
- changing visual layout;
- reducing retained history;
- changing the backend protocol.

### Constraints or design considerations

[...]

Issue #611, opened 2026-09-04
TUI: Preserve agent timing when history backfill splits an execution


AyanBinRafaih:
### Observed behavior

Tail bootstrap followed by history backfill does not reproduce full-replay timing when an agent execution starts in an older chunk and finishes in a newer chunk.

The newer suffix is folded first. Its finish event has no matching start, so `finishAgentTiming()` discards the finish timestamp. When the older prefix is merged later, it contributes the start, but the discarded finish is no longer available. The round is left with an active start and no completed interval.

This behavior is explicitly normalized out of the core-state equivalence test rather than fixed.

### Expected behavior

Folding a tail and then backfilling every older chunk should produce the same completed and active timing intervals as folding the full ordered log once, regardless of chunk boundaries. The fold should retain enough information about unmatched starts and finishes to reconcile them during prefix merge.

### Reproduction

Use this ordered event log:

1. `agent_execution_started`, execution `exec-a`, round 1.
2. `agent_execution_finished`, execution `exec-a`, round 1.
3. `round_finished`, round 1.

Bootstrap from the suffix beginning at event 2, then backfill event 1. Full replay contains one completed interval from event 1 to event 2, while tail plus backfill contains no completed interval and retains an unmatched active start.

### Affected subsystem

Optimization loop and agents

### Environment

Commit: `60c6763540943fc9998ad5fc88de818fe6d960ba`  
OS: Linux (debian)
Client: `@vibesys/core-state`  
History mode: tail bootstrap plus prefix backfill  
Frontend: rounds rail and agent elapsed-time presentation

### Impact

Blocks normal use

### Relevant logs

```shell
The regression fixture should show one completed `exec-a` interval in full replay and no completed interval, plus an active `exec-a` start, after backfill.
```

### Related issues or pull requests

#284, #468

### Pre-submission checks

- [x] I searched for an existing issue describing this problem.
- [x] I removed sensitive information from this report.

PR #657, opened 2026-09-10
TUI: project and render structured execution status

### LINKED ISSUE #614: TUI: Fold the agent status already carried on output and tool-call events
### Workstream

Core/runtime

### Problem

The backend calculates structured agent status for streamed output and tool calls: progress label, agent label, elapsed seconds, input-token count, and context-window size. `AgentLogger` computes that status for streamed content, and `AgentOutputChunkData` and `ToolCallData` include it in the generated protocol. Core state ignores it.

As a result, the TUI usage meter advances only on separate `usage_update` events, per-agent elapsed status is discarded, agent progress and label are unavailable to the renderer, and useful context-window pressure data is transmitted repeatedly and then thrown away.

### Desired outcome

Core state should retain the freshest structured status per active execution, ordered by event sequence. The TUI should show live per-agent progress and context pressure without parsing formatted text or waiting for a later usage event.

### Acceptance criteria

- [ ] `agent_output_chunk.status` and `tool_call.status` update a typed per-execution status projection.
- [ ] Status is keyed by execution identity and cannot overwrite another concurrent agent’s state.
- [ ] Older replayed status cannot replace a newer checkpoint or event.
- [ ] The existing global usage meter uses the freshest applicable status without regressing multi-agent semantics.
- [ ] At least one TUI surface shows agent label, elapsed time, and tokens versus context for the selected or active agent.
- [ ] Context pressure remains legible without color alone.
- [ ] Missing status in legacy events preserves existing behavior.
- [ ] Tail replay, reconnect, and prefix backfill produce deterministic status.
- [ ] Relevant Python event, generated protocol, core-state, and TUI tests pass.

### Scope and non-goals

In scope:

- folding the already-emitted payload;
- per-execution status state;
- one concise presentation surface.

Not in scope:

- introducing another status protocol;
- embedding formatted status strings in backend events;
- redesigning the entire header or agent graph.

Do not collapse concurrent agents into one last-write-wins global value without retaining execution identity. Backend elapsed seconds are event snapshots. The UI may derive additional wall-clock advancement for an active execution, but replayed historical events must remain deterministic.

### Verification approach

Add fixtures with interleaved status from two executions, reconnect replay, an older backfill chunk, and a terminal execution checkpoint. Assert that the newest status for each execution wins and finishing one execution does not clear the other. Validate with a real provider run and confirm that context usage advances during streaming.

#284, #517

AyanBinRafaih:
## Problem

The backend already sends structured progress, agent label, elapsed time, input tokens, and context-window size on `agent_output_chunk.status` and `tool_call.status`. Core state discarded that payload, so the activity bar could only render a generic `Working` label and a client-side timer. The global token meter also waited for a later `usage_update` event.

## Solution

Export the generated status payload through the backend-client protocol facade, then fold it into a typed `executionStatuses` map keyed by execution identity. Each record carries its source sequence and timestamp. Later partial payloads retain fields from earlier payloads for the same execution, while a positive-sequence record cannot be replaced by an older or unsequenced event. Repeated legacy events with no sequence still advance in fold order.

Prefix backfill first filters each side against the active execution generation, then merges each execution field by field. This prevents a reused id from inheriting an earlier generation's label or context. The global usage projection comes from the accepted per-execution record. Private `WeakMap` provenance retains that status snapshot and its known generation even after the execution or run finishes, so an older prefix can restore a context window omitted by a newer token-only suffix without exposing replay metadata in `CoreState`. Concurrent agents never donate context-window fields to each other.

The activity bar uses the status belonging to the active execution generation. It renders the backend agent label, progress, elapsed snapshot, and tokens versus context. Context pressure includes `high` or `critical` text at 80% and 95%, so color is not required. Events without structured status retain the existing role label, `Working`, and live elapsed timer.

### Architecture

```mermaid
flowchart LR
  E[output chunk or tool call] --> P[backend-client generated status type]
  P --> F[core-state execution status fold]
  F --> X[status by execution id]
  X --> M[fieldwise prefix merge]
  X --> U[global usage from accepted execution]
  M --> U
  X --> G[generation-safe active selector]
  G --> A[TUI activity bar]
  U --> H[existing header usage meter]
```

Pending statuses are retained while a tail batch waits for its authoritative execution checkpoint. `executionStatusFor` rejects a record older than the checkpoint's `startedAt`, so a reused execution id cannot surface stale status. A subsequent active fold prunes the stale record. Execution-finished and run-terminal events remove their status immediately.

## Verification

The regression fixture `/tmp/vibesys-614-before.test.ts` fails against base `ea6f9073`: the status payload folds to `undefined` (`0 pass, 1 fail`).

### Correctness properties

- Status is isolated by execution identity and ordered by sequence.
- Partial updates inherit fields only from the same execution.
- Reused execution ids never inherit fields from a prior generation.
- A positive-sequence status is not replaced by an unsequenced replay event.
- Prefix backfill produces the same per-execution status and global usage as a full replay, including after execution or run termination.
- Checkpoint generation timestamps prevent a reused id from exposing stale status.
- Finishing one execution preserves concurrent executions; run termination clears all pending status.
- Missing legacy status preserves the prior activity-bar behavior and does not create usage.
- Context pressure is conveyed by text as well as a percentage.
- The #611 timing and #613 run-map changes merge without conflict; their combined core suite passes.

### Testing

- `pnpm --filter @vibesys/backend-client check`
- `pnpm --filter @vibesys/core-state check`
- `pnpm --filter @vibesys/core-state test` (`111 pass, 0 fail`)
- `pnpm --filter @vibesys/tui check`
- `bun test --max-concurrency 1 src/ui/activity-bar.test.ts` (`6 pass, 0 fail`)
- Combined #611 + #613 + #614: direct core TypeScript check and core tests (`123 pass, 0 fail`)

- Full Python CI suite: 5,741 passed, 6 environment or platform skips; global and module coverage gates passed.
- Final Node 24.21.0 / Bun 1.3.9 client checks: 25 backend-client, 111 core-state and 741 TUI tests passed. Protocol generation was stable; formatting, lint, type, architecture, launcher and build checks passed with the required Unix socket access.

- Real completed-journal replay across nine chunk decompositions preserves exact round timing, terminal execution status, usage (including model identity), and all 133 transcript entries. Added regressions preserve explicit nullable model resets and keep private usage provenance independent across immutable state forks. Pre-existing phase-order/identity differences match unchanged main and are outside this change.

Closes #614

### DIFF
--- a/clients/backend-client/src/index.ts
+++ b/clients/backend-client/src/index.ts
@@ -13,6 +13,7 @@ export {
   type StreamTransport,
 } from './persistent-event-stream.js';
 export type {
+  AgentStatusData,
   ChatModelOption,
   ChatOptions,
   ChatProviderOptions,
--- a/clients/backend-client/src/protocol.ts
+++ b/clients/backend-client/src/protocol.ts
@@ -3,6 +3,10 @@ import type {ProtocolDocument} from './generated/protocol.generated.js';
 export type ProtocolRequest = ProtocolDocument['request'];
 export type ProtocolResponse = ProtocolDocument['response'];
 export type RunEvent = ProtocolDocument['event'];
+/** Structured status carried by streamed agent output and tool calls. */
+export type AgentStatusData = NonNullable<
+  Extract<NonNullable<RunEvent['data']>, {channel: unknown}>['status']
+>;
 export type RunSnapshot = ProtocolDocument['snapshot'];
 /** Run lifecycle statuses the backend reports. Source: `RunStatus` in `src/server/run_lifecycle.py`. */
 export type RunStatus = RunSnapshot['status'];
--- a/clients/core-state/src/core-state-prefix.test.ts
+++ b/clients/core-state/src/core-state-prefix.test.ts
@@ -101,6 +101,239 @@ describe('prefix backfill equivalence', () => {
 });
 
 describe('prefix merges across the chunk boundary', () => {
+  it('keeps the newest per-execution status across tail replay and prefix backfill', () => {
+    const events = [
+      executionStartedEvent(1, 'active'),
+      executionStatusEvent(2, 'active', 'agent_output_chunk', {
+        progress: 'Inspecting',
+        agent_label: 'Implementer',
+        elapsed_seconds: 2,
+        input_tokens: 4_000,
+        context_window: 200_000,
+      }),
+      executionStatusEvent(3, 'active', 'tool_call', {input_tokens: 9_000}),
+    ];
+    const checkpoint = [
+      {
+        execution_id: 'active',
+        agent_kind: 'implementer',
+        round_label: 'round-1-implementer',
+        stage: 'implementation',
+        attempt: 1,
+        assignment: 'Implement',
+        started_at: timestamp(1),
+        activity: {
+          kind: 'agent_execution_activity_changed' as const,
+          mode: 'thinking' as const,
+          summary: 'Working',
+          tool: null,
+        },
+      },
+    ];
+    const full = reduceEventBatch(initialCoreState(), events, checkpoint, 3);
+    const tail = reduceEventBatch(initialCoreState(), events.slice(2), checkpoint, 3, 2);
+    const merged = reduceEventPrefix(tail, events.slice(0, 2), 0);
+
+    expect(merged.executionStatuses).toEqual(full.executionStatuses);
+    expect(merged.executionStatuses['active']).toMatchObject({
+      sequence: 3,
+      progress: 'Inspecting',
+      agentLabel: 'Implementer',
+      inputTokens: 9_000,
+      contextWindow: 200_000,
+    });
+    expect(merged.usage).toEqual(full.usage);
+    expect(merged.usage).toEqual({inputTokens: 9_000, contextWindow: 200_000, model: null});
+  });
+
+  it('does not fill a restarted execution from an older generation with the same id', () => {
+    const older = [
+      executionStartedEvent(1, 'reused'),
+      executionStatusEvent(2, 'reused', 'agent_output_chunk', {
+        progress: 'Old work',
+        agent_label: 'Old implementer',
+        context_window: 200_000,
+      }),
+    ];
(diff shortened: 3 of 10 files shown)


Nano-AI:
> [!NOTE]

**Summary:** Folds `agent_output_chunk.status` and `tool_call.status` into a per-execution `executionStatuses` map in core-state, derives global usage from the accepted record, and renders label, progress, elapsed snapshot, and context usage in the activity bar.

**Findings**

1. **[P3] Activity-bar context meter diverges from the header rule** (`clients/tui/src/ui/activity-bar.ts:141-155`)
   `statusSuffix` renders a ratio whenever `contextWindow > 0`. `header.ts:169` renders only the count when `inputTokens > contextWindow`, because the window is a static per-model lookup that can be stale (comment above `usageText`). The header usage now comes from the same status record, so one reading renders two ways. With `inputTokens: 1_200_000, contextWindow: 1_000_000` the activity bar shows `1.2M/1.0M context critical 120%` and the header `1.2M tokens`. `formatTokenCount` (`activity-bar.ts:151-155`) is also a verbatim copy of `header.ts:177-181`.
   Suggested direction: share one token/context formatter between the two views so the over-window rule applies to both.

2. **[P3] `AgentStatusData` re-derived instead of re-exported** (`clients/backend-client/src/protocol.ts:6-9`)
   `generated/protocol.generated.ts` already exports `interface AgentStatusData`. The `Extract<..., {channel: unknown}>` resolves to it only because `AgentOutputChunkData` is currently the sole event payload with a `channel` field; a `channel` field on any other payload would change or break the alias. `export type {AgentStatusData} from './generated/protocol.generated.js';` yields the same type without the incidental discriminator.

3. **[P3] Unrelated `agentshim` pin conflicts with main** (`pyproject.toml:20`)
   Commit `e0bd143` pins `agentshim>=0.5.0,<0.6.0`, which main has superseded with `>=0.6.1,<0.7`; it is the only conflict in `git merge-tree --write-tree origin/main pr/657`. Consider dropping the commit when rebasing.

PR #659, opened 2026-09-10
TUI: stream journal attachment with a validated index cache

### LINKED ISSUE #612: TUI: Remove full-file allocation from event-journal attach
### Workstream

Platform/backends

### Problem

Tail bootstrap bounds how much history is validated and sent to the TUI, but attaching the event journal still reads the complete `run-events.jsonl` file into memory and splits it into a second set of line-sized byte objects.

Evidence:

- [`_scan_unlocked`](https://github.com/uw-syfi/vibesys/blob/60c6763540943fc9998ad5fc88de818fe6d960ba/src/server/events.py#L641-L684) calls `read_bytes()` followed by `splitlines(keepends=True)`.
- The full raw byte string and split line objects coexist during scanning.
- Every record is parsed with `json.loads` to recover four header fields, even though only the recent tail and sparse run-level spine are validated as full events.
- [PR #468](https://github.com/uw-syfi/vibesys/pull/468) measured attach at 431 ms for a 99 MB, roughly 73,000-event journal and identified attach as the remaining linear startup slope.

This makes TUI startup and resume proportional to total historical log bytes even when the frontend initially consumes only a bounded tail. Peak memory is also proportional to the full log.

### Desired outcome

Attaching a large run should not allocate the entire journal plus a split copy. Startup work should scale with the bounded tail plus a compact index update. Existing logs must remain readable without migration, and strict corruption handling and legacy sequence repair must remain intact.

### Acceptance criteria

- [ ] `EventStore` attach does not call `Path.read_bytes()` for the complete journal.
- [ ] Peak additional memory during attach is bounded independently of total journal size, apart from the compact index itself.
- [ ] Warm attach can reuse a validated sidecar index or incrementally scan only bytes appended since the last valid index watermark.
- [ ] A missing, stale, or corrupt index is detected and rebuilt safely from the JSONL source of truth.
- [ ] The JSONL source remains authoritative and tail reads continue validating requested records and required spine records.
- [ ] Malformed-tail recovery, complete-record validation, legacy cursor repair, and append semantics remain covered.
- [ ] Benchmarks report wall time and peak RSS for at least 20 MB, 100 MB, and 500 MB journals.
- [ ] Relevant server, tail-subscription, protocol, core-state, and TUI startup checks pass.

### Scope and non-goals

In scope:

- journal indexing and attach;
- incremental index validation;
- memory and startup benchmarks;
- compatibility with existing JSONL logs.

Not in scope:

- removing the JSONL audit log;
- weakening schema validation;
- changing the TUI tail size;
- changing event semantics.

The JSONL file remains authoritative. Any sidecar or memory-mapped structure is a disposable cache and must be validated against file identity, size, and a safe content boundary. Index replacement should be atomic. An interrupted index write must never invalidate a healthy journal. Append and reopen behavior must remain correct when the last JSONL record is incomplete.

### Verification approach

Use the long-history benchmark shape from PR #468 and record attach wall time, peak RSS, number of fully validated events, bytes read, cold attach, warm attach, and attach after appending a small suffix. Run the event-store, tail-subscription, runtime, protocol-generation, core-state, and TUI startup suites.

#284, #329, #468, #470

AyanBinRafaih:
## Problem

Fixes #612.

`EventStore` attached a durable run by calling `read_bytes()` and then `splitlines(keepends=True)`. The complete journal and a second set of line byte objects coexisted during startup, so attach memory grew with the full JSONL payload. Every reopen also parsed every record header even when the source had not changed.

## Solution

Attach now streams the JSONL source one line at a time. The scanner retains only the in-memory record index, two current lines, and the bounded eager-validation tail. If the cheap header scan cannot classify a complete record, that record is fully validated in place, which preserves strict corruption errors without retaining every parsed payload.

A versioned `run-events.jsonl.idx` sidecar caches record offsets and the four header fields used during bootstrap. The journal is append-only, so reuse requires the same device and inode, a source at least as large as when it was indexed, and matching SHA-256 fingerprints of the indexed head (first 64 KiB), seven interior 4 KiB samples, and the 64 KiB window ending at the last safe newline boundary. A source of exactly the indexed size must also have identical modification and change times. `EventJournal.attach` appends `SERVER_STARTED` after building the store, so an exact size/mtime check would reject the sidecar on every real reattach; a grown source instead has only its suffix scanned and the extended index is republished atomically. The sidecar also has a checksum and validates its count, contiguous offsets, field types, and legacy sequence-repair recurrence before use. A shrunken or replaced source, any fingerprint mismatch, or a sidecar of another format version (now 2) performs a full streaming rebuild. The fingerprints do not cover interior bytes outside the sampled windows, which is the accepted residual for an append-only file. The JSONL file remains authoritative.

Sidecars are written to a mode-0600 temporary in the journal directory, flushed, fsynced, and published with `os.replace`. Sidecar read, write, directory-sync, and cleanup failures are cache misses and do not affect a healthy journal.

### Architecture

```mermaid
flowchart TD
    A[EventStore attach] --> B[Capture JSONL identity]
    B --> C{Indexed prefix unchanged?}
    C -->|yes| D[Stream and validate sidecar records]
    D --> S[Scan only the suffix past the boundary]
    S --> F
    S --> I
    C -->|no| E[Stream JSONL records]
    F[Validate bounded event tail]
    E --> G[Validate uncertain records in place]
    G --> F
    F --> H[Expose in-memory record index]
    E --> I[Write temporary sidecar]
    I --> J[Recheck JSONL identity]
    J --> K[Atomic sidecar replacement]
```

`server.events` continues to own JSONL parsing, corruption policy, lazy event validation, and legacy cursor repair. The new leaf `server.event_index` module owns only primitive sidecar serialization, validation, source identity, and atomic cache I/O. No protocol, journal-spine, core-state, or TUI behavior changes.

## Verification

| Before | After |
| --- | --- |
| Before | After |

The regression test fails on the merge base by replacing `Path.read_bytes` with an assertion during cold attach. New tests also cover exact warm reuse, suffix-only scanning and index extension after append (asserting scan counts, including that a third attach scans nothing new), the real `EventJournal.attach` path reusing the index despite `SERVER_STARTED`, rejection of head, boundary, and interior edits combined with growth, replaced files, truncation, partial sidecars and other sidecar versions, a same-size overwrite with current stat fields copied into the stale cache, changes during scan and cache load, corrupt and truncated sidecars, invalid event types and non-ASCII checksums, valid unterminated and malformed tails, legacy cursor repair, unreadable sidecars, failed atomic replacement, and cleanup failure.

The checked-in benchmark runs every attach in a fresh subprocess and reports median and p95 wall time and peak RSS for cold, warm, and changed-source cases. The changed-source mode appends one event to a warm journal; since the follow-up fix it measures the suffix-scan path, not a full rebuild, and the table below predates that change.

### Correctness properties

- Attach never reads or splits the complete JSONL file into memory.
- A sidecar is accepted only after structural validation, checksum validation, same-file and no-shrink checks, bounded content fingerprint matching of the indexed prefix, and a second source-identity check after the sidecar read. Any doubt is a full rebuild, never a stale index.
- A source change during a cold scan prevents that scan from publishing an index.
- Only newline-terminated records enter the persistent safe boundary. Valid unterminated records and malformed append tails are rescanned, so append repair behavior remains unchanged.
- Complete invalid records still raise, while only an incomplete final append is ignored and truncated by the next append.
- Duplicate, out-of-order, and reset legacy sequences retain the same strictly increasing cursors across cold attach, warm attach, append, and reopen.
- Failure to read, create, sync, replace, or clean up a sidecar cannot hide or damage the authoritative JSONL journal.

### Testing

- Maintainer follow-up (`d67c2b2e`): `tests/server` passes (382 tests) and Ruff check/format are clean locally; CI on the pushed head is the authority for the rest. The benchmark table and full-suite numbers below were measured before this follow-up.

- Regression reproduced against `ea6f9073`: cold attach fails at `Path.read_bytes`; the branch passes the same invariant.
- `.venv/bin/python -m pytest -n 8 --dist loadgroup --cov-report=xml --cov-report=json --cov-fail-under=75 -q`, 5,738 passed and 28 platform/toolchain skips.
- `.venv/bin/python scripts/check_coverage_floor.py coverage.json`, passed. `.venv/bin/diff-cover coverage.xml --compare-branch=upstream/main --include-untracked --fail-under=80`, passed with 90% patch coverage, including the new module.
- Ruff lint and formatting, `ty check`, `lint-imports`, and Markdown link validation passed.
- Protocol regeneration produced zero diff. `pnpm check:ts`, `pnpm check:clients`, `pnpm check:ts-architecture`, `pnpm test:clients`, and `pnpm build:clients` passed using Node 24.21.0 and Bun 1.3.9. Client tests: 25 backend-client, 98 core-state, 738 TUI, zero failures.

For each module root (merge-base main and this branch), ran `.venv/bin/python scripts/benchmark_event_store_attach.py --module-root ROOT --python /tmp/vibesys-issues-20260910/612/.venv/bin/python --sizes-mib 20 100 500 --repeats 3 --work-dir UNIQUE --output REPORT.md`. Each sample uses a fresh Python 3.12 subprocess. With three repetitions, p95 is the largest observed sample; these are local measurements, not a statistical latency guarantee.

| Journal | Mode | Main median / p95 | Branch median / p95 | Main peak RSS median / p95 | Branch peak RSS median / p95 |
| --- | --- | ---: | ---: | ---: | ---: |
| 20 MiB | cold | 0.225 / 0.243s | 0.328 / 0.343s | 80.6 / 80.6 MiB | 42.7 / 42.7 MiB |
| 20 MiB | warm | 0.230 / 0.233s | 0.177 / 0.181s | 80.6 / 80.6 MiB | 43.1 / 43.2 MiB |
| 20 MiB | changed source | 0.226 / 0.231s | 0.337 / 0.345s | 80.6 / 80.6 MiB | 42.7 / 42.8 MiB |
| 100 MiB | cold | 1.086 / 1.086s | 1.521 / 1.533s | 264.5 / 264.5 MiB | 62.1 / 62.2 MiB |
| 100 MiB | warm | 1.076 / 1.082s | 0.814 / 0.834s | 264.5 / 264.5 MiB | 64.6 / 64.7 MiB |
| 100 MiB | changed source | 1.065 / 1.069s | 1.562 / 1.586s | 264.6 / 264.7 MiB | 62.4 / 62.4 MiB |
| 500 MiB | cold | 5.657 / 10.604s | 7.300 / 7.728s | 1184.8 / 1185.3 MiB | 162.6 / 162.7 MiB |
| 500 MiB | warm | 5.352 / 5.391s | 4.408 / 4.409s | 1184.7 / 1184.8 MiB | 178.3 / 178.3 MiB |
| 500 MiB | changed source | 5.339 / 5.429s | 6.158 / 7.648s | 1184.8 / 1184.9 MiB | 162.6 / 162.6 MiB |

At 500 MiB, cold peak RSS falls 86% and warm attach is 18% faster. Cold and changed-source attach pay extra serialization, checksum, and atomic-publication cost. Index files are 0.9, 4.6, and 24.1 MiB at the three sizes. Index memory still grows with record count; this change removes whole-journal payload allocation, not the record index itself.

Closes #612

### DIFF
--- a/scripts/benchmark_event_store_attach.py
+++ b/scripts/benchmark_event_store_attach.py
@@ -1,0 +1,223 @@
+#!/usr/bin/env python3
+"""Measure event-journal attach wall time and peak RSS in fresh processes."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import math
+import os
+import resource
+import statistics
+import subprocess
+import sys
+import tempfile
+import time
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+_MIB = 1024 * 1024
+_DEFAULT_SIZES_MIB = (20, 100, 500)
+
+
+@dataclass(frozen=True, slots=True)
+class Measurement:
+    """One isolated attach measurement."""
+
+    wall_seconds: float
+    peak_rss_mib: float
+    records: int
+    parsed_records: int
+    source_bytes: int
+    index_bytes: int
+
+
+@dataclass(frozen=True, slots=True)
+class BenchmarkTarget:
+    """Stable inputs shared by every mode for one journal size."""
+
+    repeats: int
+    python: Path
+    module_root: Path
+    source: Path
+
+
+def _event_line(sequence: int, payload_bytes: int = 900) -> bytes:
+    event = {
+        "protocol_version": 1,
+        "sequence": sequence,
+        "run_id": "benchmark",
+        "timestamp": "2026-01-01T00:00:00Z",
+        "type": "output",
+        "text": "x" * payload_bytes,
+    }
+    return (json.dumps(event, separators=(",", ":")) + "\n").encode()
+
+
+def _generate_journal(path: Path, target_bytes: int) -> int:
+    path.parent.mkdir(parents=True, exist_ok=True)
+    count = 0
+    written = 0
+    with path.open("wb") as stream:
+        while written < target_bytes:
+            count += 1
+            line = _event_line(count)
+            stream.write(line)
+            written += len(line)
+    return count
+
+
+def _worker(path: Path) -> None:
+    from server.events import EventStore  # noqa: PLC0415  # implementation selected by PYTHONPATH
+
+    started = time.perf_counter()
+    store = EventStore(path, run_id="benchmark")
+    wall_seconds = time.perf_counter() - started
+    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+    peak_rss_mib = rss / (_MIB if sys.platform == "darwin" else 1024)
+    index_path = path.with_name(f"{path.name}.idx")
+    print(  # benchmark result protocol
+        json.dumps(
+            {
+                "wall_seconds": wall_seconds,
+                "peak_rss_mib": peak_rss_mib,
+                "records": len(store.event_headers()),
+                "parsed_records": store.parsed_record_count,
+                "source_bytes": path.stat().st_size,
+                "index_bytes": index_path.stat().st_size if index_path.exists() else 0,
+            }
+        )
+    )
+
+
+def _measure(python: Path, module_root: Path, source: Path) -> Measurement:
+    environment = dict(os.environ)
+    source_path = str(module_root / "src")
+    environment["PYTHONPATH"] = os.pathsep.join(
+        [source_path, environment["PYTHONPATH"]] if environment.get("PYTHONPATH") else [source_path]
+    )
+    result = subprocess.run(  # noqa: S603  # explicit benchmark interpreter and script
+        [str(python), str(Path(__file__).resolve()), "--worker", str(source)],
+        check=True,
+        capture_output=True,
+        text=True,
+        env=environment,
+    )
+    return Measurement(**json.loads(result.stdout))
+
+
+def _prepare_warm(python: Path, module_root: Path, source: Path) -> None:
+    _measure(python, module_root, source)
+
+
+def _percentile(values: Sequence[float], percentile: float) -> float:
+    ordered = sorted(values)
+    return ordered[max(0, math.ceil(percentile * len(ordered)) - 1)]
+
+
+def _summary(measurements: list[Measurement]) -> dict[str, float | int]:
+    walls = [item.wall_seconds for item in measurements]
+    rss = [item.peak_rss_mib for item in measurements]
+    sample = measurements[-1]
+    return {
+        "wall_median_seconds": statistics.median(walls),
+        "wall_p95_seconds": _percentile(walls, 0.95),
+        "peak_rss_median_mib": statistics.median(rss),
+        "peak_rss_p95_mib": _percentile(rss, 0.95),
+        "records": sample.records,
+        "parsed_records": sample.parsed_records,
+        "source_bytes": sample.source_bytes,
+        "index_bytes": sample.index_bytes,
+    }
+
+
+def _run_mode(
+    mode: str,
+    *,
+    target: BenchmarkTarget,
+    next_sequence: int,
+) -> tuple[list[Measurement], int]:
+    index_path = target.source.with_name(f"{target.source.name}.idx")
+    measurements: list[Measurement] = []
+    if mode == "warm":
+        index_path.unlink(missing_ok=True)
+        _prepare_warm(target.python, target.module_root, target.source)
+    for _ in range(target.repeats):
+        if mode == "cold":
+            index_path.unlink(missing_ok=True)
+        elif mode == "changed-source":
+            index_path.unlink(missing_ok=True)
+            _prepare_warm(target.python, target.module_root, target.source)
+            with target.source.open("ab") as stream:
+                stream.write(_event_line(next_sequence))
+            next_sequence += 1
+        measurements.append(_measure(target.python, target.module_root, target.source))
+    return measurements, next_sequence
+
+
+def _render_markdown(results: dict[str, dict[str, dict[str, float | int]]]) -> str:
+    lines = [
+        "| Journal | Mode | Wall median | Wall p95 | Peak RSS median | Peak RSS p95 | Index |",
+        "| --- | --- | ---: | ---: | ---: | ---: | ---: |",
+    ]
+    for size, modes in results.items():
+        for mode, values in modes.items():
+            lines.append(
+                f"| {size} MiB | {mode} | {values['wall_median_seconds']:.3f}s | "
+                f"{values['wall_p95_seconds']:.3f}s | {values['peak_rss_median_mib']:.1f} MiB | "
+                f"{values['peak_rss_p95_mib']:.1f} MiB | "
+                f"{values['index_bytes'] / _MIB:.1f} MiB |"
+            )
+    return "\n".join(lines) + "\n"
+
+
+def main() -> None:
+    """Generate journals and benchmark isolated cold, warm, and changed attaches."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--worker", type=Path)
+    parser.add_argument("--python", type=Path, default=Path(sys.executable))
+    parser.add_argument("--module-root", type=Path, default=Path.cwd())
+    default_work_dir = Path(tempfile.gettempdir()) / "vibesys-event-attach-bench"
+    parser.add_argument("--work-dir", type=Path, default=default_work_dir)
+    parser.add_argument("--sizes-mib", type=int, nargs="+", default=_DEFAULT_SIZES_MIB)
+    parser.add_argument("--repeats", type=int, default=3)
+    parser.add_argument("--output", type=Path)
+    args = parser.parse_args()
+    if args.worker is not None:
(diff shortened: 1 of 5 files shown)


Nano-AI:
> [!NOTE]

**Summary:** Replaces the whole-file `read_bytes()` attach scan with a line-streaming scan, validates header-scan misses in place instead of falling back to a full eager parse, and adds a `run-events.jsonl.idx` sidecar that restores the record index on an exact source match.

**Findings**

1. **[P2] The sidecar is never reused on the production attach path** (`src/server/journal.py:110-122`, `src/server/event_index.py:92-93`)
   `EventJournal.attach` is the only `EventStore` constructor in `src/`. After constructing the store it appends `SERVER_STARTED` (or replays pending events), which changes the source size and mtime. The next process's `load_event_index` then fails the exact identity check, rebuilds from JSONL and rewrites the sidecar. As a result, every real attach pays the PR's "cold" cost (0.328 vs 0.225 s at 20 MiB, 1.521 vs 1.086 s at 100 MiB, 7.300 vs 5.657 s median at 500 MiB, compared with main) plus a sidecar rewrite (24 MiB at 500 MiB). The "warm" row is only reachable through a bare `EventStore` with no append, which is what the benchmark (`scripts/benchmark_event_store_attach.py:78`) and the new tests measure. #612 lists "incrementally scan only bytes appended since the last valid index watermark" and "attach after appending a small suffix" as acceptance and verification items.
   Evidence: 4 successive fresh `EventJournal(...).attach(log_dir)` calls on one run give `load_event_index` results of (no JSONL), `False`, `False`, `False`. Two bare `EventStore(path)` constructions give `False, True`.
   Suggested direction: accept a sidecar whose boundary fingerprint still matches when the source has only grown past `boundary` (the warm path already resumes `_scan_stream_unlocked` at `cached.boundary`), and cover attach, append, reattach through `EventJournal`.

2. **[P3] Tail records validated in place are parsed a second time** (`src/server/events.py:727-752`, `src/server/events.py:791-794`)
   A record the header scan cannot classify (e.g. legacy `"sequence":"1"`) is fully validated in `_scan_stream_unlocked` and stored with `event` set. If it falls inside the last `_EAGER_TAIL_RECORDS`, `_parse_eager_tail` reads and validates it again without checking `record.event`. The cost is bounded by the tail window and is still well below base (which parsed the whole file in this case), but `parsed_record_count` overstates the work. Evidence: a cold attach of a 1-record legacy journal reports `parsed_record_count == 2` (base: 1).
   Suggested direction: skip records whose `event` is already set in `_parse_eager_tail`.

3. **[P3] Text still describes the removed eager fallback** (`src/server/events.py:474-479`, `src/server/events.py:719-722`, `tests/server/test_lazy_event_store.py:43`)
   The `EventStore` docstring says "Any doubt during the scan discards the index and falls back to validating the whole file". The scan comment says the record "falls to the eager path". The `_EagerEventStore` docstring calls itself "the real fallback the store takes on any doubt". None of these mention the sidecar. `_read_unlocked`, `_records_from_events` and `_repair_legacy_sequences` now have no production caller; only that test oracle uses them.

AyanBinRafaih:
Out of delegated-merge scope by a single file, so `/merge-scoped` will refuse the PR as a whole and it needs a maintainer to merge. Everything else is within the TUI/backend-server capabilities in `.github/delegated-merge.toml`; the one path outside is:

- `scripts/benchmark_event_store_attach.py`

Splitting the benchmark out just to fit the policy would be gaming it, so this is better as a maintainer merge.

AyanBinRafaih:
UI smoke evidence for real journal attachment and the index sidecar: loaded copies of the same recorded run through the real Python server and tui, opened round 1, and captured at 150×40 with the dark theme using Microsoft [`tui-test`](https://github.com/microsoft/tui-test) 0.1.0-beta.3. The before/after PNG, SVG, and terminal text are byte-for-byte identical. This is the expected visible behavior; performance evidence remains the PR's regression tests and benchmarks.

| Before ([ea6f9073](https://github.com/uw-syfi/vibesys/commit/ea6f9073f061402819ed2e6e8ab8e1670d6f2469)) | After ([17a57e1e](https://github.com/uw-syfi/vibesys/commit/17a57e1e04df816b86a42ab3eefab5bf32f31de9)) |
|---|---|
| Before UI smoke | After UI smoke |

[Capture metadata and commands](https://raw.githubusercontent.com/AyanBinRafaih/vibe-serve/20f0be627d90721f2afb2f7cf6e8b20c561d4c70/pr-659/metadata.json) · [Before terminal text](https://raw.githubusercontent.com/AyanBinRafaih/vibe-serve/20f0be627d90721f2afb2f7cf6e8b20c561d4c70/pr-659/before.txt) · [After terminal text](https://raw.githubusercontent.com/AyanBinRafaih/vibe-serve/20f0be627d90721f2afb2f7cf6e8b20c561d4c70/pr-659/after.txt)

The attachment harness reports `running` after loading the completed historical journal; both captures use that same harness state.

vic-lsh:
The maintainer took over this PR to fix one review finding: `EventJournal.attach` appends SERVER_STARTED after the scan, so the exact size/mtime identity check rejected the sidecar on every real reattach. The index now accepts an append-only grown source when the head, interior samples, and boundary fingerprints of the indexed prefix still match, scans only the suffix, and republishes the extended sidecar; tests and the PR body are updated.

PR #661, opened 2026-09-10
TUI: preserve bounded stderr diagnostics from failed sandbox commands

### LINKED ISSUE #610: TUI: Preserve stderr when bounded sandbox output is truncated
### Observed behavior

Docker and Modal execution flatten output as `stdout + stderr` and then retain only the beginning of the combined string.

If a command produces more than the output cap on stdout and then fails with a short, useful stderr message, the retained output contains only stdout and a truncation marker. The actual compiler, checker, or process failure is lost.

The accuracy-gate event then labels this flattened content as `stream="stdout"`, so the TUI cannot recover the distinction or preferentially show the failure.

### Expected behavior

For failed commands, bounded output must preserve the diagnostic-bearing stderr tail even when stdout exceeds its budget. Preferably the executor contract preserves stdout and stderr separately. If the response shape must remain, truncation should allocate an explicit budget to each stream and retain the stderr tail for non-zero exits. The TUI event must identify the stream accurately.

### Reproduction

Run an accuracy command equivalent to:

```python
import sys

print("x" * 100001)
print("fatal compiler error", file=sys.stderr)
raise SystemExit(1)
```

With the default 100,000-character cap, observe that the TUI receives the stdout prefix and truncation marker but not `fatal compiler error`.

### Affected subsystem

### Environment

OS: Linux (debian) 
Backends: Docker and Modal sandbox execution  
Consumer: accuracy gate and TUI subprocess event presentation

### Impact

Blocks normal use

### Relevant logs

```shell
The failing regression should assert that the bounded TUI-facing event contains `fatal compiler error` even when stdout alone exceeds the configured cap.
```

#284, #516

### Pre-submission checks

AyanBinRafaih:
## Problem

Docker and Modal concatenated stdout and stderr, retained the combined prefix, then appended a marker outside the configured cap. A failed command with verbose stdout could therefore lose its short fatal stderr diagnostic. Accuracy and benchmark events also labeled the combined text as stdout, so the TUI could not identify the diagnostic stream.

## Solution

Both sandboxes now return an additive `SandboxExecutionResult`, a compatible `ExecuteResponse` subclass that retains bounded `stdout` and `stderr` alongside the combined `output`. All three strings share one character budget and preserve `output == stdout + stderr`. Successful and unknown-exit executions retain the existing combined-prefix behavior. Failed executions reserve half of the available payload for the stderr tail, give unused space back to either stream, and keep a deterministic stdout prefix. The truncation marker fits inside the cap and is attached to a stream whose content was truncated.

Accuracy and benchmark gates publish the retained stdout and stderr as separate semantic subprocess events. Backends that still return the base combined-only response retain the prior single stdout event for compatibility. Other sandbox consumers continue reading `.output`, which now contains the same retained stream content within a strict cap.

### Architecture

```mermaid
flowchart LR
    P[Docker or Modal process] --> B[bounded_execution_result]
    B -->|stdout prefix| R[SandboxExecutionResult]
    B -->|failed: stderr tail reservation| R
    R -->|compatible output| G[Gate result parsing and feedback]
    R -->|stdout field| SO[stdout subprocess event]
    R -->|stderr field| SE[stderr subprocess event]
    SO --> T[TUI]
    SE --> T
    L[Legacy combined-only backend] -->|one compatibility event| T
```

## Verification

| Before | After |
| --- | --- |
| Before | After |

The Docker and Modal regressions reproduce a failing command with 200 stdout characters and a short compiler error under an 80-character cap. On unchanged main, both fail because the result exceeds the cap and omits the compiler error. With this change, both backends retain the stderr tail, keep the stdout prefix, expose separate streams, and stay within the cap.

### Correctness properties

- Combined output never exceeds the configured nonnegative character cap, including the truncation marker.
- `output` is always exactly `stdout + stderr`.
- Failed truncated commands preserve a deterministic stdout prefix and reserve half of available payload capacity for the stderr tail.
- Empty streams, exact boundaries, Unicode content, stderr-only failures, long paired streams, tiny caps, negative caps, and nullable exit codes remain deterministic.
- Accuracy and benchmark TUI events label stream-aware sandbox content as stdout or stderr accurately.
- Base `ExecuteResponse` implementations retain their prior combined event behavior.
- Benchmark parsing, gate feedback, persisted gate tails, evaluator installation errors, and sandbox setup errors continue consuming the compatible `.output` field.

### Testing

- Unchanged-main Docker/Modal regression: `2 failed`; both responses exceeded the configured cap and omitted `fatal compiler error`.
- `VIBESYS_SKYPILOT_CALLER_STATE=/tmp/vibesys-610-review-focused uv run pytest -q --no-cov libs/vs-sandbox/tests/test_execution.py libs/vs-sandbox/tests/test_docker_sandbox.py libs/vs-sandbox/tests/test_modal_sandbox.py tests/vibesys/loops/test_gates.py`, `123 passed`.
- Sandbox, gate, and agent orchestration regression: `267 passed`.
- `uv run ty check`, passed.
- Targeted `uv run ruff check` and `uv run ruff format --check`, passed.
- `git diff --check`, passed.
- Full Python CI suite: 5,760 passed, 6 environment or platform skips. Global and module coverage gates passed; patch coverage including untracked source was 96%.
- Node 24.21.0 / Bun 1.3.9: 25 backend-client, 98 core-state, and 738 TUI tests passed. Protocol regeneration was stable; all repository formatting, lint, type, architecture, launcher, and client build checks passed.
- Real Docker validation could not run: the host denied access to `/var/run/docker.sock`, including outside the filesystem sandbox. Docker and Modal execution paths were verified through mocked provider APIs and the shared bounded-result implementation.

Closes #610

### DIFF
--- a/libs/vs-sandbox/src/vs_sandbox/__init__.py
+++ b/libs/vs-sandbox/src/vs_sandbox/__init__.py
@@ -13,6 +13,7 @@ from vs_sandbox.project_paths import ProjectPathPolicy, ProjectPathPolicyError
 
 if TYPE_CHECKING:
     from vs_sandbox.docker_sandbox import DockerSandbox
+    from vs_sandbox.execution import SandboxExecutionResult
     from vs_sandbox.host_resources import (
         HostResource,
         HostResourceAccess,
@@ -53,6 +54,7 @@ __all__ = [
     "ModalSandbox",
     "ProjectPathPolicy",
     "ProjectPathPolicyError",
+    "SandboxExecutionResult",
     "SandboxLifecycle",
     "SandboxLifecycleError",
     "SandboxLifecycleHooks",
@@ -66,6 +68,10 @@ __all__ = [
 
 
 def __getattr__(name: str) -> Any:  # noqa: ANN401, PLR0911  # tracked: #288
+    if name == "SandboxExecutionResult":
+        from vs_sandbox.execution import SandboxExecutionResult  # noqa: PLC0415
+
+        return SandboxExecutionResult
     if name == "DockerSandbox":
         from vs_sandbox.docker_sandbox import DockerSandbox  # noqa: PLC0415  # tracked: #288
 
--- a/libs/vs-sandbox/src/vs_sandbox/docker_sandbox.py
+++ b/libs/vs-sandbox/src/vs_sandbox/docker_sandbox.py
@@ -16,13 +16,13 @@ from typing import TYPE_CHECKING
 
 from deepagents.backends.protocol import (
     EditResult,
-    ExecuteResponse,
     FileDownloadResponse,
     FileUploadResponse,
     WriteResult,
 )
 from deepagents.backends.sandbox import BaseSandbox
 
+from vs_sandbox.execution import SandboxExecutionResult, bounded_execution_result
 from vs_sandbox.lifecycle import SandboxLifecycle, SandboxLifecycleHooks
 
 if TYPE_CHECKING:
@@ -188,7 +188,7 @@ class DockerSandbox(BaseSandbox):
             start_timeout: Timeout in seconds for the initial ``docker run``.
                 This bounds hidden image pulls or Docker daemon stalls before
                 the first agent has a chance to start.
-            max_output_bytes: Maximum output bytes before truncation.
+            max_output_bytes: Maximum output characters before truncation.
             env: Environment variables to set in the container.
             bind_mounts: List of (host_path, container_path, readonly) tuples.
             passthrough_paths: Container paths outside /workspace that should
@@ -617,7 +617,7 @@ class DockerSandbox(BaseSandbox):
         command: str,
         *,
         timeout: int | None = None,
-    ) -> ExecuteResponse:
+    ) -> SandboxExecutionResult:
         """Execute a command inside the Docker container."""
         if self._container_id is None:
             raise RuntimeError("Container not started — call start() first")  # noqa: TRY003  # tracked: #288
@@ -644,10 +644,11 @@ class DockerSandbox(BaseSandbox):
             )
         except subprocess.TimeoutExpired:
             self._log_cmd(exec_cmd, error=f"timeout after {effective_timeout}s")
-            return ExecuteResponse(
-                output=f"Command timed out after {effective_timeout}s",
+            return bounded_execution_result(
+                stdout="",
+                stderr=f"Command timed out after {effective_timeout}s",
                 exit_code=-1,
-                truncated=False,
+                max_output_chars=self._max_output_bytes,
             )
         self._log_cmd(exec_cmd, result)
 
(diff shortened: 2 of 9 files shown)


Nano-AI:
> [!NOTE]

**Summary:** Adds `bounded_execution_result()` and a `SandboxExecutionResult` subclass of `ExecuteResponse` that keep a stdout prefix and, for failed commands, a stderr tail within the output cap. The accuracy and benchmark gates now emit separate `stdout` and `stderr` subprocess events.

**Findings**

1. **[P1] Branch conflicts with main: ModalSandbox was deleted and DockerSandbox restructured** (`libs/vs-sandbox/src/vs_sandbox/modal_sandbox.py:634-688`, `libs/vs-sandbox/src/vs_sandbox/docker_sandbox.py:25`, `libs/vs-sandbox/src/vs_sandbox/__init__.py:70-74`)
   `git merge-tree --write-tree origin/main pr/661` reports modify/delete conflicts on `modal_sandbox.py` and `tests/test_modal_sandbox.py` (#684 removed ModalSandbox), plus textual conflicts in the `docker_sandbox.py` imports (#681 added the `WorkspaceSandbox` imports) and in `vs_sandbox.__getattr__` (main now resolves `AGENT_HOME` and `DockerSandbox` through one branch). The Modal changes and the Modal regression test have no target on main, and the PR body's "Docker and Modal" scope no longer applies.
   Suggested direction: rebase, drop the Modal hunks, and reapply the three `bounded_execution_result` call sites to `DockerSandbox.execute` on main (`docker_sandbox.py:894-956`). That method still has the same timeout, stderr-only, and combined-prefix structure, so the port should be mechanical.

The remaining checks passed. The new Docker and Modal regressions fail on the merge base (`ea6f907`) for the stated reason: the output is 115 characters against a cap of 80, and `fatal compiler error` is dropped. The stderr-only regression also shows that the old stderr-only path was unbounded (410 characters against a cap of 60). `OutputStream` already includes `"stderr"`, so no protocol or TS client regeneration is needed.

AyanBinRafaih:
Out of delegated-merge scope, so `/merge-scoped` will refuse this and it needs a maintainer to merge. It touches the sandbox library and a core gate, outside the TUI/backend-server capabilities in `.github/delegated-merge.toml`:

- `libs/vs-sandbox/src/vs_sandbox/__init__.py`
- `libs/vs-sandbox/src/vs_sandbox/docker_sandbox.py`
- `libs/vs-sandbox/src/vs_sandbox/execution.py`
- `libs/vs-sandbox/tests/test_docker_sandbox.py`
- `libs/vs-sandbox/tests/test_execution.py`
- `src/vibesys/loops/gates.py`
- `tests/vibesys/loops/test_gates.py`

AyanBinRafaih:
Captured the failed-gate transcript before and after this PR with the real TUI using [`microsoft/tui-test`](https://github.com/microsoft/tui-test). Each revision's unmodified `DockerSandbox.execute` received the same mocked process result: 200 stdout characters, `fatal compiler error` on stderr, exit code 1, and an 80-character cap. The returned result was replayed as the corresponding gate events, then I opened Round 1 and zoomed the transcript at 160x45.

| Before ([`bf0aa491`](https://github.com/uw-syfi/vibesys/commit/bf0aa491fb9610157e5119f2f576436aefa1050e)) | After ([`479cd672`](https://github.com/uw-syfi/vibesys/commit/479cd672762db5ec850334d71c5475fbe640f65b)) |
|---|---|
| Before: verbose stdout and a truncation marker, with the fatal compiler error missing | After: bounded stdout followed by the retained fatal compiler error |

Observed: before this PR, the stdout prefix consumes the cap and the fatal stderr diagnostic disappears. After this PR, the shared cap retains a shorter stdout prefix plus the `fatal compiler error` stderr tail.

The process itself is mocked, so this pair does not exercise a Docker daemon or Modal. It does exercise each revision's real sandbox truncation code and the real TUI rendering of the resulting event stream.

[Capture metadata, original SVGs, and fixture source](https://github.com/AyanBinRafaih/vibe-serve/tree/e77800737056c8712a152524495c14072e205d3d/pr-661).

PR #633, opened 2026-09-08
Add reusable microservice fuzzing and hotel correctness gates

Ryclic:
## Problem

Microservice authors need a reusable way to generate request histories and define correctness independently of those inputs, then use the result to gate VibeSys optimization. Hotel Reservation previously had a dedicated Go oracle rather than an implementation of that reusable interface.

## Solution

Add `vs_correctness` with input-only test cases, seeded generators, HTTP execution, pluggable custom execution, independent oracles, replay, bounded shrinking, and schema-v2 reports. Port Hotel Reservation's existing sequential correctness checks into a Python suite and register it as a packaged accuracy entrypoint. Remove the superseded hotel Go oracle while retaining its performance benchmark and other applications' accuracy support.

Packaged Python entrypoints use `${PYTHON}` to select VibeSys's interpreter locally without requiring a bare `python` on PATH, or the environment's `python3` in Docker/Modal. Isolated images own the compatible interpreter and evaluator dependencies. The hotel example configures the existing accuracy gate before benchmarking.

### Architecture

```mermaid
flowchart LR
    V[Verifier] --> G[User generator]
    G --> T[Input-only TestCase]
    T --> E[Executor]
    E --> O[Observation]
    O --> C[User oracle]
    T --> C
    C --> D[Decision and report]
    D --> A[VibeSys accuracy gate]
    A -->|PASS| B[Benchmark]
```

The library owns execution coordination and reports. Users own input grammars, business rules, reference models, deployment, readiness, and state reset. Hotel-specific catalog, schema, model, executor, and lifecycle code live in the evaluator package, outside DeathStarBench. The framework rejects metadata/expected-output fields in test cases; semantic separation inside arbitrary Python/custom payloads remains the author's responsibility.

### Correctness properties

- Oracle expectations derive from executed inputs and the oracle's own model.
- Successful observations must match the requested environment and every declared action position.
- Semantic failure, inconclusive execution, transport/reset errors, and empty suites cannot pass the gate.
- Shrinking retains reductions only after a fresh failing execution; reset is called for each attempt.
- Hotel checks cover protocol/schema, seeded data, authentication, search/recommendations, sequential reservation capacity and isolation, rejected-write behavior, read-after-write, and restart persistence.
- Concurrency/linearizability checking is not implemented; the removed Go oracle was also sequential.
- Schema-v1 cases containing metadata are rejected rather than silently replayed under changed semantics.

### Testing

Using the repository virtual environment and all repository source roots on `PYTHONPATH`:

[...]

PR #666, opened 2026-09-10
Server: name the event store a subscription streams from

### LINKED ISSUE #470: Tail subscription can re-deliver bootstrap-store sequences when the durable run log is shorter than the tail
## Observed behavior

`transport._stream` (`src/vibesys/server/transport.py`) re-bootstraps a tail
subscription when the watermark jumps past the tail after
`RunSupervisor.attach` swaps the event store mid-stream:

```python
if request.tail is not None and latest_sequence - cursor > request.tail:
    cursor, reported_floor = self._write_bootstrap(service, request, latest_sequence)
```

This only detects a store swap when the durable log is *longer* than the
tail. When a resumed run's durable log is shorter than the tail (e.g.
`tail=1000`, durable log has 200 events), `latest_sequence - cursor` never
exceeds `request.tail`, so no re-bootstrap fires.

The client has already folded the pre-attach bootstrap store's sequences
(`server_started`, `server_ready`, `run_started` at sequences 1..~11). After
attach, the durable store's events occupy overlapping low sequence numbers.
Two things go wrong at once:

- Genuine durable events at or below the client's already-folded cursor are
  silently skipped by the client's out-of-order guard.
- Later durable events re-use sequence numbers the client already consumed,
  with different content, and the client has no signal that these are not
  the same logical events replayed.

`history_after_sequence` also stays 0 in this path, so the client's backfill
mechanism has no reason to fetch the skipped prefix: it believes it already
has everything from the start.

## Expected behavior

The client should end up with state equivalent to a full replay of the
durable store, regardless of the relative sizes of the tail window and the
resumed run's durable log. A store swap during an active subscription should
always be detectable by the client, and should always cause it to discard
folded pre-attach state and re-fold from the durable store, the same way the
already-fixed long-log case does via `reduceEventRebootstrap`.

Relative-size watermark comparison cannot detect this case in general: it
needs a store-identity signal in the protocol (e.g. a store/epoch id
attached to `SubscribedMessage` and `EventBatchMessage`) so the client can
recognize a store swap independent of log length and re-fold from scratch.

## Reproduction

1. Start a run so the server begins streaming from the near-empty
   bootstrap store (pre-attach), with a subscription `tail` of 1000.
2. Let the client fold the bootstrap store's handful of spine events
   (`server_started`, `server_ready`, `run_started`) at sequences 1..~11.
3. Resume/attach a durable event log with roughly 200 events (i.e. a log
   shorter than the subscription's `tail`), so
   `latest_sequence - cursor <= request.tail` at the next watermark check in
   `transport._stream`.
4. Observe that no re-bootstrap occurs: durable events at or below the
   client's cursor are dropped by the out-of-order guard, later durable
   events overwrite already-folded sequence numbers with different content,
   and `history_after_sequence` stays 0 so backfill never recovers the
   skipped prefix.

## Affected subsystem

## Environment

Commit: current `vic/homebase-integration` branch, follows PR #468
(`perf: make TUI boot roughly flat in run length`), which fixed the
longer-than-tail case of this same defect class.

## Impact

Workaround exists (the affected window is a specific resume-timing/log-size
combination; most runs are unaffected), but when it triggers the client's
folded state genuinely diverges from the durable log with no client-visible
signal, and no automatic recovery path.

## Relevant logs

N/A - not observed from a live log; identified by code inspection of the
watermark-jump detection added in PR #468.

#468

AyanBinRafaih:
## Problem

Fixes #470. A run subscribes clients against a scratch bootstrap event store, then `EventJournal.attach` swaps in the durable log and re-appends the pending bootstrap events. When the durable log is nonempty those pending events are renumbered onto its tail, so the low sequences the client has already folded (`server_started`, `run_status_changed`, `run_started`, and whatever else the spine emitted) now name completely different events. The transport's only defense was PR #468's watermark check, `latest_sequence - cursor > request.tail`, which by construction cannot fire when the durable log is shorter than the tail.

Both halves of the client then fail silently. Durable events at or below the stale cursor are dropped by `foldEvent`'s out-of-order guard, and durable events above it are folded on top of unrelated state as if they continued it. `history_after_sequence` stays at 0 the whole time, so the backfill path is told the history is complete and never asks for the prefix it is missing.

Measured on `main` with a 202-event durable log and the standard `BOOTSTRAP_TAIL` of 1000: after attach the client's next batch begins at sequence 3, durable sequences 1 and 2 are never delivered in any batch, `run_started` never arrives, and the declared floor remains 0. The same run with `tail=40` re-bootstraps correctly, which is the shape of the defect: correctness depends on the accident of whether the resumed log happens to be longer than the tail.

## Solution

Sequences are only meaningful inside one event store, so the protocol now names the store. `EventStore` mints an opaque `store_id` (a `uuid4` hex) at construction, `EventBatchMessage` carries the id of the store each batch was read from, and both the transport and the TUI compare that id instead of inferring a swap from log lengths.

The identity names a *sequence space*, not a file and not a run. `path` cannot serve because a retired store can be reopened at the same path, and `run_id` cannot because it is reassigned in place. The distinction earns its keep at the one place the two diverge: attaching into an *empty* durable log re-appends the bootstrap events with their sequences preserved, so every sequence still names the same event and a subscriber has nothing to re-fold. `attach` therefore hands the retired store's identity to the new one in exactly that case, which is why a fresh run still streams incrementally rather than re-bootstrapping the instant its log is attached. Behavior for fresh runs is unchanged.

On the server, `subscription_checkpoint` takes the caller's `store_id` and reads the journal's current one under the same condition lock that guards the event read, so a swap cannot race a batch: on a mismatch it returns the new identity and reads nothing. The transport re-bootstraps on that mismatch. PR #468's watermark check is kept rather than replaced, now scoped to what it actually detects, a live burst inside one store that outruns the tail bound, so bounded replay still holds.

On the client, `SessionController` tracks the last store it saw and treats "the batch names a different store" and "the declared floor moved up" as two independent triggers for the same re-fold. Either one routes the batch through `applyEventRebootstrap` and resets the history floor instead of lowering it.

`store_id` is deliberately added to `EventBatchMessage` only, not to `SubscribedMessage` as the issue sketches. The bootstrap message's id would have no consumer: the client learns the identity from the first batch, and a field no reader uses is a field that drifts. One known gap is left open on purpose and is not in this issue's scope: a *reconnect* that redials with `resumed` set still resumes from the client cursor, so a store swap across a dropped connection is the same defect class through a different door. The identity now exists to fix it; wiring the resume dial is separate work.

### Architecture

```mermaid
flowchart TD
    A["EventStore.__init__: store_id = uuid4"] --> B["EventJournal.attach: empty durable log inherits the id, nonempty log renumbers and keeps its own"]
    B --> C["FrontendApi.subscription_checkpoint(cursor, store_id): compares under the condition lock, reads nothing on mismatch"]
    C --> D["unix_jsonl._stream: id differs -> _rebootstrap; tail watermark still guards a live burst"]
    D --> E["EventBatchMessage.store_id on the wire"]
    E --> F["SessionController: different store -> applyEventRebootstrap + reset floor; same store -> applyEventBatch + lower floor"]
```

## Verification

The defect is a client/server interaction, so it is pinned from both ends: server-side transport tests drive a real journal through a real attach over the real socket, and TUI tests drive the controller through the batch sequence that attach produces.

### Correctness properties

- A client's folded state after a mid-stream attach equals a full replay of the durable log, for any relative size of the durable log and the subscription tail. The regression test asserts the whole log is replayed from sequence 1 with `run_started` first, on the exact `tail=1000` / 202-event shape the issue reports.
- A store swap is detectable independent of log length. Identity comparison replaces size comparison as the *correctness* signal; the size check remains only as a bounded-replay guard within a single store.
- Attaching into an empty log preserves every sequence, so it preserves the identity, so a fresh run streams incrementally with no spurious re-fold. This is asserted at the journal level (sequences and identity both unchanged) and at the transport level (the post-attach batch starts at exactly `bootstrap.through_sequence + 1`).
- The identity read and the event read are atomic with respect to `attach`: both happen under `FrontendApi._condition`, so no batch can be stamped with one store's id and carry another store's events.
- Backward compatible in both directions. `store_id` defaults to `""`, and the client only treats a change as a swap once it has seen a non-null previous value, so a server that never reports identity leaves the client on exactly its pre-existing watermark behavior.
- After a re-bootstrap the declared floor is reset, not lowered, so the client reports the new store's real floor rather than the retired store's stale one and backfill asks for what is actually missing.

### Testing

- Fail-before: a standalone repro against `main` (real journal, real socket, real attach) showed the post-attach batch starting at sequence 3 with durable sequences 1 and 2 never delivered and `history_after_sequence` stuck at 0. With this branch the same script replays sequences 1 through 202 with `run_started` first and a correct floor.
- The client-side condition is load-bearing, not decorative: reverting only the `SessionController` change and rerunning fails 2 of the 4 new TUI tests ("reaches the state a full replay of the run log would have built" and "folds the run log prefix its stale cursor covered").
- New server tests. `tests/server/test_transport_subscription.py`: the issue's repro (short durable log, `tail=1000`, asserts a changed `store_id`, floor 0, whole-log replay, `run_started` first), the empty-log attach that must *not* re-bootstrap, live batches carrying the store they were read from, and a checkpoint reading nothing from a store the cursor predates. `tests/server/test_journal.py`: renumbering starts a new sequence space, an empty-log attach keeps it.
- New TUI tests: 4 cases under "a stream that re-bootstraps into a log shorter than the tail", covering equivalence with a full replay, folding the prefix the stale cursor covered, declaring a complete history so nothing is left to backfill, and extending rather than re-folding while the store stays the same.
- Full local suite: 5066 passed, 24 skipped, 5 failed, all five in `tests/vibesys/skypilot/test_evaluator_helper.py`. That file's whole import graph (167 modules) contains nothing this branch touches, and the same five fail identically on an unrelated branch in this environment, so they are local infrastructure. Unrelated but worth knowing: the suite parks at interpreter exit waiting on leaked non-daemon `_omnigent_runtime._serve` loop threads after the last test has reported.
- `uv run pytest tests/server`: 312 passed. `uv run pytest tests/architecture`: 248 passed, 22 skipped. `pnpm test:clients`: 25 backend-client, 98 core-state, 742 TUI, 0 failures. `./scripts/check_format.sh`, `./scripts/check_types.sh`, `uv run lint-imports`, `pnpm check:ts`, `pnpm check:ts-architecture`, `pnpm check:clients`, `pnpm build:clients` all pass. `uv run ruff check src/server tests/server` is clean.
- The generated protocol files were regenerated with `pnpm --dir clients/backend-client generate:protocol`, so the CI parity check and `tests/server/test_protocol_schema.py` both agree with `src/server/api/protocol.py`.
- Merges cleanly (`git merge-tree --write-tree`) against the current `main` tip and against 23 of the 25 open sibling branches. The two exceptions, #633 and #467, conflict in `resources/evaluators/microservice/` and `src/vibesys/loops/` respectively, files this branch does not touch, and both conflict with plain `main` the same way.

Closes #470.

### DIFF
--- a/clients/backend-client/src/generated/protocol.generated.ts
+++ b/clients/backend-client/src/generated/protocol.generated.ts
@@ -458,6 +458,7 @@ export type Type16 = "event_batch";
 export type Events1 = RunEvent[];
 export type ThroughSequence = number;
 export type ActiveExecutions1 = ActiveAgentExecution[];
+export type StoreId = string;
 export type HistoryAfterSequence = number;
 export type Type17 = "protocol_error";
 export type RequestId16 = string | null;
@@ -1106,6 +1107,7 @@ export interface EventBatchMessage {
   events: Events1;
   through_sequence?: ThroughSequence;
   active_executions?: ActiveExecutions1;
+  store_id?: StoreId;
   history_after_sequence?: HistoryAfterSequence;
 }
 export interface ProtocolErrorMessage {
--- a/clients/backend-client/src/generated/protocol.schema.json
+++ b/clients/backend-client/src/generated/protocol.schema.json
@@ -1135,6 +1135,11 @@
           "title": "Active Executions",
           "type": "array"
         },
+        "store_id": {
+          "default": "",
+          "title": "Store Id",
+          "type": "string"
+        },
         "history_after_sequence": {
           "default": 0,
           "minimum": 0,
--- a/clients/tui/dev/mock-server.ts
+++ b/clients/tui/dev/mock-server.ts
@@ -339,6 +339,17 @@ class Replay {
   }
 
   /**
+   * The one sequence space this replay ever serves.
+   *
+   * A real server swaps stores when it attaches a run's durable log, and the
+   * client re-folds when the id changes. The mock replays a finished log from
+   * the start, so its id is constant and no batch ever asks for a re-fold.
+   */
+  get storeId(): string {
+    return `mock-store-${this.runId}`;
+  }
+
+  /**
    * Sequence of the newest event delivered so far, or 0 before any.
    *
    * The cursor is a count, not an index, so a zero cursor means nothing has
@@ -445,6 +456,7 @@ class Replay {
       events: [event],
       through_sequence: event.sequence,
       active_executions: this.activeExecutions,
+      store_id: this.storeId,
     });
     if (this.#options.verbose) {
       process.stderr.write(`mock: seq ${String(event.sequence)} ${String(event.type)}\n`);
@@ -519,6 +531,7 @@ function main(): void {
             events: replay.delivered,
             through_sequence: replay.latestSequence,
             active_executions: replay.activeExecutions,
+            store_id: replay.storeId,
             // 0 means the stream carries its whole history, so the TUI never
             // asks for a backfill it cannot get.
             history_after_sequence: 0,
--- a/clients/tui/src/session-controller.test.ts
+++ b/clients/tui/src/session-controller.test.ts
@@ -1833,6 +1833,88 @@ describe('a stream that re-bootstraps at a raised floor', () => {
   });
 });
 
+/**
+ * The same late attach, but the run log is shorter than the subscription's
+ * tail, so the re-bootstrap replays it whole and declares floor 0 like the
+ * batch before it. Nothing about the floors distinguishes the two logs; only
+ * the store they name does.
+ */
+describe('a stream that re-bootstraps into a log shorter than the tail', () => {
+  const runLog: RunEvent[] = [
+    {
+      ...event(1, 'run_started'),
+      data: {kind: 'run_started', outer_loop: 'agent', input: '.', max_rounds: 3},
+    },
+    event(2, 'agent_output_chunk', 'two\n'),
+    roundFinished(3, 1),
+    event(4, 'agent_output_chunk', 'four\n'),
+  ];
+  // What the client folded from the server's own log before the attach: the
+  // same sequence numbers, different events.
+  const preAttach = [
+    event(1, 'agent_output_chunk', 'server started\n'),
+    event(2, 'agent_output_chunk', 'server ready\n'),
+  ];
+
+  async function rebootstrapped(transport: HistoryTransport): Promise<SocketSessionController> {
+    const controller = new SocketSessionController(transport);
+    await controller.start();
+    transport.emitBatch(preAttach, 0, 'bootstrap-store');
+    transport.emitBatch(runLog, 0, 'run-store');
+    return controller;
+  }
+
+  it('reaches the state a full replay of the run log would have built', async () => {
+    const replayedTransport = new HistoryTransport(runLog);
+    const replayed = new SocketSessionController(replayedTransport);
+    await replayed.start();
+    replayedTransport.emitBatch(runLog, 0, 'run-store');
+
+    const controller = await rebootstrapped(new HistoryTransport(runLog));
+
+    expect(controller.state.core.transcript).toEqual(replayed.state.core.transcript);
+    expect(controller.state.core.rounds).toEqual(replayed.state.core.rounds);
+    expect(controller.state.core.sequence).toBe(replayed.state.core.sequence);
+  });
+
+  it('folds the run log prefix its stale cursor covered', async () => {
+    const controller = await rebootstrapped(new HistoryTransport(runLog));
+
+    // `run_started` is sequence 1 in the attached log and sequence 1 was
+    // already folded from the log it replaces, so the out-of-order guard drops
+    // it unless the batch is recognized as superseding what came before.
+    expect(controller.state.core.maxRounds).toBe(3);
+    expect(controller.state.core.outerLoop).toBe('agent');
+    expect(controller.state.core.rounds.map(round => round.number)).toEqual([1]);
+  });
+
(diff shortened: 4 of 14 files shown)


AyanBinRafaih:
The four `Build <platform>` checks on this PR are red for a reason that predates it and is unrelated to its diff.

They fail at `installed release verification` with:

```
File ".../site-packages/server/chat/options.py", line 17, in <module>
  from vibesys._agent_cli.opencode import OPENCODE_DEFAULT_MODEL
File ".../site-packages/vibesys/_agent_cli/opencode.py", line 4, in <module>
  from agentshim.events import AgentEventHandler
ModuleNotFoundError: No module named 'agentshim.events'
```

`src/vibesys/_agent_cli/opencode.py` arrived with #630 (`ea6f9073`, the current `main` tip). That module is imported at import time by `server/chat/options.py`, so it is on the path of every server start, including the installed-release smoke.

The release workflow is path-filtered and #630 touched no `clients/` files, so its Build jobs never ran. #634 was the last PR before it that did touch `clients/`, and all four of its Build jobs passed. This PR is the first since #630 to touch `clients/` (a regenerated protocol file, the mock server, and the session controller), so it is the first to run that workflow against the new import, and it surfaces the break rather than causing it.

Note: the local venv resolves `agentshim` 0.5.0 with an `events` module present, so the release install is resolving a different distribution than `uv.lock` pins. Nothing in this PR touches `opencode.py`, `pyproject.toml`, or `uv.lock`.

PR #667, opened 2026-09-10
Chat: file an experiment-chat execution's output under the thread that asked

AyanBinRafaih:
## Problem

An experiment-chat execution's presentation output (the `agent_output_chunk` and tool events it emits while answering) reached the wire with no `chat_thread_id`, so `applyChatEvent` filed it under the run's default thread. Only the terminal `chat` answer event carried the real thread.

This is latent while a chat answer arrives whole, but it becomes a correctness hazard the moment chat output streams incrementally: a named thread's partial answer would accumulate in the default transcript while its terminal answer folded in under the real thread, and `foldChatAnswer` cannot merge two transcripts.

## Solution

`ExecutionTracker.presentation_scope` now carries a `chat_thread_id`, and `publish_presentation` stamps it on every event scoped to that execution. `ExperimentChatSession` passes the thread it answers on into the scope. The run's default chat carries `None`, matching how the terminal `chat` event already names that thread, so a streamed partial answer and the terminal answer that folds over it agree on their thread. Non-chat agents pass no thread and are unchanged, and the scope restores its enclosing values, including under nesting.

## Verification

Server tests, run in CI.

### Correctness properties

- A named thread's chat presentation events carry that thread end to end (`test_chat_presentation_events_carry_the_thread_that_asked`).
- The default chat's presentation output carries no thread, exactly as its terminal answer does, so the two land in one client transcript (`test_default_chat_presentation_matches_its_terminal_answer`).
- Presentation events outside chat carry no thread (`test_presentation_events_outside_chat_carry_no_thread`).
- `presentation_scope` restores the enclosing scope, including under nesting (`test_presentation_scope_restores_the_enclosing_scope`).
- An `ExperimentChatSession` driven over real server parts files its output under its own thread (`test_streamed_output_is_filed_under_the_thread_that_asked`).

## Scope

This does not deliver #463's remaining streaming criterion (partial tokens rendered in the pane as they arrive). That streaming lives in the Claude CLI driver under `src/vibesys/`, which is vibesys core and outside this PR: an earlier revision of this branch carried it, but the merge with `main` resolved those core files back to their non-streaming form. This PR keeps only the server-side thread plumbing the streamed answers will need, so it is filed under the right thread once the driver work lands. Issue #463 stays open for that driver work.

Refs #463.


Nano-AI:
> [!NOTE]

**Summary:** Streams Claude's chat answer into the pane: the Claude driver passes `--include-partial-messages` and forwards `text_delta` events as `AgentEventKind.TEXT`, and chat presentation events now carry `chat_thread_id` so streamed chunks land in the transcript of the thread that asked.

**Findings**

1. **[P1] Branch no longer merges: `main` deleted the module the streaming logic lives in** (`src/vibesys/_agent_cli/claude.py:174-190`, `src/vibesys/_agent_cli/claude.py:308-337`)
   Since the `ea6f907` branch point, #668 replaced the `_agent_cli` fork with agentshim's public API, and #672 pins `agentshim>=0.6.1,<0.7`. Against `main` at `d6cd655`, `git merge-tree --write-tree origin/main pr/667` reports modify/delete on `src/vibesys/_agent_cli/claude.py` and `tests/vibesys/_agent_cli/test_claude.py`, plus content conflicts in `src/vibesys/agents/drivers/agentshim.py`, `tests/server/test_chat_session.py` and `tests/vibesys/agents/drivers/test_agentshim_driver.py`. GitHub reports `CONFLICTING`. The body's "merges cleanly against main" was accurate when written.

**Port direction** (checked against the agentshim 0.6.1 sdist)

- The wrong-channel half is already fixed on `main`. `_translate` maps `agentshim.AssistantText` to `AgentEventKind.TEXT` (`src/vibesys/agents/drivers/agentshim.py:275-280` on `main`), which `_LoggerObserver` sends to `log_text` on the `assistant` channel. Claude text already arrives there, one event per finished block. That makes the new `on_text` handler method unnecessary, and the "analysis to assistant for every claude agent" change has already shipped.
- Token streaming can't be reached through 0.6.1's public API. The flag itself could go through `agentshim.TurnRequest.extra_args`, since Claude's `build_argv` appends it. But `providers/claude/events.py::parse_frame` returns `None` for every frame type other than `system`, `assistant`, `user` and `result`, so the library drops `stream_event` lines internally. The driver no longer sees raw stdout.
- Suggested direction: an upstream agentshim change to `ClaudeStreamParser` that:
  - handles `stream_event`, `content_block_delta` and `text_delta` by emitting `AssistantText` per delta;
  - suppresses the whole-block `TextBlock` echo in `_assistant` but still emits that frame's `UsageReport`;
  - adds `--include-partial-messages` to the Claude argv.

  The Copilot parser in the same release already streams deltas and suppresses the echo (`providers/copilot/parser.py`, `_delta` and `_message`). The `_PartialAssistantText` dedup and its test cases fit that change naturally. With it in place, VibeSys needs only a version bump, because `_translate` already sends each `AssistantText` to `TEXT`. A VibeSys-side `CommandExecutor` wrapper that tees stdout would also work through the public API, but it puts Claude wire-format parsing back into VibeSys after #668 moved it out.

**What survives unchanged**

- The `chat_thread_id` threading merges cleanly: `src/server/execution.py:130-143`, `src/server/execution.py:378`, `src/server/chat/session.py:62-169`, `src/server/chat/factory.py:310`, plus `tests/server/test_execution_tracker.py`. Applied onto `d6cd655` with agentshim 0.6.1, together with a one-line resolution in `tests/server/test_chat_session.py` (`chat_thread_id=thread_id` in `_chat`), `pytest tests/server/test_execution_tracker.py tests/server/test_chat_session.py` gives 28 passed. Without that line, 10 fail with `missing 1 required positional argument: 'chat_thread_id'`.
- This part is worth landing now. `main` already publishes each Claude text block on `assistant` without `chat_thread_id`, and `applyChatEvent` (`clients/core-state/src/core-state.ts:894` on `main`) files such chunks under the default thread. A separate PR with just the thread stamping would fix that without waiting on agentshim.

PR #691, opened 2026-09-12
feat(tui): resize the Agents pane by columns with < > and =

Nano-AI:
## Problem

The Agents graph and the transcript beside it split the row at whatever width automatic
sizing picks. An operator who wants to read a wide graph, or wants the graph out of the way
to read a long transcript, has no way to say so.

This replaces the design in the first revision of this PR, which stepped a ratio and, once
the graph hit its floor, swapped the pane to the stacked agent list. Both halves were wrong:
a resize should change a width in columns, and the stacked list cannot draw a node with
in-degree above one, so it is not a view the graph can degrade into.

| before | after |
| --- | --- |
| before | after |

Both at 160x40, the after taken with six `>` presses. What to look at: the pane boundary moves
from column 82 to 91, each node box gains two columns of interior, the transcript rewraps, and
the key-help line gains `<>=: width`. The range is deliberately modest, since a step is two
columns and the clamp stops at this round's natural ceiling; six presses is already past it.

## Solution

`<` and `>` step the pane by two columns and store the result in `graphWidthOverride`, a
session-only piece of UI state. `=` clears it, the way vim's `<C-w>=` does. `null` keeps
#686's automatic sizing untouched, including its rule that no agent name is ever cut, and
its stacked-list fallback on a narrow terminal.

An explicit width is sticky, which is the point of asking for one and the reason `=` exists:
without a reset, one press would cost the pane its automatic sizing for the life of the
process. Truncation is what the override buys. Below the full-name floor a label loses
characters to `…`, and only ever under an explicit override.

### Architecture

The override is clamped to what can actually be drawn: `agentGraphMinWidth` at the low end
(every stage column at its 14-column floor plus a gutter per gap, never under
`STACKED_WIDTH`, so an override cuts agent names and not the round heading), and
`agentPaneCeiling` at the high end, past which more columns are padding rather than name.
The transcript keeps `TRANSCRIPT_MIN` either way.

**A press that would not change the rendered width stores nothing.** This one rule replaces
four special cases, each of which would otherwise arm an invisible override and pin every
later round to it: the stacked list, a round with no phases yet, a round of one short-named
stage, and a terminal where automatic sizing already sits at the ceiling.

### Correctness properties

- Automatic sizing is byte-for-byte unchanged and still never truncates.
- `<` can never reach the stacked list, which stays a narrow-terminal fallback: from the list
  `<` is inert and `>` opens the narrowest graph.
- The clamp never yields a pane wider than the terminal, narrower than `STACKED_WIDTH`, or
  one that leaves the transcript under `TRANSCRIPT_MIN`; it is idempotent and re-runs against
  the current round, so a stale override survives neither a resize nor a stage-count change.
- The keys are inert while a pane is zoomed, while the experiment log holds the row, while a
  visualization split has room beside the transcript, and while the command input has text.
- Fan-in (in-degree above one) still draws every edge at a narrowed width.

### Testing

- `pnpm test:clients`: 856 pass, 0 fail.
- `pnpm check:ts`, `pnpm check:clients`, `pnpm check:ts-architecture`,
  `pnpm test:ts-architecture`, `scripts/check_doc_links.py`: clean.
- Clamp invariants checked by exhaustive sweep over stage counts 0-5, agent-name lengths
  1-40, terminal widths 20-300 and overrides, comparing the rendered result against storing
  nothing; plus BFS reachability of the clamp band and key sequences across a terminal resize
  and a round switch. Each guard was mutation-checked: reverting it fails a named test.

### DIFF
--- a/clients/tui/src/session-controller.ts
+++ b/clients/tui/src/session-controller.ts
@@ -79,6 +79,7 @@ import {
   setChatThreadPending,
   setDesignLog,
   setExperiments,
+  setGraphWidthOverride,
   setPaneContent,
   setTheme,
   showDetail,
@@ -126,6 +127,8 @@ export interface SessionController {
   focusRound(focus: RoundFocus): void;
   selectNextTodo(delta: number): void;
   toggleTodos(): void;
+  /** `<`/`>`: the Agents pane's explicit column width, already clamped; `=`: null. */
+  setGraphWidthOverride(width: number | null): void;
   /** Expands the latest prompt in view; the view owns what "latest" means. */
   togglePrompt(): void;
   onTogglePrompt(handler: () => void): void;
@@ -428,6 +431,10 @@ export class SocketSessionController implements SessionController {
 
   toggleTodos(): void {
     this.#setState(toggleTodos(this.#state));
+  }
+
+  setGraphWidthOverride(width: number | null): void {
+    this.#setState(setGraphWidthOverride(this.#state, width));
   }
 
   setTheme(themeName: ThemeName): void {
--- a/clients/tui/src/session-model.ts
+++ b/clients/tui/src/session-model.ts
@@ -66,6 +66,17 @@ export interface SessionState {
   /** Non-null while the composer's inline command menu is open. */
   chatMenu: ChatMenu | null;
   todosExpanded: boolean;
+  /**
+   * Explicit column width for the Agents pane, set and stepped by `<`/`>` and
+   * cleared by `=`. `null` means automatic: `agent-map.ts#agentPaneWidth`
+   * decides, exactly as it always has, including its no-truncation floor and
+   * the stacked-list fallback on a narrow terminal. Once set it is sticky, so
+   * the pane stops following the terminal and stops growing with longer agent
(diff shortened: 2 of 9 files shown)


Nano-AI:
Force-pushed a full rebuild: the collapse design is gone.

The previous revision stepped a ratio and, once the graph hit its floor, swapped the pane to the stacked agent list. Both halves were wrong. A resize should change a width in columns, and the stacked list cannot draw a node with in-degree above one, so the graph cannot degrade into it.

This revision steps two columns at a time into a `graphWidthOverride`, adds `=` to return to automatic sizing, and keeps the stacked list strictly as an automatic narrow-terminal fallback that no key can reach. Rebased onto main; the diff is 9 files.

Reviewed in four rounds. Three of them found the same bug class: a keypress that changed nothing on screen still stored an override, converting automatic sizing into a sticky one that pinned every later round to a narrow graph with truncated names. Reached first via the stacked fallback, then a too-narrow terminal, then a round with no phases. It is now one rule, "a press that would not change the rendered width stores nothing", which subsumed all three plus a fourth case (the first `>` when automatic sizing already sits at the ceiling) and let a special-case branch be deleted.

PR #714, opened 2026-09-12
fix(tui): stop painting an active node's outbound edges as live

### LINKED ISSUE #713: TUI: don't render an active node's outbound edges as live in the agent graph
### Observed behavior

In the agent graph (Agents pane), an edge leaving an **active** node is drawn in the live accent color regardless of what it points at. `edgeTone` in `clients/tui/src/ui/agent-graph.ts:378-390` is:

```ts
function edgeTone(source: AgentPhase, target: AgentPhase): EdgeTone {
  if (
    source.status === 'failed' ||
    source.status === 'cancelled' ||
    source.status === 'interrupted'
  )
    return 'failed';
  if (source.status === 'active' || (source.status === 'completed' && target.status === 'active')) {
    return 'live';
  }
  if (source.status === 'completed' && target.status !== 'pending') return 'done';
  return 'idle';
}
```

The first disjunct, `source.status === 'active'`, marks every edge leaving an active node `'live'`, including edges into stages that have not started. An active node has not produced any output yet, so painting its outbound edges as live shows dataflow that has not happened. What an operator actually sees on the agent graph: an active node's outbound edges light up as though data has already flowed out of it, before it has finished anything, i.e. spurious outward highlighting.

### Expected behavior

`'live'` should mean data has flowed along an edge into something that is now working on it: a completed source feeding an active target. That is already the second disjunct, `source.status === 'completed' && target.status === 'active'`, and it is correct. An edge leaving a still-active source, into a target that has not started, should render `'idle'` (the resting tone) until the source completes.

### Reproduction

1. Call `layoutAgentGraph` with a two-stage chain where the source is active and the target is pending, e.g. `[{kind: 'implementer', status: 'active', ...}, {kind: 'judge', status: 'pending', ...}]`.
2. Inspect `graph.cells`: every cell's `tone` is `'live'`.
3. Expected: every cell's `tone` is `'idle'`, since the implementer has not produced anything for the judge to consume yet.

Equivalently, in a live run: watch the Agents pane while a stage is active and its downstream stage is still pending. The edge between them renders live before the active stage has finished.

### Affected subsystem

CLI and packaging

### Environment

Commit: d363042907deed4298dd689fd40771abd542024b (origin/main)
OS: n/a (pure rendering/geometry logic, reproduces on any platform)
Hardware: n/a
Backend: n/a, bug is in `clients/tui` presentation logic only
Profiler: n/a

### Impact

Minor or cosmetic

### Relevant logs

```
clients/tui/src/ui/agent-graph.ts:378-390, edgeTone(source, target):

  if (source.status === 'active' || (source.status === 'completed' && target.status === 'active')) {
    return 'live';
  }

The `source.status === 'active' ||` disjunct is the bug: it ignores the
target entirely and marks every edge leaving an active node live.
```

#284

Nano-AI:
## Problem

In the agent graph (Agents pane), an edge leaving an active node renders in
the live accent color regardless of what it points at. An operator sees
`implementer -> judge` light up as though data already flowed out of the
implementer while it is still active. `edgeTone()` treats
`source.status === 'active'` alone as sufficient for `'live'`, ignoring the
target. Closes #713.

## Solution

Drop the `source.status === 'active' ||` disjunct from `edgeTone()`.
`'live'` now requires only the already-correct second clause: a completed
source feeding an active target, i.e. data that has actually flowed into
work now in progress. An active source's outbound edges fall through to
`'idle'` until the source completes. Reviewers should check the fan-out/
fan-in tests: an active node's outbound edges all go idle, while its
completed-source inbound edges stay live.

### Architecture

Change is confined to `edgeTone()` in `clients/tui/src/ui/agent-graph.ts`.
Its only consumer is `edgeColor()` in `agent-map.ts` (presentational); the
stacked-list fallback never calls `layoutAgentGraph` and there is no legend.

## Verification

| Before (`main`) | After |
| --- | --- |
| before | after |

Round 1 with `implementer` active (fixture cut after its `invocation_started`), 140x40, dark theme, cropped to the Agents pane. `implementer -> judge` goes from live to idle; `orchestrator -> implementer` stays live.

Reproduced the active-to-pending case from #713 against pre-fix code (tone
read `live`) and confirmed it reads `idle` post-fix.

### Correctness properties

- `'live'` implies a completed source and an active target.
- An active source's outbound edges are `'idle'` until it completes.
- Failed/cancelled/interrupted sources still force `'failed'` regardless of target.
- A completed source into any non-pending, non-active target still reads `'done'`.

### Testing

- `pnpm --filter @vibesys/tui test`: 832 pass, 0 fail (30 files)
- `pnpm check:clients`: clean

### DIFF
--- a/clients/tui/src/ui/agent-graph.test.ts
+++ b/clients/tui/src/ui/agent-graph.test.ts
@@ -75,10 +75,63 @@ describe('layoutAgentGraph', () => {
   test('tones an edge by the phases it connects', () => {
     const four = [...CHAIN, phase('profiler', 'pending')];
     const graph = layoutAgentGraph(four, graphPaneBounds(four).max - 4);
-    // The frontier glows: an edge is live while either end is running. Two
-    // stages that have not run yet stay idle.
+    // Live means data has flowed: a completed source feeding an active
+    // target. A stage that has not produced anything yet, or that feeds a
+    // stage that has not started, stays idle.
     expect(graph.cells.some(cell => cell.tone === 'live')).toBe(true);
     expect(graph.cells.some(cell => cell.tone === 'idle')).toBe(true);
+  });
+
+  test('an edge from an active source to a pending target is idle, not live', () => {
+    // The active node has not produced anything yet, so its outbound edge
+    // must not be painted as live dataflow.
+    const activeToPending = [phase('implementer', 'active'), phase('judge', 'pending')];
+    const graph = layoutAgentGraph(activeToPending, graphPaneBounds(activeToPending).max - 4);
+    expect(graph.cells.every(cell => cell.tone === 'idle')).toBe(true);
+  });
+
+  test('an edge from a completed source to an active target is still live', () => {
+    const completedToActive = [phase('implementer', 'completed'), phase('judge', 'active')];
+    const graph = layoutAgentGraph(completedToActive, graphPaneBounds(completedToActive).max - 4);
+    expect(graph.cells.every(cell => cell.tone === 'live')).toBe(true);
+  });
+
+  test.each([
+    'failed',
+    'cancelled',
+    'interrupted',
+  ] as const)('a %s source still fails the edge, even into an active target', status => {
+    const badSource = [phase('implementer', status), phase('judge', 'active')];
+    const graph = layoutAgentGraph(badSource, graphPaneBounds(badSource).max - 4);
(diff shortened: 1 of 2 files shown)


Nano-AI:
Superseded by #728: reopened from a fork after the delegated-merge role change (Triage cannot push to this branch) and rebased onto main with no code change.

PR #715, opened 2026-09-12
fix(tui): align the unowned-round row to the Experiments table's own columns

### LINKED ISSUE #712: Experiments pane shows hypothesis column headers over rows with no columns
## Observed behavior

The Experiments pane draws the hypothesis table's column header
(Hypothesis / Rounds / Implementation Details / Measured / Outcome) over a row
that has none of those columns.

`#renderTable` in `clients/tui/src/ui/experiment-log.ts` (around line 252)
draws the header unconditionally via `headerRow(columns, ...)`, using the same
`columns` object real hypothesis rows use. But when the row is an "unowned"
round, `#roundRow` (`experiment-log.ts:415-435`) builds one concatenated
string cell instead: `` ` Round ${roundNumber} · recorded agent turns · no
hypothesis` ``. It never calls `entryCells` and never consults `columns`. The
result: five column headers sit above a row with no columns, and the row
itself leaves roughly 80% of the pane's width empty.

An unowned round is a round that produced agent turns but falls outside every
hypothesis's `[first_round, last_round]` range
(`unownedExperimentRounds`, `clients/tui/src/session-model.ts:1022-1040`).
This is not an edge case: a run typically profiles before proposing
hypothesis 1, so nearly every run produces an unowned round first. The
mismatched-header state is the default landing view most operators see, not a
rare one.

A correct empty state already exists for the genuine zero-hypotheses case
(`experiment-log.ts:206-209`): no header, "No hypotheses have been recorded
yet." with footer "The first one appears once the orchestrator has planned a
round." Per `#render` (~lines 197-211), that path only fires when
`unownedExperimentRounds(state).length === 0`. As soon as one unowned round
exists, `#renderTable` runs instead and the mismatched header appears.

Contributing factors that make the mismatch harder to read at a glance:
- No rule is drawn under the header row, so it reads as one block with the
  row beneath it rather than a separate section.
- Every column, including the numeric Measured, Rounds, and Kept columns, is
  left-aligned via `padEnd`/`fitColumn` (`headerRow` line 594, `fitColumn`
  line 649).
- No bold or weight differentiates the header from body rows anywhere in this
  table.

## Expected behavior

The hypothesis column header is not shown over a row that structurally isn't
a hypothesis row. Unowned rounds either render under no column header (as the
existing "No hypotheses have been recorded yet." state already does) or get
their own labeled section that is visibly separate from the hypothesis
table's header.

## Reproduction

1. Start or resume a run where the first round (commonly a profiling round)
   has produced agent turns but no hypothesis has been proposed yet.
2. Open the Experiments pane.
3. Observe the Hypothesis / Rounds / Implementation Details / Measured /
   Outcome header drawn above the "Round N · recorded agent turns · no
   hypothesis" row, which fills only a fraction of the row's width.

## Affected subsystem

CLI and packaging

## Environment

Commit: d363042 (origin/main)
Component: clients/tui/src/ui/experiment-log.ts

## Impact

Minor or cosmetic

## Relevant logs

N/A: visual layout defect, no error output.

Closed and unrelated to this specific defect: #465 and PR #487 (unlabeled
Measured number), #593 (Measured baseline semantics), #285 (original
hypothesis-log feature). PR #435 (merged) introduced the unowned-round
concept this bug depends on. Open PR #706 (`ayan/log-cjk-width`) touches the
same `headerRow`/`fitColumn` functions but for a different concern (padding
and truncation by display width, not by code units); noted here so the two
aren't worked twice. Fix in progress on branch `adi/experiments-table`.

## Pre-submission checks

Nano-AI:
## Problem

The Experiments pane draws the hypothesis table header (Hypothesis / Rounds /
Implementation Details / Measured / Outcome) over a row with no columns:
`#roundRow` concatenated one string for an unowned round instead of calling
`entryCells`, and `#renderTable` drew the header unconditionally. An unowned
round (agent turns outside every hypothesis's range) is the default landing
state, since a task profiles before proposing hypothesis 1.

Closes #712

## Solution

Route the round row through the same column grid `entryCells` uses
(`unownedRoundCells`/`unownedRoundRow`), so it lands under the real headers
with a genuine round number and the honest "recorded agent turns" label;
other fields render the existing `—` placeholder. Add a permanent
fixed-height rule under the header. Right-align the numeric columns (Rounds,
Measured, Kept) via a new `align` parameter on the existing `fitColumn`
helper: open PR #706 (different author) reroutes that same helper for CJK
display width, so both fixes converge on one primitive rather than two.

### Architecture

`clients/tui/src/ui/experiment-log.ts` (`ExperimentLogView`) owns the
Experiments pane's table layout and column widths.

## Verification

| Before (`main`) | After |
| --- | --- |
| before | after |

Default landing state (an unowned round), 140x36, dark theme, cropped to the Experiments pane.

Formatter tests assert `unownedRoundCells` matches `entryCells`'s column
widths at every panel width; renderer tests confirm on-screen alignment.

### Correctness properties

- An unowned-round row shares column widths with a hypothesis row.
- Absent values always render the existing `—` placeholder, never empty.
- `fitColumn` defaults to left alignment; existing columns are unchanged.
- The header rule is fixed height, independent of row count.

### Testing

- `pnpm --filter @vibesys/tui test`: 834 pass, 0 fail (30 files)
- `pnpm check:clients`: clean

### DIFF
--- a/clients/tui/src/ui/app.test.ts
+++ b/clients/tui/src/ui/app.test.ts
@@ -5260,7 +5260,13 @@ describe('theming', () => {
     const kickoff = await frameAfter(testRenderer);
     expect(kickoff).toContain('Planning Hypothesis 1');
     expect(kickoff).toContain('Round 2');
-    expect(kickoff).toContain('Round 1 · recorded agent turns · no hypothesis');
+    // The round row now lands on the same column grid a hypothesis row uses
+    // (see experiment-log.test.ts), so its cells are checked rather than the
+    // single string the row used to be. "(no hypothes…" is the ID column,
+    // which never drops at a narrow width, unlike the claim column.
+    const unassociatedLine = kickoff.split('\n').find(line => line.includes('(no hypothes'));
+    expect(unassociatedLine).toBeDefined();
+    expect(unassociatedLine).toMatch(/\b1\b/);
 
     testRenderer.mockInput.pressEnter();
     const round = await frameAfter(testRenderer);
@@ -5285,9 +5291,10 @@ describe('theming', () => {
     registerCleanup(testRenderer.renderer, app);
     await controller.openExperimentLog();
 
-    expect(await frameAfter(testRenderer)).toContain(
-      'Round 7 · recorded agent turns · no hypothesis',
-    );
+    const indexed = await frameAfter(testRenderer);
+    const indexedLine = indexed.split('\n').find(line => line.includes('(no hypothes'));
+    expect(indexedLine).toBeDefined();
+    expect(indexedLine).toMatch(/\b7\b/);
 
     testRenderer.mockInput.pressEnter();
     const detail = await frameAfter(testRenderer);
@@ -5296,9 +5303,10 @@ describe('theming', () => {
     expect(detail).not.toContain('other round');
 
     testRenderer.mockInput.pressKey('ESCAPE');
-    expect(await frameAfterEscape(testRenderer)).toContain(
-      'Round 7 · recorded agent turns · no hypothesis',
-    );
+    const reopened = await frameAfterEscape(testRenderer);
+    const reopenedLine = reopened.split('\n').find(line => line.includes('(no hypothes'));
+    expect(reopenedLine).toBeDefined();
+    expect(reopenedLine).toMatch(/\b7\b/);
   });
 
   it('keeps later hypothesis planning below the existing history', async () => {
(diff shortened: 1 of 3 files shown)


Nano-AI:
Superseded by #729: reopened from a fork after the delegated-merge role change (Triage cannot push to this branch), rebased onto main after #706, and headers now follow their column's alignment.

PR #696, opened 2026-09-12
TUI: restore the vertical round rail, keeping the label-sized agent graph

AyanBinRafaih:
## Problem

#686 replaced the vertical round rail (`round-rail.ts`, introduced in #539) with a horizontal one-row tab bar across the top of the round view and gave the rail's column to the agents and transcript panes. That layout decision is being undone: rounds should read as a vertical rail in the leftmost column of the round view, to the left of the agent graph, so drilling deeper still moves left to right (rail, agents, transcript) and a run's rounds stay a top-to-bottom list rather than a windowed row.

#686 also bundled a genuinely orthogonal improvement that must survive the revert: agent-graph columns sized to their own widest label, so the graph never truncates an agent name.

## Solution

Revert #686's layout and keep its graph sizing, composed with the rail.

Restored, byte-identical to the pre-#686 state: `round-rail.ts` and `round-rail.test.ts` (including the #624 compact-redraw fix and the #645 failed-judge-verdict glyph), the `rounds` member of `RoundFocus` and the `focusedPane` rail comment in `session-model.ts`, the rail order in `keybindings.ts` (Left/Right step rounds -> agents -> transcript while the rail is on screen, Up/Down drive the rail when it holds focus), the rail mounting and `railWidth` sizing in `app.ts`, and the rail paragraphs of the README and `todo-strip.ts`. `round-tabs.ts` and `round-tabs.test.ts` are deleted, along with the tab paragraph #686 added to `tui-conventions.md` (its note on reserving space for variable-width numbers predates #686 and is untouched). The rail's width thresholds are unchanged: 28 columns full at >= 100, 13 compact at >= 85, collapsed below.

Kept from #686, byte-identical: `agent-graph.ts` and `agent-graph.test.ts`. `graphPaneBounds` and `layoutAgentGraph` take the phases and a `labelWidth` callback; each stage column asks for its widest label plus its border, never under the readable floor; columns stay even while every label fits the even split, and only a pane under the bounds' `min` shrinks columns, widest first, never below the floor.

[...]

PR #716, opened 2026-09-12
feat(tui): animate round tabs and agent nodes with the shared spinner

Nano-AI:
## Problem

The running round and the running agent-graph node draw a static glyph (`⟳`, `●`) with nothing distinguishing "running right now" from any other moment in that status, unlike the activity bar and chat composer, which already animate exactly this state.

## Solution

Both views draw the existing shared spinner (`SPINNER_FRAMES`, `SPINNER_INTERVAL_MS` in `activity-bar.ts`) instead of a second animation. The spinner replaces the status glyph in its existing cell: same width, no layout change. Each view rewrites the glyph's `TextRenderable.content` in place on a 120ms `setInterval`, so a tick never calls `render()`/`#clear()` and never rebuilds the tree. Timers run only while a round or phase is live and stop on idle and `destroy()`. Other statuses keep their static glyphs.

Round display is the vertical rail (maintainer decision on #696). This PR originally animated the horizontal round tabs; the maintainer took it over and ported that half to `round-rail.ts`:

- The live rail row shows the spinner frame in place of `⟳`, in both the full (`▸r2 ⠹ run 1m 12s`) and compact (`▸r2⠹`) rail. The rail's existing timer, previously 1s for the elapsed label, now ticks at the spinner rate and rewrites the row's own cell; the elapsed text changes only when a whole second does.
- The frame counter lives on the view, so the animation does not restart on redraw.
- The agent-node spinner (`agent-map.ts`) is unchanged from the original PR, merged with the label-sized graph and width controls now on main.
- The tab bar changes are dropped with `round-tabs.ts`.

Limitation: the rail, graph-node, and activity-bar spinners keep independent frame counters, so they are not phase-locked.

## Verification

Live rail row, before and after (a still frame of each; the spinner glyph changes every 120ms):

```
before  ▸r2 ⟳ run 1m 12s
after   ▸r2 ⠹ run 1m 12s
```

### Correctness properties

- The rail spinner replaces only the live row's glyph; width is unchanged across frames (asserted).
- Timers stop on `destroy()` (asserted: the row content freezes).
- Non-live outcomes keep their static glyphs.

### Testing

- `pnpm check`, `biome check src`: clean
- `bun test src/ui`: 755 pass, 0 fail (new rail tests: frame advance without width change, compact glyph, stop on destroy)

### DIFF
--- a/clients/tui/src/ui/agent-map-selection.test.ts
+++ b/clients/tui/src/ui/agent-map-selection.test.ts
@@ -3,6 +3,7 @@ import {createTestRenderer} from '@opentui/core/testing';
 import type {AgentPhase} from '@vibesys/core-state';
 import type {SessionController} from '../session-controller.js';
 import {initialSessionState, type SessionState} from '../session-model.js';
+import {SPINNER_FRAMES} from './activity-bar.js';
 import {AgentMapView, nodeLabel} from './agent-map.js';
 import {resolveTheme} from './theme.js';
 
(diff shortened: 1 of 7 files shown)


vic-lsh:
Maintainer took this over: rounds stay a vertical rail (decision recorded on #696, now merged), so I merged main, dropped the tab-bar half, and ported the spinner to the rail row. The agent-node spinner is unchanged.

Issue #564, opened 2026-09-02
TUI: pressing Enter on an empty command box shows an error banner


Nano-AI:
### Observed behavior

When the command box is empty and no pane action consumes Enter, the input
submits an empty string. `parseCommand('')` returns `Enter a slash command. Use
/help.`, which appears as an error banner.

### Expected behavior

Enter on an empty command box does nothing unless the focused pane assigns an
action to it.

### Reproduction

Open a round with no selected expandable tool result and press Enter while the
command box is empty.

### Acceptance criteria

- [ ] Empty input is not submitted as a command.
- [ ] Enter still runs a typed command.
- [ ] Existing pane-specific Enter actions still work.

### Affected subsystem

`clients/tui/src/ui/command-input.ts` and Enter routing in `ui/keybindings.ts`.

### Related issues or pull requests

Roadmap #284. No existing issue or PR found.

PR #717, opened 2026-09-12
fix(tui): move input-scope errors onto the command input's hint row

Nano-AI:
## Problem

Any `scope: 'input'` report (empty box, plain text, an unknown or malformed command) raised the same full-width error banner used for real backend failures, mixing operator typos with orchestrator diagnostics on one surface. #564/#635 already made this case for the empty-box instance: "Enter on an empty box is not a malformed command, it is a no-op. Reporting it as an error trains the operator to ignore the error banner, which is the surface that has to stay trustworthy for real failures." `ErrorBannerState` has eight `scope` values; only `'input'` is client-side validation of what was just typed, the other seven carry backend `detail`/`hint` with no length contract.

## Solution

`reportError` now routes `scope: 'input'` to a new `state.inputError: string | null` instead of `errorBanner`; the other seven scopes are unchanged. `command-input.ts` gains a reserved, always-present hint row above the box (mirroring `chat-composer.ts`'s `COMPOSER_CHROME`): at rest it shows key hints (`Enter: run · Tab: complete`), on an input error it shows `✗ <message>` in `theme.error`, and the box border also turns `theme.error` (the glyph is the non-colour channel, WCAG 1.4.1). The row is reserved rather than inserted on demand, since an appearing/disappearing row resizes every pane under it (`tui-conventions.md`); this costs every view one row. The error clears on the next keystroke and on Esc. The suggestions popup now anchors to the box instead of the panel, covering the hint row while open.

### Architecture

`session-model.ts` owns the input/backend routing split and `inputError` state; `command-input.ts` renders the hint row and border from it; `session-controller.ts` and `keybindings.ts` clear it on input and Esc.

## Verification

| Before (`main`) | After |
| --- | --- |
| before | after |

Plain text + Enter in the command box, 120x32, dark theme.

`command-input.test.ts` and `session-model.test.ts` cover the render and routing logic directly; `app.test.ts` exercises the keystroke-to-hint-to-clear path, including Esc and border colour.

### Correctness properties

- `scope: 'input'` never reaches `errorBanner`; the other seven scopes keep the unchanged full-width banner with `detail`/`hint`.
- The hint row's height is constant whether or not an error stands, so no view resizes when one appears or clears.
- The error clears on the next keystroke and on Esc, never lingering on stale input.
- The error colour is paired with a `✗` glyph, not colour alone.

### Testing

- `pnpm --filter @vibesys/tui test`: 833 pass, 0 fail (31 files)
- `pnpm check:clients`: clean

### DIFF
--- a/clients/tui/src/session-controller.test.ts
+++ b/clients/tui/src/session-controller.test.ts
@@ -34,9 +34,14 @@ describe('session controller', () => {
 
     expect(transport.requests).toEqual([]);
     expect(controller.state.chatConversation).toEqual([]);
-    expect(controller.state.errorBanner?.message).toContain('Commands start with /');
+    // Ordinary text is a routing mistake, not a malformed command: it lands on
+    // the command input's own hint row rather than raising the shared banner
+    // (#564/#635's reasoning, applied to a wrong command instead of an empty
+    // one).
+    expect(controller.state.inputError).toContain('Commands start with /');
+    expect(controller.state.errorBanner).toBeNull();
 
-    controller.dismissErrorBanner();
+    controller.clearInputError();
     await controller.submitChat('what is happening?');
 
     expect(transport.requests).toEqual([{type: 'query.chat', text: 'what is happening?'}]);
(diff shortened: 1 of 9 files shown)


Nano-AI:
Superseded by #727: reopened from a fork after the delegated-merge role change (Triage cannot push to this branch), rebased onto main, and the not-a-command hint reworded.

PR #718, opened 2026-09-12
feat(tui): give a framework-validation gate command code treatment

Nano-AI:
## Problem

A framework-validation gate line (`[framework-validation] running build-and-correctness-gate: mkdir -p .cache/tmp && ...`) renders as wrapped prose in the transcript: the generic entry branch in `conversation.ts` draws every non-markdown entry as one `TextRenderable` with `wrapMode: 'word'`, breaking a shell command at its argument-separating spaces. Stacked on #694, not main, because #693/#694 already edit `conversation.ts` in the same region.

## Solution

`TranscriptEntry` and `ConversationEntry` gain an optional `command?: string`. When set, `conversation.ts` keeps drawing the prose as before and adds a standalone `CodeRenderable` with `wrapMode: 'char'` on the theme's code surface via `drawOnCodeSurface` (now exported from `styles.ts`), so the command breaks anywhere instead of at a space.

The split is a temporary adapter: core-state's `eventToTranscriptEntry` (diagnostic branch) calls `splitFrameworkValidationCommand`, matching only the exact `[framework-validation] running <recipe>: ` prefix. #697 adds a typed `command: str | None` to `GateStartedData`, but its core-state changes still flatten it back into prose (`content: \`running${recipe}${command}\``), so the field dies at the fold today; once #697 lands, delete `splitFrameworkValidationCommand` in `core-state.ts` and read `data.command` directly.

### Architecture

Owning modules: `core-state.ts` (entry shape, adapter) and `conversation.ts`/`styles.ts` in `clients/tui` (rendering). Standard core-state/tui boundary: core-state produces `TranscriptEntry.command`, the TUI only renders it.

## Verification

| Before (#694) | After |
| --- | --- |
| before | after |

Round 1 transcript of `bad-cpp-round1.jsonl`, 120x36, dark theme, cropped around the gate line.

Unit tests cover the adapter's prefix matching (ordinary colons, "running" in unrelated prose, a different bracket tag, multi-line and empty commands) and the renderer's char-wrap output on the code surface.

### Correctness properties

- Only the exact `[framework-validation] running <recipe>: ` prefix splits into a command; other diagnostic prose (colons, "running", other bracket tags, `[framework-validation] PASS`) is unchanged.
- An entry without `command` renders exactly as before: word-wrapped text, no code block.
- The command block has no syntax highlighting (no grammar attached): flat code background/foreground, always `wrapMode: 'char'`.

### Testing

- `pnpm --filter @vibesys/tui test`: 832 pass, 0 fail (30 files)
- `pnpm --filter @vibesys/core-state test`: 105 pass, 0 fail
- `pnpm check:clients`: clean

Stacked PR: CI only runs for `main`-based PRs, so these local runs are the signal.


Nano-AI:
Superseded by #730: reopened from a fork after the delegated-merge role change (Triage cannot push to this branch), rebased onto main now that #693/#694 merged, and the command block tagged bash.

PR #719, opened 2026-09-12
feat(tui): resize the docked chat pane by columns with < > and =

Nano-AI:
## Problem

The home page's docked chat pane is sized only by automatic sizing: an
operator who wants a wider pane for a long answer, or a narrower one to give
the experiment log more room, cannot ask for it. #691 gave the round view's
Agents pane exactly this escape hatch; this PR generalizes it to the home
page. It stacks on #691 (base `adi/graph-share`) and stays separate because
it is a second, independent surface adopting the same convention.

## Solution

`<`/`>` step the chat pane's width by `PANE_WIDTH_STEP` (#691's constant,
relocated from `agent-map.ts` to shared `pane-resize.ts`) and store the
result in `chatWidthOverride`, mirroring `graphWidthOverride`. `=` clears it
back to automatic sizing. The experiment log holds no width of its own
(`flexGrow: 1`), absorbing whatever the chat pane leaves, so it needs no
resize key.

`clampChatWidthOverride` lets an override exceed `CHAT_PANE_MAX`, keeping
only the log's `LOG_COMPACT_PANEL_WIDTH` floor: automatic sizing caps there
because that width is already comfortable, but an explicit ask may go past
it, mirroring how #691 lets a graph override go below `agentPaneFloor`.

### Architecture

State in `session-model.ts`/`session-controller.ts`; sizing/clamping in
`chat-pane.ts`; keys in `keybindings.ts`, mutually exclusive with #691's
Agents-pane block; shared step constant in `pane-resize.ts`.

## Verification

| Before (#691) | After |
| --- | --- |
| before | after |

Home page after pressing `>` four times (then Backspace four times, which clears the `>>>>` the base branch types into the Command box), 140x36, dark theme.

`app.test.ts`: stepping, clamping at both bounds, `=` reset, no-op presses on
the round view and on a terminal too narrow to dock the chat.

### Correctness properties

- `=` restores automatic sizing exactly.
- Clamps at `CHAT_PANE_MIN`; the high end may exceed `CHAT_PANE_MAX` but
  never narrows the log below `LOG_COMPACT_PANEL_WIDTH`.
- Resize keys advertise only on `LOG_CHAT_KEY_HELP`, never `LOG_KEY_HELP`.
- A press that would not change the rendered width stores nothing.

### Testing

- `pnpm --filter @vibesys/tui test`: 869 pass, 0 fail (30 files)
- `pnpm check:clients`: clean

Stacked PR: CI only runs for `main`-based PRs, so these local runs are the signal.

### DIFF
--- a/clients/tui/src/session-controller.ts
+++ b/clients/tui/src/session-controller.ts
@@ -77,6 +77,7 @@ import {
   setChatMenuCustomModel,
   setChatModelMenuOptions,
   setChatThreadPending,
+  setChatWidthOverride,
   setDesignLog,
   setExperiments,
   setGraphWidthOverride,
@@ -129,6 +130,8 @@ export interface SessionController {
   toggleTodos(): void;
   /** `<`/`>`: the Agents pane's explicit column width, already clamped; `=`: null. */
   setGraphWidthOverride(width: number | null): void;
+  /** `<`/`>`: the docked chat pane's explicit column width, already clamped; `=`: null. */
+  setChatWidthOverride(width: number | null): void;
   /** Expands the latest prompt in view; the view owns what "latest" means. */
   togglePrompt(): void;
   onTogglePrompt(handler: () => void): void;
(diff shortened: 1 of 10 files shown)


vic-lsh:
Rebased onto main after #691 merged (dropped its commits) and reconciled the key-help lines with main's Tab: complete change.

PR #726, opened 2026-09-13
feat(tui): run a gradient band along a completed-to-active agent edge

Nano-AI:
## Problem

While a phase runs, the agent graph highlights the active node but gives no cue that it
is drawing input from an already-completed upstream node: the live edge between them is
static. Based on #716 (spinner-driven node and tab animation) to reuse its 120 ms
interval instead of adding a second timer.

Opened from a fork, so it cannot join a native stack; it will be retargeted to `main`
once #716 merges.

## Solution

`flowRuns()` (`agent-map.ts`) picks the edge runs whose source node is `completed` and
whose target is `active`. It selects by node status rather than the `'live'` tone because
`routeEdges` gives every lane segment of a bent edge the same tone, and those segments
have no arrowhead.

On #716's tick, `paintEdgeFlow()` recolours each flow run in place (styled text, no new
renderable, no new timer). The edge keeps its live colour, and a 2-cell band moves one
cell per tick toward the arrowhead, then wraps. Glyphs never change; the arrowhead keeps
its glyph and live colour.

```
tick 1   █ ▒ ▒ ▒ ▶
tick 2   ▓ █ ▒ ▒ ▶
tick 3   ▒ ▓ █ ▒ ▶
tick 4   ▒ ▒ ▓ █ ▶      ▒ live   ▓ brighter   █ brightest
```

`brighter` and `brightest` mix the live colour 35% and 70% toward `textStrong`, so they
gain contrast on light themes as well as dark (dark `#22d3ee` → `#6de1f3` → `#b8eef8`,
light `#0e7490` → `#0a4e66` → `#06273b`). An earlier spike used a 3-cell fading tail,
which on a 4-cell edge recoloured the whole line at once and read as a blink; keeping
the live colour as the base avoids that.

Known limits:
- Only straight single-row hops animate; a bent edge onto a stacked column does not.
- Several inbound edges into one node would tick in lockstep; not reachable while
  execution is sequential.

### Architecture

All in `clients/tui/src/ui/agent-map.ts` (plus `agent-map-edge-flow.test.ts`), driven by
the spinner interval #716 already runs.

## Verification

Round 1 with `implementer` active, dark theme, the `edge-active` run. A still frame of a
120 ms animation.

| Before (#716) | After |
| --- | --- |
| before | after |

### Correctness properties

- An active node's outbound edges never band, across ticks.
- The head advances one cell per tick and wraps before the arrowhead; the tail never wraps.
- The arrowhead and the cells outside the band keep the live colour; glyphs never change.
- Edge renderables are mutated in place; with no active node nothing animates and no
  timer runs.
- `brighter` and `brightest` have strictly more contrast against the canvas than the live
  colour in all 8 themes (tested).

### Testing

In `clients/tui` at `6ce5995` (merged with #716's current head), 0 failures:
- `bun test src/ui/agent-map-edge-flow.test.ts`: 5 pass.
- `bun test src/ui/app.test.ts`: 216 pass.
- The remaining 32 test files: 700 pass.

`pnpm check:ts`, `pnpm check:clients`, `pnpm check:ts-architecture`: clean. CI runs only
for `main`-based PRs, so these local runs are the signal until the retarget. `edfff65`
after it changes comments only.


vic-lsh:
Superseded by #831: GitHub auto-closed this when #716's branch was deleted on merge, and it could not be reopened or retargeted. #831 carries the same commits (authorship unchanged), merged with main and ported to the vertical rail layout.

PR #731, opened 2026-09-13
feat(tui): highlight bash code blocks

Nano-AI:
## Problem

**Maintainer path**: not eligible for `/merge-scoped`. Touches `clients/tui/package.json` and `MANIFEST.in`, plus new `clients/tui/assets/bash/`; none sit under a scoped prefix in `.github/delegated-merge.toml`. Adds +1.36 MB on disk for the vendored grammar.

Fenced ```bash blocks in the transcript render as flat text: the TUI only bundles the JavaScript/TypeScript, Zig, and Markdown grammars from `@opentui/core`. Once PR #730 lands, a framework gate's shell command (tagged `filetype: 'bash'`) hits the same gap. Bash highlighting was requested during review of the gate-command PR. Issue #575 covered other languages (C++, Rust, Go, Python), and vendoring grammars for it was measured and declined for bundle size; this PR is a targeted exception for bash.

## Solution

Vendors `tree-sitter-bash@0.25.1` from the npm tarball: only `tree-sitter-bash.wasm` (1,358,224 bytes) and `queries/highlights.scm`, plus its MIT `LICENSE`. ABI 15, loads under the `web-tree-sitter@0.25.10` version `@opentui/core@0.4.3` depends on.

`clients/tui/src/ui/grammars.ts` registers `bash` (aliases `sh`, `shell`) through `@opentui/core`'s `addDefaultParsers()` at module load. `styles.ts` imports it for the side effect at the top of the file, so registration runs before the shared tree-sitter client's first `initialize()`. `GRAMMAR_FILETYPES` gains `bash` and `shell` (`sh` fences are already normalized to `bash` by `@opentui/core`'s extension table before a block is built).

The #575 languages (C++, Rust, Go, Python) stay declined for bundle size. This PR is a bash-only exception, not a reopening of that decision, and needs a maintainer's sign-off on the tradeoff.

`package.json`'s `files` array gains `assets`, and `MANIFEST.in` gains `graft clients/tui/assets`. Without both, `pnpm deploy --prod` and `uv build --sdist` silently drop the grammar from the published package; both build paths were checked.

### Architecture

Grammar files live in `clients/tui/assets/bash/`, resolved from `grammars.ts` with `new URL('../../assets/bash/...', import.meta.url)`, which lands on the same path from source (`src/ui/`) and compiled output (`dist/ui/`) with no copy step.

## Verification

sha256: wasm `8292919c88a0f7d3fb31d0cd0253ca5a9531bc1ede82b0537f2c63dd8abe6a7a`, highlights `b74220d954f485b7626d2b2b61f37b522e12eb1830803e388e57dd797dc99f11`.

### Correctness properties

- `bash`, `sh`, and `shell` fences highlight with more than one token color.
- `sh` and `shell` resolve to the same grammar as `bash`.
- A fence with no registered grammar (e.g. `lua`) still falls back to flat text.
- The vendored wasm loads without an ABI mismatch under the shared client's `web-tree-sitter` version.

### Testing

Run on `e9b1f6b`:
- `pnpm --filter @vibesys/tui test`: 870 pass, 0 fail (32 files)
- `pnpm check:ts`, `pnpm check:clients`, `pnpm check:ts-architecture`, `pnpm test:ts-architecture`, `scripts/check_doc_links.py`: clean
- `grammars.test.ts` loads the vendored wasm through a real `TreeSitterClient`, not the mock.


Nano-AI:
Superseded by #734, which highlights bash and nine other languages that have no bundled grammar through highlight.js, instead of vendoring one tree-sitter grammar per language.

PR #734, opened 2026-09-13
feat(tui): highlight fenced code with no bundled grammar via highlight.js

### LINKED ISSUE #575: TUI: syntax-highlight fenced code blocks in the transcript

Nano-AI:
## Problem

#644 (which closed #575) highlights fenced code only where `@opentui/core` bundles a
tree-sitter grammar: JavaScript, TypeScript, Markdown and Zig. The code VibeSys
optimizes is mostly elsewhere (the examples hold 122 Python, 17 Rust, 16 Go and 3 C++
files), and it still renders as plain text in the transcript.

Breadth was held back for size. #644's out-of-scope section measured vendoring
tree-sitter grammars at 5.0 MB of wasm for C++/Rust/Go/Python and 13.8 MB for eleven
languages, against 5.5 MB for all ~190 highlight.js languages, and recommended
`lowlight` through `CodeRenderable`'s `onHighlight` hook. That choice was confirmed on
2026-09-10. #731 (one vendored bash grammar) is the per-language form of the declined
route and is closed in favour of this PR.

## Solution

`drawOnCodeSurface` sets `onHighlight` on a fence whose filetype has no bundled grammar
but is registered with lowlight: rust, c, cpp, go, python, bash, json, yaml, ini (toml)
and diff, plus their hljs aliases (`sh`, `py`, `hpp`, ...). JS/TS/Zig/Markdown never
reach lowlight; any other language still draws flat.

The callback walks lowlight's hast tree and gives each text node the family of its
nearest mapped `hljs-*` ancestor, so spans never overlap. Families are the existing
`markdown.*` token colours; diff's `+`/`-` lines use the theme's existing `success` and
`error` through new `diff.plus`/`diff.minus` styles. No theme colour changes. Offsets
are UTF-16 indices, which is what `treeSitterToTextChunks` slices with.

Cost: `lowlight` 3.3.0 and `highlight.js` 11.11.2, pinned. The imported code (hljs core
plus ten languages) is 156 KB; the `highlight.js` package installs at 5.5 MB because it
ships every language. `package.json` and the lockfile are outside `/merge-scoped`, so
this needs a maintainer merge.

Known misses: builtins (`printf`, `len`, `echo`) and Go function names stay plain;
C++ generic params, Rust macro scope and bash heredoc interpolation are not
distinguished; parameter-list punctuation takes the variable colour. Function names use the existing
`markdown.function` grey, which is dimmer than plain code on dark themes; that palette
issue already affects TypeScript highlighting on main and is out of scope here. Untagged fences
are not auto-detected.

### Architecture

All in `clients/tui/src/ui/styles.ts`. No renderer, theme or protocol change.

## Verification

Dark theme, `dev/fixtures/markdown.jsonl`. Rust is flat on main and highlighted here;
the TypeScript block is identical in both.

**Before** (`main`)
before

**After**
after

### Correctness properties

- Grammar filetypes never call lowlight; their tree-sitter spans are used as before.
- A language in neither set (e.g. `lua`) still draws flat without waiting on highlighting.
- Spans stay aligned after multibyte text and never overlap.

### Testing

- `bun test --max-concurrency 1 src dev` in `clients/tui`: 883 pass, 0 fail.
- New `styles.test.ts` cases through `createTestRenderer`: rust keywords, the `sh` alias,
  multibyte offsets, diff `+`/`-` colours (the diff case fails before its mapping).
- `pnpm check:ts`, `pnpm check:clients`, `pnpm check:ts-architecture`: clean.


vic-lsh:
Merged main in and updated the gate command entry test to expect the highlighted (lowlight bash) output instead of flat.

PR #737, opened 2026-09-13
Server: compare store identity on a resumed subscription, not just mid-stream

### LINKED ISSUE #736: Reconnect resume ignores store_id, so an event-store swap during a dropped connection silently corrupts the client's folded state
## Observed behavior

When a subscriber's connection drops and the client redials the subscription with `resumed` set, the server resumes streaming from the client's last cursor **without comparing the event store's identity**. PR #666 added `store_id` to `EventBatchMessage` and made the mid-stream `subscription_checkpoint` path re-bootstrap on an identity mismatch, but the resume dial has no such check.

If an `EventJournal.attach` swap replaced the scratch bootstrap store with a nonempty durable log while the connection was down, the durable log's sequences are renumbered onto its own tail and the store keeps a different `store_id`. The cursor the client carries now names different events in the new store. On resume:

- durable sequences at or below the stale cursor are dropped by `foldEvent`'s out-of-order guard,
- durable sequences above it are folded on top of unrelated state as if they continued it,
- `history_after_sequence` stays at its stale value, so the backfill path is told the history is complete and never asks for the prefix it is missing.

The failure is silent: no error, and no re-bootstrap. This is the same defect class as #470, reached through the reconnect door rather than the live mid-stream attach.

## Expected behavior

A resumed subscription should be validated against the current store identity exactly as the mid-stream path is. If the client's last-seen `store_id` differs from the journal's current `store_id` (read under the same condition lock that guards the event read), the server re-bootstraps the subscription and the client routes the batch through `applyEventRebootstrap`, resetting (not lowering) its history floor, so a resumed client reaches the same folded state a full replay would produce.

When the store is unchanged, resume from the cursor exactly as today. When the client reports no prior identity (a server that never sends `store_id`, or a first connection), fall back to today's watermark behavior. The change must be backward compatible in both directions, matching the `store_id` defaults #666 established.

## Reproduction

1. Start a run. A client subscribes against the scratch bootstrap store and folds the low sequences (`server_started`, `run_status_changed`, `run_started`, ...).
2. Drop the client connection.
3. While disconnected, let `EventJournal.attach` swap in a **nonempty** durable log whose length differs from the preserved bootstrap sequences. This renumbers the pending bootstrap events onto the durable tail and the store keeps its own `store_id`, distinct from the bootstrap store's.
4. Reconnect. The client redials with `resumed` set and its old cursor.
5. Observe: the resumed batch continues from the stale cursor against the new store. Low durable sequences are never delivered, events above the cursor fold onto unrelated state, and the declared history floor never drops to request the missing prefix. No re-bootstrap occurs.

This mirrors #470's measured shape (post-attach batch begins mid-log, low durable sequences never delivered, declared floor stuck), but the trigger is reconnect, not the live attach. #666 explicitly left this out of scope: "a reconnect that redials with `resumed` set still resumes from the client cursor ... wiring the resume dial is separate work."

## Affected subsystem

## Environment

- Commit: `main` at `ad8a770f`, which includes #666 (`6e2bfdaa`)
- OS: any (protocol-level defect, hardware/OS independent)
- Backend: VibeSys server (`src/server/`) and TUI client (`clients/tui`, `clients/core-state`, `clients/backend-client`)
- Profiler: n/a

## Impact

Major functionality is impaired

## Relevant logs

```shell
None. The failure is silent: the resumed subscription emits no error and never re-bootstraps, so the client renders a folded state that diverges from the durable log with no diagnostic.
```

#470, #666, #468, #588

## Pre-submission checks

AyanBinRafaih:
## Problem

Closes #736.

PR #666 made a mid-stream event-store swap detectable by identity: `EventStore` mints an opaque `store_id`, `EventBatchMessage` carries it, and the live checkpoint path re-bootstraps when the id changes. It fixed the swap while a client stays connected, but left the reconnect door open, and said so: "a reconnect that redials with `resumed` set still resumes from the client cursor ... wiring the resume dial is separate work."

That gap is a silent correctness defect. When a client's connection drops and it redials with `resumed` set, the server resumes from the client's cursor with no identity check. If an `EventJournal.attach` swapped a nonempty durable log in while the client was gone, that log's sequences are renumbered onto its own tail and the store keeps a different `store_id`, so the cursor now numbers a store that is gone. On resume, durable sequences at or below the stale cursor are dropped by `foldEvent`'s out-of-order guard, sequences above it fold onto unrelated state as if they continued it, and `history_after_sequence` stays put so backfill is told the history is complete and never asks for the missing prefix. No error, no re-bootstrap: the same defect class as #470, reached through the reconnect door instead of the live attach.

## Solution

Wire the identity #666 introduced onto the resume dial, on both ends, mirroring the mid-stream path.

`SubscribeRequest` gains a `store_id`, the store the caller's `after_sequence` numbers. It defaults empty and the client omits it on a fresh dial, so a boot subscribe and an old server stay byte-for-byte what they were; an old server that forbids the field rejects it, which the client treats as a capability signal exactly as it already does for `tail`.

On the server, `subscription_bootstrap` takes the caller's `store_id` and reads the journal's current one under the same `_condition` lock that guards the event read. When the caller names a store that is no longer live, its cursor is dropped and the live store is replayed from its floor (a resume carries no tail, so that is a full replay), returning the new store's identity on the batch. A matching or empty id resumes the cursor exactly as before, so a reconnect with no swap still streams incrementally.

On the client, `SessionController` sends the store it last folded on every resume, falling back to a plain cursor resume if the dial is rejected. When a resumed batch names a different known store, the controller routes it through `applyEventRebootstrap` and resets (not lowers) the history floor, so the resumed client reaches the state a full replay would have built rather than extending a fold with another log's sequences. A missing identity from an older server preserves the existing fold; an identity first seen on a resumed suffix is learned without resetting history, so the next resume can name it.

### Architecture

```mermaid
flowchart TD
    A["SessionController.storeId(): the store last folded"] --> B["PersistentEventStream #resumeDial: names it on the resume, falls back to a plain resume if rejected"]
    B --> C["SubscribeRequest.store_id on the wire"]
    C --> D["RunApi.subscription_bootstrap(after_sequence, tail, store_id): compares under _condition, drops the cursor on a mismatch and replays the live store"]
    D --> E["EventBatchMessage.store_id names the store the batch was read from"]
    E --> F["SessionController #onMessage: resumed batch names a different store -> applyEventRebootstrap + reset floor; same store -> applyEventBatch at the kept floor"]
```

## Verification

| Before | After |
| --- | --- |
| Before | After |

### Correctness properties

- A resume whose `store_id` names a store the journal has since replaced re-bootstraps: the cursor is dropped and the live store is replayed from sequence 1 (no tail on a resume), so the client re-folds to exactly what the durable store holds, starting at the `run_started` the stale cursor covered.
- A resume whose `store_id` matches the live store keeps its cursor and streams incrementally, so a reconnect with no swap is unchanged.
- The identity read and the event read are atomic with respect to `attach`: the comparison happens inside `RunApi._condition`, so a swap cannot race the bootstrap.
- Backward compatible in both directions. `store_id` defaults empty and is omitted on a fresh dial, so old servers and first boots are unchanged; a client that sees no store resumes its cursor as before; and a server that forbids the field makes the client fall back to a plain resume rather than fail the reconnect.
- After a resume re-bootstrap the declared floor is reset, not lowered, so the client reports the new store's real floor and a full-replay resume declares a complete history with nothing left to backfill.

### Testing

- `uv run pytest tests/server`: 355 passed. New: `tests/server/test_transport_subscription.py::test_resume_across_a_store_swap_rebootstraps` (drops the stale cursor, replays sequences 1..latest with `run_started` first, new `store_id`, floor 0) and `test_resume_within_the_same_store_extends_from_the_cursor` (same store keeps the cursor, empty extend batch). Updated the `bursty_bootstrap` monkeypatch to accept the new keyword.
- `bun test` clients: backend-client 27, core-state 130, tui 907, all passing. New: `persistent-event-stream.test.ts` cases for naming the store on a resume and falling back to a plain resume when the field is rejected; `session-controller.test.ts` case that re-bootstraps when the store was swapped while the stream was severed and asserts the resume carried the last-seen store.
- Regenerated the protocol (`pnpm --dir clients/backend-client generate:protocol`) so `protocol.schema.json` and `protocol.generated.ts` agree with `src/server/api/protocol.py`.
- `uv run ruff check src/server tests/server`, `uv run ruff format --check`, `./scripts/check_types.sh`, and `pnpm check:ts` (biome) all clean.

- CI follow-up: corrected three subscription expectations to include `storeId: undefined`. `pnpm check:clients`, `pnpm check:ts`, architecture tests, backend-client (27), core-state (130), and TUI (907) tests passed locally.
- Added two compatibility regressions for a resumed suffix with missing or first-seen store identity. `bun test src/session-controller.test.ts` passes all 100 tests; the focused reconnect tests, TUI TypeScript check, Biome, and `git diff --check` also pass. Both new tests fail against the preceding commit `1383a61a`, where the preserved `maxRounds` changes from 3 to null, and pass after the fix.


AyanBinRafaih:
Captured a controlled reconnect with [`microsoft/tui-test`](https://github.com/microsoft/tui-test) at 150×40, dark theme, through the real TUI and Unix-socket client.

| Before ([`ad8a770f`](https://github.com/uw-syfi/vibesys/commit/ad8a770fa647cfb092875c8098bdd6d7edb472de)) | After ([`d8940643`](https://github.com/uw-syfi/vibesys/commit/d89406431d09cd0e00ed69f631f26dc849adedbe)) |
| --- | --- |
| Before: stale scratch-store prefix mixed with the durable suffix | After: full durable-store replay replaces the stale fold |

The fixture bootstraps store A through sequence 4, drops the event socket, swaps to store B, and accepts the automatic resume. Before the fix, the stale cursor is resumed against store B: the transcript retains `STALE SCRATCH STORE`, misses store B's prefix, and keeps the old two-round configuration while appending `DURABLE WORK`. After the fix, the resume names store A, the server replays store B from sequence 1, and the TUI replaces the fold with `DURABLE STORE PREFIX`, `DURABLE SETUP`, `DURABLE WORK`, and the four-round configuration.

These captures use a controlled local JSONL protocol fixture rather than a real optimization backend. They still exercise the production TUI, `ServerClient`, reconnect loop, session fold, OpenTUI rendering, Unix socket, and round navigation. The controlled labels make the otherwise silent corruption visible and keep both captures deterministic.

[Capture metadata and fixture source](https://github.com/AyanBinRafaih/vibe-serve/tree/f711bb2249e0e9581536b773317b2aa0de5cab34/pr-737).

PR #766, opened 2026-09-15
Add managed Kubernetes evaluation for microservices

Ryclic:
## Problem

VibeSys needs a managed Kubernetes lifecycle for microservice correctness and performance evaluation. The existing evaluator commands cannot own a namespace, build and load candidate images, expose endpoints, or stop and restart application pods for recovery checks.

## Solution

Add an application-agnostic `kubernetes-runtime` entrypoint to the packaged microservice evaluator, plus Hotel Reservation, Social Network, and Train Ticket configurations. The adapter validates manifests, manages candidate images and readiness, forwards endpoints, runs the evaluator, and cleans up its owned namespace.

Hotel now has explicit `compose` and `kubernetes` tasks sharing the existing Go oracle. A small example-owned launcher resolves its Go module dependency against the installed evaluator package. Kubernetes recovery restarts application pods and reservation memcached while preserving MongoDB; the oracle honors its existing opt-in durable-availability strictness flag while always checking durable capacity.

### Architecture

```mermaid
flowchart LR
  V[VibeSys task] --> R[Packaged Kubernetes runtime]
  R --> N[Owned namespace and candidate deployments]
  R --> E[Go correctness evaluator or servicebench]
  N --> F[HTTP port forwards]
  E --> F
  E --> C[Local stop/start control socket]
  C --> R
  R --> X[Ownership-checked cleanup]
```

The runtime owns deployment and process lifecycle; application manifests and correctness semantics remain in their respective examples. Control requests are restricted to stop, start, and cleanup.

Hotel invocation changes to `--input examples/microservices/hotel-correctness --task compose` or `--task kubernetes`.

Application manifests, runtime configs, and workloads live in each example's task directory (`examples/microservices/<example>/.vibesys/tasks/kubernetes/`), not in the shipped evaluator package. The runtime resolves manifests relative to the `--config` file. Social Network and Train Ticket now use the same `--input <example> --task kubernetes` layout as Hotel. Each task directory has a README recording manifest provenance and post-generation edits; Train Ticket's `generate_manifest.py` is the source of truth and a unit test checks the committed manifest against it.

## Verification

[...]

PR #770, opened 2026-09-19
Move kvstore-default to the task-directory layout

vic-lsh:
## Problem

`examples/data-structures/kvstore-default` uses the legacy root-level bundle layout, which is a compatibility path (docs/cli-flags.md). PR #766 showed that leaving trusted files outside `.vibesys/tasks` makes them agent-writable.

## Solution

`git mv` OBJECTIVE.md, vibesys.input.toml, reference/, accuracy_checker/ (including the Porcupine Go checker) and benchmark/ into `.vibesys/tasks/default/`. Update the two command paths in the toml and the README commands. The scripts locate siblings via `Path(__file__).parent`, so they need no change. Single task named `default`, so `--input examples/data-structures/kvstore-default` still resolves.

## Architecture

No source changes. Only example files move. The empty `__init__.py` stays at the example root: nothing in tests/, pyproject or docs imports the example as a package. The example shares no files with examples/kv-store.

## Verification

- `policy_check.py`: read_only_paths ['.vibesys']; trusted ['.vibesys/tasks']; READ-ONLY `.vibesys/tasks/default/accuracy_checker/checker.py`, READ-ONLY `.vibesys/tasks/default/benchmark/benchmark.py`; POLICY CHECK OK.
- `entrypoints.headless validate` resolves with and without `--task default`.
- format, check_format, check_lint pass.
- Not verified: the checker run needs Go, which is not installed here (reference run reached the Porcupine step, then failed on missing Go). No docker/GPU/live run.

## Correctness properties

- Every path read by the accuracy and benchmark commands is read-only (policy_check output above).
- The example resolves via validate, with `--task` omitted and given.
- Behavior unchanged: same commands, same scripts, history-preserving moves.

## Testing

No test covers this example (CI `validate-examples` globs only `examples/*/repositories/*/...`). Checked by policy_check and validate as above.


vic-lsh:

PR #773, opened 2026-09-19
Move large model-serving examples to the task-directory layout

vic-lsh:
## Problem

The legacy input layout (OBJECTIVE.md, checkers and benchmark at the input root) is a compatibility path (docs/cli-flags.md). PR #766 showed that leaving evaluator-trusted files outside `.vibesys/tasks` makes them agent-writable. Six large model-serving examples still use the legacy layout.

## Solution

Move OBJECTIVE.md, vibesys.input.toml, objectives.toml, accuracy_checker/ and benchmark/ into `<example>/.vibesys/tasks/default/` with `git mv` for these examples: deepseek-v3.2-8xb200, glm-5.2-8xb200, kimi-k3-8xb200, llama-70b-2xh100, qwen3.5-397b-a17b-4xb200, qwen3.6-35b-a3b-2xh100. Checker and benchmark command paths and the benchmark command in each OBJECTIVE.md now point at `.vibesys/tasks/default/...`. Each README notes the layout. Single task named `default`, so `--input <example>` still works without `--task`.

The checkers and benchmarks are independent per-model copies (no shared paths or cross references; timeouts and scripts differ per model), so each moves on its own. Scripts only use `--url`/`--output-json` and have no cwd or `__file__` path assumptions. requirements.txt and README.md were not trusted in the legacy layout and stay at the example root.

### Architecture

No code change. `project_policy` now trusts only `.vibesys/tasks` for these inputs.

## Verification

No docker, GPU, Modal or kubernetes runs; behavior is unchanged by construction (pure file moves plus path strings).

### Correctness properties

For every example, `policy_check` prints:

```
read_only_paths: ['.vibesys']
```

`headless validate <example>` resolves with `--task` omitted and reports the new command paths. requirements.txt is not read by the checker or benchmark commands.

### Testing

- `uv run python /tmp/tl-tools/policy_check.py examples/model-serving/<example>` for all six: OK
- `uv run python -m entrypoints.headless validate examples/model-serving/<example>` for all six: OK
- `./scripts/format.sh`, `./scripts/check_format.sh`, `./scripts/check_lint.sh`: pass
- CI `validate-examples` globs only `examples/*/repositories/*/...`, so it does not cover these examples.


vic-lsh:

PR #774, opened 2026-09-19
Move olmo-hybrid-prefix-caching examples to the task-directory layout

vic-lsh:
## Problem
`olmo-hybrid-prefix-caching` and `olmo-hybrid-prefix-caching-vllm` use the legacy root-level input layout, which is a compatibility path (docs/cli-flags.md). PR #766 showed that leaving evaluator-owned files outside `.vibesys/tasks` makes them agent-writable once a task directory exists.

## Solution
Move OBJECTIVE.md, vibesys.input.toml, reference/, accuracy_checker/, benchmark/ into `.vibesys/tasks/default/` for both examples (`git mv`). Update the accuracy/benchmark command paths, OBJECTIVE.md paths, and READMEs. `pyproject.toml` and `requirements.txt` stay at the root (candidate environment, not legacy-trusted). Single task `default`, so `--input <example>` works unchanged. The -vllm variant is an independent copy (its vLLM checkout is a `[[workspace.sources]]` entry, not the git submodule), so it is migrated the same way.

## Architecture
No architecture change. Uses the existing task-directory layout and project policy.

## Verification
policy_check (both examples identical):
```
task: default | initialized layout: True
read_only_paths: ['.vibesys']
```
`entrypoints.headless validate` resolves both examples with and without `--task default`. format, lint, doc-link, and distribution-packaging tests pass. Not verified: docker/GPU runs, live optimization.

## Correctness properties
- Every path read by the accuracy and benchmark commands is read-only (above).
- Scripts use only cwd-relative defaults (`../model`) and args, so behavior is unchanged with cwd = project root.
- Both examples resolve via validate.

## Testing
No test covers these examples (CI `validate-examples` globs only `examples/*/repositories/*`); validated manually as above.


vic-lsh:

PR #775, opened 2026-09-19
Move train-ticket example into a default task directory

vic-lsh:
## Problem
The legacy root layout is a compatibility path (docs/cli-flags.md). PR #766 showed that leaving trusted files outside `.vibesys/tasks` makes them agent-writable. `examples/microservices/train-ticket` still used root OBJECTIVE.md, vibesys.input.toml and benchmark/, and used `evaluator.source` (a source-tree copy).

## Solution
Move the input into `examples/microservices/train-ticket/.vibesys/tasks/default` with `git mv`. Switch to the packaged evaluator (`name`/`version`, `servicebench` entrypoint) because `evaluator.source` cannot leave the task dir; `--input examples/microservices/train-ticket` still works since `default` is the only task. Update the workload path in both commands, the example and evaluator READMEs, and tests (test_microservice_inputs.py, test_kubernetes_inputs.py). `scripts/start-local-cluster.sh` (`check`/`bench`) and the `validate-microservice` workflow step now point at `.vibesys/tasks/default/benchmark/workload.toml`. The packaged evaluator (`vibesys-evaluator-microservice` 0.1.0) is the same one train-ticket-kubernetes already uses.

## Architecture
No boundary changes. The sibling train-ticket-kubernetes test compares its workload with this task's workload.toml at the new path. Kubernetes siblings had no untrusted leftovers.

## Verification
- `uv run pytest tests/vibesys/test_microservice_inputs.py tests/vibesys/test_kubernetes_inputs.py` passes (DeathStarBench submodule initialized).
- format/lint scripts pass.
- Not verified: docker/GPU/kubernetes runs.
- With the deathstarbench overlay fetched, `VIBESYS_REQUIRE_EXAMPLE_OVERLAYS=1 pytest tests/vibesys/test_microservice_inputs.py` passes (11). `servicebench --validate-only` was not run locally (no Go); CI's `validate-examples` covers it.

## Correctness properties
```
task: default
read_only_paths: ['.vibesys']
  READ-ONLY                        .vibesys/tasks/default/benchmark/workload.toml
```
`headless validate` resolves with and without `--task default`.

## Testing
Updated the two test modules above to load the `default` task via Project and assert the packaged evaluator command.


vic-lsh:

vic-lsh:
Fixed the stale workload paths in test.yml (validate-examples) and start-local-cluster.sh, and merged main after #771 landed (kept both workflow changes). The packaged evaluator matches the one train-ticket-kubernetes uses.

PR #776, opened 2026-09-19
Move examples/kv-store to the task-directory layout

vic-lsh:
## Problem

`examples/kv-store` still uses the legacy bundle layout (manifest, checker, benchmark, reference at the input root), which is a compatibility path per `docs/cli-flags.md`. PR #766 showed that leaving trusted files outside `.vibesys/tasks` makes them agent-writable and not integrity-diffed.

## Solution

Move OBJECTIVE.md, vibesys.input.toml, run_test.sh, reference/, accuracy_checker/, benchmark/ into `examples/kv-store/.vibesys/tasks/default/` with `git mv`. Command paths now point at `.vibesys/tasks/default/...`. The single task is named `default`, so `--input examples/kv-store` still works without `--task`. `requirements.txt` stays at the example root (not read by the checker or benchmark; only used by the README setup step).

The benchmark used to download YCSB into `benchmark/ycsb`, which is now read-only. It downloads to `~/.cache/vibesys/kv-store/ycsb` instead (override: `KV_STORE_YCSB_HOME`); if the home directory is not writable (some sandboxes) it falls back to `<tmpdir>/vibesys-kv-store/ycsb`. This keeps the download-on-first-run behavior with no new dependency. The obsolete `.gitignore` entries are removed. README, OBJECTIVE seed path and `test_kv_store_modality.py` are updated.

## Architecture

No Python source changes. The `_modality/kv_store/{implementer,judge,profiler}.j2` templates referenced `reference/seed_server.py` and `{{ bench_path }}`/`{{ accuracy_checker_path }}` (variables the loop never passes, so they would not render in a real run). They now use the workspace reference path (`{{ reference_path }}`, i.e. `.vibesys/tasks/default/reference/seed_server.py`), the profiler's `{{ benchmark_command }}`, and the accuracy/benchmark commands from the input manifest. The YCSB path in the judge text names the new cache location.

## Verification

```
task: default | initialized layout: True
read_only_paths: ['.vibesys']
```

- `entrypoints.headless validate examples/kv-store` resolves with and without `--task default`.
- `pytest` on test_kv_store_modality, test_profiler_template_contract, test_distribution_packaging: 16 passed.
- format, check_format, check_lint pass.
- Not verified: an actual run (docker/YCSB/Redis/Java not exercised).

## Correctness properties

- Every path read by the accuracy and benchmark commands is read-only (policy_check output above).
- The example resolves via `validate` for `--input examples/kv-store` and with `--task default`.
- The YCSB download no longer writes into the trusted task directory.

## Testing

`tests/vibesys/loops/agent/test_kv_store_modality.py` now loads the bundle via `Project.open` + `load_project_task`. CI `validate-examples` globs only `repositories/*` so it does not cover this example; the conftest `example_input_bundles` fixture does.


vic-lsh:

vic-lsh:
Fixed the kv_store prompt templates (stale reference/bench paths, undefined path variables) and added a temp-dir fallback for the YCSB download when the home directory is unwritable. Merged main; tests and checks pass locally.

PR #825, opened 2026-09-19
Use a fine-grained PAT for /merge-scoped landing writes

vic-lsh:
## Problem

The org cannot install the GitHub App used to mint the landing token in `delegated-merge.yml`. `GITHUB_TOKEN`-created queue entries do not start `merge_group` CI, so landing writes need a different credential.

## Solution

Remove the `Mint landing token` step. `LANDING_GH_TOKEN` now comes from `secrets.MERGE_QUEUE_PAT`, a fine-grained PAT scoped to vibesys (Pull requests write, Contents write). Reads, the role check, and audit comments keep `GITHUB_TOKEN`. Script logic is unchanged: it still fails closed when `LANDING_GH_TOKEN` is empty and never falls back to `GITHUB_TOKEN`. Docs, a docstring, and the workflow test are updated.

### Correctness properties

- Empty `LANDING_GH_TOKEN` still refuses before any landing write, with sanitized errors.
- No App action or App secrets remain in the workflow.

### Testing

- `uv run pytest tests/architecture/test_delegated_merge.py`: 87 passed
- `ruff check` and `ruff format --check`: clean
- Not verified: a live `/merge-scoped` run with the PAT (workflow file is hard-denied for the bot; needs the `MERGE_QUEUE_PAT` secret set).


vic-lsh:
Superseded: the org will install the GitHub App instead of using a PAT. The App-token workflow already on main stays.

PR #850, opened 2026-09-19
refactor(server,clients): migrate the backend/frontend control protocol to protobuf

vic-lsh:
## Problem

The backend/frontend control protocol is defined as Pydantic models, exported to JSON Schema, and turned into TypeScript with `json2ts`. That works for one Python/TS pair, but more SDKs and clients in other languages are expected to talk to the server. Protocol Buffers is a schema format every one of those languages has first-class codegen for, with `buf breaking` for compatibility checks. This PR migrates the existing protocol so the approach can be judged on real code. The evaluator protocol is out of scope.

## Solution

`proto/server/wire/v2/*.proto` is now the only definition of the contract. The server and all TypeScript clients use the generated messages. The JSONL transport carries proto3 canonical JSON with snake_case field names.

- Requests, `ServerMessage`, `RunEvent` payloads and `Response` use typed `oneof`s in place of the string-discriminated `type`/`kind` unions.
- `protocol_version` is now 2. Journals recorded under version 1 are upgraded line by line on load (`server.wire`), and version 1 clients get a clear rejection.
- Deleted: the Pydantic wire models, `server.api.schema`, the JSON Schema export, `json2ts` and `json-schema-to-typescript`, `protocol.schema.json`, `generate:protocol`, the parity test and the old CI drift check.
- Tooling: `buf` and `protoc-gen-es` from pnpm, `protoc` from the `grpcio-tools` dev dependency. Scripts (run from `clients/`, the pnpm workspace root): `pnpm proto:lint`, `proto:generate`, `proto:check` (regenerate and diff), `proto:breaking`. Generated code is committed, as the JSON Schema output was. Details in `docs/contributing/development.md` ("Wire protocol").

### Technical approach and tradeoffs

- **Proto as source vs Pydantic as source.** One definition now feeds every language, and `buf breaking` gives real compatibility checking. The cost is that Python loses Pydantic's validation and frozen models.
- **Validation moved to hand-written code.** Protobuf cannot express ranges, non-empty text, finite floats or required enums, so `src/server/wire/validate.py` (and the request builders in `backend-client`) re-implement them. This is the main loss relative to Pydantic.
- **Unknown fields.** The server parses strictly, so an unknown field is rejected. The `tail` and `store_id` subscribe options still rely on that rejection as their capability probe. Any client using a lenient parse would defeat it, which is a real fragility of this design.
- **Wire break.** Typed `oneof` changes the JSON shape, hence version 2. proto3 JSON also omits nulls and zero defaults and normalizes timestamp text, so consumers must treat absent and null alike.
- **Mutability.** Generated messages are mutable, so the frozen-model guarantees on `RunEvent` are gone. Tests cover the aliasing cases that mattered.
- **Toolchain cost.** Adds `buf`, `protoc-gen-es` and `grpcio-tools`. The committed generated code adds about 6.3k lines of bindings.
- **Size.** See the diff stat below the description: roughly +6.3k lines are generated bindings and +1.1k are `.proto`, against about 6.1k lines of removed Pydantic wire models and JSON Schema output.

### Correctness properties

- Every request, response, snapshot, server message and event payload is defined once, in `proto/`.
- Recorded version 1 journals load through the upgrade path with identical content.
- Unknown fields are rejected by the server.
- Values Pydantic used to validate (ranges, non-empty text, finite floats) are still rejected.
- No reference to the old JSON Schema pipeline remains.

### Testing

Run locally against the branch merged with current main, all passing:

- Protocol: `pnpm proto:lint`, `pnpm proto:check` (from `clients/`)
- Python: `ruff check` and `ruff format --check`, `./scripts/check_types.sh`, `uv run tach check`, `scripts/check_tach_graph.py --check`, `scripts/check_file_length.py`, `scripts/check_doc_links.py`
- Full `pytest -n auto --dist loadgroup --cov-fail-under=75`: 6370 passed, 62 skipped. The per-module coverage floor and patch coverage (95% against origin/main) pass once the generated `*_pb2.py` files are excluded from coverage.
- TypeScript (from `clients/`): `pnpm check:ts`, `check:ts-architecture`, `check:clients`, `test:clients` (backend-client 35, core-state 163, tui 1087 tests), `build:clients`, and the packed and deployed TUI smoke tests from the `tui` CI job.

Notes:

- `pnpm proto:breaking` finds no baseline yet: origin/main has no `buf.yaml`, so it passes with a notice. It starts comparing once this merges.
- One tui test (scratch removal with an open file) failed once under load during a full run and passed on three reruns.
- The tui test files were ported to the generated types (about 630 type errors), including the launcher test's hand-written subscribe request.


vic-lsh:
Note: this is just an exploration. May not merge.

